PR #49, opened 2026-02-03
 fix(tools): prefer source files for library classes in FindClassTool

Knight-ZXW:
Updated convertToSymbolMatch to resolve to source navigation element
    before building SymbolMatch, mirroring IDEA's behavior when sources
    are attached (e.g., SomeClient.java in source.jar).

    Changes:
    - Use SourceNavigationHelper.getInstance(project).getSourceElement(element) to resolve to source element, falling back to original element when no sources are available
    - File path, name, qualifiedName, line, kind, and language now come from the source element when present

    This means:
    - If dependency has source.jar, results point to .java/.kt files
    - If not, results still point to compiled classes in classes.jar

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@
 
 ## [Unreleased]
 
+## [3.3.3] - 2026-02-03
+
+### Fixed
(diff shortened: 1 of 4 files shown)


hechtcarmel:
Merged & submitted for Jetbrains review

PR #50, opened 2026-02-05
feat(tools): add ide_read_file tool with library source support

Knight-ZXW:
- Add new ide_read_file tool for reading source files and library dependencies
  - Support multiple file path formats (relative, absolute, jar paths, jar URLs)
  - Support qualified class name lookup (e.g., java.util.ArrayList)
  - Optional line range extraction with startLine/endLine parameters
  - Returns file metadata (language, line count, library file flag)

- Enhance library source navigation across tools
  - Prefer source files (.java) over compiled files (.class) for libraries
  - Add PsiUtils.getNavigationElement() for consistent navigation resolution
  - Update ide_find_definition and findClassByName to use navigation elements

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 build
 tmp
 .claude
+local.properties
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,21 @@
 
 ## [Unreleased]
 
+## [3.3.4] - 2026-02-05
+
+### Added
+- **New tool: `ide_read_file`** - Read source file contents from project or library dependencies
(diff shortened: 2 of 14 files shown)


hechtcarmel:
hmm, interesting usecase.
Maybe it should be disabled by default (opt-in) to not overload the context with another tool with a very specific usecase. what do you think? 
Anyways, i think the explanation should be for the agent that he should use it for library java files (otherwise it will perform worse than his native tools for most files)

Unfortunately i am on an almost continuous 3 weeks holiday starting tomorrow and i don't think i'll have the time to review properly.
Sorry, i'll get to it when i get back.

hechtcarmel:
Merged & sent new version for Jetbrains approval. May take a couple of days until they approve and it's published

PR #65, opened 2026-02-21
fix(call-hierarchy): restore callers for Kotlin suspend funs in concrete classes

mareurs:
## Summary

`ide_call_hierarchy` with `direction="callers"` returned empty results for `suspend fun` in concrete Kotlin classes. Confirmed deterministic (not flaky) across multiple symbols via systematic testing against a real project.

## Root Cause

Two bugs, both required to fully fix callers:

### Bug 1 — `MethodReferencesSearch` misses `suspend fun` call sites

`MethodReferencesSearch` operates on the JVM-desugared `PsiMethod`. For `suspend fun`, the Kotlin compiler appends a `Continuation<T>` parameter to the JVM signature — a parameter that is invisible in Kotlin source call sites. So `MethodReferencesSearch` finds zero call-site references for suspend functions.

The existing code had a `ReferencesSearch.search(navigationElement)` fallback, but it was **guarded by `if (allReferences.isEmpty())`**. If `MethodReferencesSearch` returned *any* non-empty result (e.g. declaration-site or annotation references later filtered out by the `containingMethod \!= method` check), the guard was satisfied and the fallback was silently skipped — leaving `calls: []`.

### Bug 2 — `resolveKotlinMethod` returns `null` for callers inside `val x = fn(...)` assignments

`resolveKotlinMethod` walks up the PSI parent chain looking for a `KtNamedFunction`, `KtPropertyAccessor`, or `KtProperty` and converts it to a `PsiMethod` via `toLightMethods()`. The bug: **local variable declarations are also `KtProperty` nodes** in the PSI tree.

When a caller uses the common pattern:
```kotlin
val result = service.publishSchedule(solutionId, request)
```
the reference element's parent chain first hits the `KtProperty` node for the local `val result` — *before* the enclosing `KtNamedFunction` (the actual caller). `toLightMethods()` returns an empty list for local variables (they have no backing JVM method), so `firstOrNull()` returned `null`, and `resolveKotlinMethod` returned `null` — without ever reaching the enclosing function.

The `containingMethod \!= null` guard in `findCallersRecursive` then silently dropped every such reference, producing empty `calls: []` even after Bug 1 was fixed.

`ide_find_references` (FindUsagesTool) is unaffected because it never calls `resolveKotlinMethod` — it uses `ReferencesSearch.search(targetElement)` unconditionally.

## Fix

**Fix 1:** Remove the `if (allReferences.isEmpty())` guard so that `ReferencesSearch.search(navElement)` **always runs** for Kotlin methods in addition to `MethodReferencesSearch`. Existing `seenKeys` deduplication prevents double-counting.

**Fix 2:** In `resolveKotlinMethod`, when `toLightMethods()` returns empty (local `val`/`var` — no backing JVM method), **continue walking up** the parent chain instead of returning `null`. This skips local property nodes and finds the enclosing `KtNamedFunction`.

```kotlin
// Before — stopped at local val/var, returned null:
if (isKotlinDeclaration) {
    return try {
        val lightMethods = toLightMethodsMethod.invoke(null, current) as? List<*>
        lightMethods?.firstOrNull() as? PsiMethod  // null for local val/var → drops the reference
    } catch (e: ReflectiveOperationException) { null }
}

// After — continues up the chain when toLightMethods() returns empty:
if (isKotlinDeclaration) {
    try {
        val lightMethods = toLightMethodsMethod.invoke(null, current) as? List<*>
        val lightMethod = lightMethods?.firstOrNull() as? PsiMethod
        if (lightMethod \!= null) return lightMethod
        // Empty (local val/var) — continue walking up to find enclosing function
    } catch (e: ReflectiveOperationException) { /* continue */ }
}
```

## Behaviour

| Method type | Before | After |
|-------------|--------|-------|
| `suspend fun` in concrete class | ❌ `calls: []` | ✅ returns callers |
| `fun` in concrete class | ✅ | ✅ unchanged |
| Interface method | ✅ | ✅ unchanged |
| Java method | ✅ | ✅ unchanged (`language.id \!= "kotlin"` guard) |

## Test evidence

Tested against a 426-file Kotlin backend project (`ScheduleApprovalService.publishSchedule`, a `suspend fun` with 5 known callers confirmed by `ide_find_references`). All 5 callers now appear — including callers in test methods that used the `val result = service.publishSchedule(...)` assignment pattern.

### DIFF
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/handlers/SearchMatchUtils.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/handlers/SearchMatchUtils.kt
@@ -12,20 +12,18 @@ import com.intellij.psi.codeStyle.NameUtil
 
 /**
  * Path prefixes that are excluded only when they appear at the project root.
- * These are common build output dirs that could legitimately appear as nested source dirs.
+ * These are common build output dirs or environment dirs that could legitimately appear
+ * as nested source dirs (e.g. `src/config/env/`, `docker/env/`).
  */
 internal val ROOT_ONLY_EXCLUDED_PREFIXES = listOf(
-    "bin/", "build/", "out/", ".gradle/"
+    "bin/", "build/", "out/", ".gradle/",
+    ".env/", "env/"  // Python venv aliases — root-only to avoid false positives at depth
 )
 
 /**
  * Path segments that are excluded at any depth in the project tree.
  * Virtual environments and package manager directories should never contain source files
  * regardless of where they appear in the project hierarchy.
- *
- * Note: `.env/` and `env/` are intentionally omitted — they are too generic and would
- * produce false positives (e.g. `src/config/env/`, `docker/env/`). Only the conventional
- * Python venv names `.venv/` and `venv/` are excluded.
  */
 internal val DEEP_EXCLUDED_SEGMENTS = listOf(
     ".venv/", "venv/",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## [Unreleased]
 
-## [3.9.1] - 2026-02-21
+## [3.10.0] - 2026-02-22
 
 ### Added
 - **`matchMode` parameter for `ide_find_symbol` and `ide_find_class`** - Control how queries match symbol names
@@ -22,7 +22,7 @@
 - **`ide_call_hierarchy` callers for Kotlin `suspend fun`** - `MethodReferencesSearch` misses `suspend fun` call sites because the Kotlin compiler appends a hidden `Continuation<T>` parameter to the JVM signature. Added unconditional `ReferencesSearch.search(navigationElement)` alongside `MethodReferencesSearch` (with deduplication) so callers are always found
 - **`ide_call_hierarchy` callers inside `val`/`var` assignments** - `resolveKotlinMethod` was stopping at local `val`/`var` PSI nodes (`KtProperty` with no backing JVM method) and returning `null`, silently dropping every such caller reference. Now continues walking up the PSI tree to find the enclosing named function
 - **`ide_call_hierarchy` "unknown" caller names for JSX arrow functions** - `findContainingCallable` was returning unnamed anonymous arrow functions (`const App = () => ...`) instead of the enclosing `JSVariable`. Now skips unnamed `JSFunction` nodes and falls back to the containing `JSVariable` for correct caller name resolution
-- **`ide_find_symbol` and `ide_find_class` wildcard queries returning empty results** - Both tools now request `limit × 5` results from IntelliJ before applying path exclusion filters. Previously, when all results within the requested limit were from excluded paths (venv, node_modules, etc.), the final result was empty even though real project symbols existed
+- **`ide_find_symbol`, `ide_find_class`, `ide_search_text`, `ide_find_file` polluted by excluded paths** - All search tools now use a `DelegatingGlobalSearchScope` subclass (`ExcludedPathScope`) that rejects venv, node_modules, build output, and worktree files at the IntelliJ search-infrastructure level. Excluded files never consume buffer slots, replacing the fragile over-fetch-then-filter approach
 - **`ide_find_symbol` and `ide_find_class` polluted by venv/node_modules in subdirectories** - Exclusion filter now matches `.venv/`, `venv/`, `node_modules/`, and `.worktrees/` at any path depth (not only at the project root). Fixes multi-module projects where the virtual environment is inside a subdirectory (e.g. `python-services/.venv/`)
 - **`ide_find_symbol` exact `matchMode` was case-insensitive** - Changed from `name.equals(pattern, ignoreCase = true)` to `name == pattern`. `"CalendarService"` with `exact` no longer matches `calendarService` properties
 - **`ide_find_references` duplicate entries for JSX components** - Opening and closing JSX tags (`<Foo>` / `</Foo>`) resolved to identical `file:line:column` positions, producing duplicate entries. Results are now deduplicated by position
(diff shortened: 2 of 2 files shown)


mareurs:
## Update: version → 3.9.1 + CHANGELOG + additional bug fixes

Addressed the reviewer feedback:
- ✅ Version changed to **3.9.1** (patch bump, not minor, since all changes are bug fixes)
- ✅ **CHANGELOG.md** updated with `[3.9.1]` entry

---

### All fixes in this PR

**Original: Kotlin `suspend fun` callers (`ide_call_hierarchy`)**
- `MethodReferencesSearch` misses `suspend fun` call sites — Kotlin compiler appends hidden `Continuation<T>` parameter to JVM signature. Added unconditional `ReferencesSearch.search(navElement)` fallback (with deduplication).
- `resolveKotlinMethod` returned `null` for callers inside `val result = fn(...)` assignments — local `val`/`var` are `KtProperty` PSI nodes with no backing JVM method. Now continues walking up the PSI tree to find the enclosing named function.

**Found during systematic testing against 3 real projects:**

| # | Tool | Bug | Fix |
|---|------|-----|-----|
| 3 | `ide_find_class` wildcard | Returned venv/node_modules classes instead of project classes | Exclude `.venv/`, `venv/`, `node_modules/`, `.worktrees/` at any path depth; over-fetch 5× to compensate |
| 4 | `ide_find_symbol` / `ide_find_class` | `.worktrees/` files polluted results in projects with git worktrees | Same exclusion filter |
| 5 | `ide_find_symbol` exact mode | Case-insensitive (`"CalendarService"` matched `calendarService`) | Changed to strict `name == pattern` |
| 6 | `ide_file_structure` | Returned "Language not supported" for JS/TS/JSX/TSX files | Added JS/TS handler via reflection |
| 7 | `ide_call_hierarchy` callers (JSX) | Caller names showed as `"unknown"` for arrow functions (`const App = () => ...`) | Skip unnamed `JSFunction`; fall back to enclosing `JSVariable` |
| 8 | `ide_find_references` (JSX) | Duplicate entries for `<Foo>` / `</Foo>` at same position; `truncated: true` after dedup | Deduplicate by `file:line:column`; fix `truncated = total > maxResults` |

All 8 bugs verified fixed against the same 3 projects after plugin reinstall.

src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/handlers/SearchMatchUtils.kt
```diff
@@ -10,12 +10,34 @@ import com.intellij.psi.codeStyle.NameUtil
...
+ * regardless of where they appear in the project hierarchy.
+ */
+internal val DEEP_EXCLUDED_SEGMENTS = listOf(
+    ".venv/", "venv/", ".env/", "env/",
+    "node_modules/",
+    ".worktrees/", ".claude/worktrees/"
+)
+
+// Keep for backwards compatibility / tests
+internal val EXCLUDED_PATH_PREFIXES = ROOT_ONLY_EXCLUDED_PREFIXES + DEEP_EXCLUDED_SEGMENTS
```
hechtcarmel: Property "EXCLUDED_PATH_PREFIXES" is never used

src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/navigation/FindUsagesTool.kt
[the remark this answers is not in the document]
hechtcarmel: also (so maybe they won't be the same) - 

The slot is claimed at line 121 (collected.incrementAndGet()), but if getDocument() returns null at line 123, no UsageLocation is added to the usages queue. The slot is "wasted". If many references have null documents (unlikely but possible with in-memory files), the tool could return fewer results than maxResults even though more valid references exist.

TESTING_BUGS.md
```diff
@@ -0,0 +1,179 @@
+# MCP Plugin Systematic Testing — Bug Tracking
```
hechtcarmel: this should not exist in git

gradle.properties
```diff
@@ -4,7 +4,7 @@ pluginGroup = com.github.hechtcarmel.jetbrainsindexmcpplugin
 pluginName = IDE Index MCP Server
 pluginRepositoryUrl = https://github.com/hechtcarmel/jetbrains-index-mcp-plugin
 # SemVer format -> https://semver.org
-pluginVersion = 3.9.0
+pluginVersion = 3.9.1
```
hechtcarmel: Again, not critical and fine if you leave as is, but now that you added some new features and now just bug fixes it actually should be 3.10.0 following semantic versioning

mareurs:
## Review comments addressed

All 7 comments from the review have been resolved. Here's what changed:

### `TESTING_BUGS.md` removed
Deleted from git entirely — test investigation notes don't belong in the repository.

### Version bumped to 3.10.0
Minor version bump per SemVer guidance — this PR adds features (new bugs fixed, new scope-based filtering), not just a patch.

### `LanguageHandler.kt` KDoc fix
Corrected the `@param matchMode` doc: `"exact"` match is **case-sensitive**, not case-insensitive.

### `SearchMatchUtils.kt` cleanup
- Removed `env/` and `.env/` from `DEEP_EXCLUDED_SEGMENTS` — these are too generic and caused false positives (e.g. `src/config/env/`, `docker/env/`). Only the conventional Python venv names `.venv/` and `venv/` are excluded.
- Removed unused `EXCLUDED_PATH_PREFIXES` constant.

### `FindUsagesTool.kt` simplification
Removed the redundant `collected` `AtomicInteger` — it was always equal to `totalFound`, so `totalFound` now serves both roles (counting total references and controlling the buffer slot).

### Scope-based path exclusion (`ExcludedPathScope.kt`)
Replaced the over-fetch pattern (`fetch N×limit`, filter, take `limit`) with a new `ExcludedPathScope` that extends `DelegatingGlobalSearchScope`. Excluded files are now rejected at the IntelliJ search-infrastructure level and never consume buffer slots. The `createFilteredScope()` factory wraps either `projectScope` or `allScope` with this filter.

Applied to all consumers:
- `FindClassTool`, `FindSymbolTool`, `FindFileTool`, `SearchTextTool`
- All 6 language handler `searchSymbols` methods (Java, Python, JavaScript, Go, PHP, Rust)

hechtcarmel:
Also, CHANGELOG.md still has the old version and not 3.10.0.
Also update CHANGELOG wording to describe scope-based filtering instead of over-fetch

hechtcarmel:
submitted to jetbrains

Issue #41, opened 2026-01-07
Extend / Move to built-in Intellij MCP Server


Haehnchen:
> A JetBrains IDE plugin that **exposes** an MCP (Model Context Protocol) server.

To avoid reinventing the wheel,t tried searching for existing projects that expose intelliJ features via mcp. You might consider use the now built-in mcp server:
https://plugins.jetbrains.com/plugin/26071-mcp-server

I think features like port and project resolution would improve the user experience + initial setup would be much smoother.

Since the MCP server is relatively new, here are some examples i was playing with:
https://github.com/Haehnchen/idea-mcp-devkit

PR #93, opened 2026-03-14
feat!: port plugin MCP tools to JetBrains built-in MCP server

Gabriel-Hiss:
replaced ktor mcp server to use intelij built in mcp server so when you add the intelij idea mcp server all the tools from this plugin are transferred to the default server without adding any more server

#This is a breaking change so be aware that any pull reqeust might need to adapt to the new format before merge


hechtcarmel:
Hi,
Thank you.

I was meaning to move it to use the official mcp sdk which is now more mature. But this is something else.

Please look at the following thread: #41  

Does it mean that if we merge this:
1. The ports will change (breaking users)
2. The different IDEs won't be able to configure their own port/host (which was a requested feature peopel use)
3. If we run multiple IDEs, it's a lottery which ports get grabbed? (breaking the mcp and causing the need to re-install)

Gabriel-Hiss:
the ide mcp server it sends all tools by the intelij mcp it's not like you have another one running in parallel it just adds some tools to the inteloj mcp server i've seen the thread i think that is what the guy means and you don't need te ui anymore

Gabriel-Hiss:
i just have a buch of ides mcp configured because when claude tries to reach them and fail it just does't list them to the it is a dumb work around but it works

PR #119, opened 2026-04-07
Fix: resolveWindows path issues in jar library security check

johnwas:
Summary

  Fixes a critical bug in resolveVirtualFileAnywhere that caused false-negative rejections of legitimate JAR file access on Windows, introduced in PR #110 when JAR reading restrictions were added.

  Problem

  When commit 053c895 added JAR library whitelist validation, it used simple string startsWith() comparison which failed on Windows due to:
  1. Backslash path separators not being normalized to forward slashes
  2. VFS normalizing paths to lowercase (Z:/Temp → Z:/temp) while NIO Path comparison is case-insensitive on Windows
  3. JarFileSystem returning paths with leading / before drive letters (/C:/path)

  This caused ide_read_file and other tools to incorrectly block access to project library JARs on Windows.

  Solution

  - Normalize path separators at function entry (backslash → forward slash)
  - Use NIO Path-based isPathPrefixOf for library validation, which correctly handles case-insensitive comparison on Windows
  - Add isPathPrefixOf helper with fallback to string comparison for invalid paths
  - Normalize getProjectLibraryJars() output to handle Windows drive letter prefix and backslashes

  Files Changed

  - PsiUtils.kt: Path normalization and isPathPrefixOf helper
  - PsiUtilsTest.kt: Test uses forward slashes for valid jar:// URLs
  - CHANGELOG.md: Add 4.10.1 entry
  - gradle.properties: Bump version to 4.10.1

  Testing

  Test testResolveVirtualFileAnywhere_ResolvesJarEntry verifies the fix works correctly with project library JARs.

### DIFF
--- a/.idea/gradle.xml
+++ b/.idea/gradle.xml
@@ -5,7 +5,7 @@
     <option name="linkedExternalProjectsSettings">
       <GradleProjectSettings>
         <option name="externalProjectPath" value="$PROJECT_DIR$" />
-        <option name="gradleJvm" value="21" />
+        <option name="gradleJvm" value="temurin-21" />
         <option name="modules">
           <set>
             <option value="$PROJECT_DIR$" />
--- a/.idea/gradle.xml
+++ b/.idea/gradle.xml
@@ -5,7 +5,7 @@
     <option name="linkedExternalProjectsSettings">
       <GradleProjectSettings>
         <option name="externalProjectPath" value="$PROJECT_DIR$" />
-        <option name="gradleJvm" value="temurin-21" />
+        <option name="gradleJvm" value="21" />
         <option name="modules">
           <set>
             <option value="$PROJECT_DIR$" />
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@
 
 ## [Unreleased]
 
+## [4.4.1] - 2026-03-19
+### Fixed
+- **2026.1 compatability issues** 
+
+## [4.4.0] - 2026-03-18
+### Added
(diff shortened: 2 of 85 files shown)


hechtcarmel:
Hi,
Thank you for this fix.


1. This introdues some other issues.

To name one - 

PsiUtils.resolveVirtualFileAnywhere() now rewrites every backslash to / up front (src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/util/PsiUtils.kt:117-154). That fixes Windows-style input, but it also changes the meaning of valid macOS/Linux file names, because on POSIX \ is a normal filename character, not a path separator.

That means a request for a legitimate file such as src/a\\b.txt or Foo\\Bar.java will now be rewritten to src/a/b.txt or Foo/Bar.java and fail to resolve. Before this change, resolveLocalFile() delegated to Path.of(...), which preserves \ as a literal path character on POSIX.

This is not isolated to ide_read_file; it also affects Java-to-Kotlin conversion because that tool uses the same helper for input resolution and for post-conversion deletion checks. 

-----

I looked into it more, what do you think about using something like
```
import com.intellij.openapi.util.SystemInfo

private fun normalizeSeparatorsIfWindows(path: String): String =
    if (SystemInfo.isWindows) path.replace('\\', '/') else path

```

to do the windows specific changes only on windows?

2. Curious why you added `pluginVerifier()` in build.gradle.kts ?

johnwas:
Good catch, You're right that the unconditional replace('\\', '/') breaks POSIX semantics — on macOS/Linux, \ is a valid filename character and should not be treated as a separator.

Fixed in the latest commit: the normalization is now guarded by SystemInfo.isWindows, since \ is forbidden in Windows filenames and is unambiguously a path separator there. The other changes from that commit
  (isPathPrefixOf, getProjectLibraryJars fixes) are unaffected.
    
When I submitted my pull request for the first time, GitHub's CI verification process threw an error, prompting me to add `pluginVerifier()` to `build.gradle.kts`, which I did. After that, GitHub's CI worked correctly. I don't know why this happened; it hadn't occurred before.

hechtcarmel:
Submitted for jetbrains approval, thanks for the contribution!

Issue #149, opened 2026-04-22
Markdown PSI support?


pancsta:
It's useful for pulling repo docs in an efficient way, thx.

hechtcarmel:
Hi,
Could you please explain in more detail what would you want out of it? how would usage look like?
I looked into it and i'm not sure i see the benefit over agents simply reading the markdowns.

refactoring markdowns - not sure i see the benefit, as agents are pretty good at refactoring single files.
"file_structure" tool support - simillarly, unless the markdown is HUGE agents are pretty good at reading documents..

pancsta:
Mostly structure and go to symbol, it's useful if all your docs are in markdown (meaning many huge files, say ~20k tokens for a starter pack). Go to symbol would overlap with looking for code symbols and lead to expanded docs. There's a couple of LSPs for markdown, eg [marksman](https://github.com/artempyanykh/marksman).

I'm also missing pulling docs for symbols, but subsidizing with `go doc URL.Symbol` (same for package structure).

hechtcarmel:
Hi, 
Submitted version 4.15.0 to jetbrains, which adds markdown psi support.
(10 minutes to 5 days till they approve and it's available).

PR #152, opened 2026-04-24
fix(find-symbol): align ide_find_symbol with IntelliJ's Go to Symbol popup

vcth4nh:
## Summary

Makes `ide_find_symbol` behave like IntelliJ's **Go to Symbol** popup — same results, same ordering, same qualified-query handling — and extends the tool to every JetBrains IDE. Released as **4.16.0** (rebased after upstream took 4.15.0 for Markdown PSI support).

## What changed

`ide_find_symbol` now issues a **single** call to IntelliJ's Go to Symbol popup stack (`OptimizedSymbolSearch.search(...)`) instead of looping over per-language `SymbolSearchHandler`s and concatenating their slices. This preserves the popup's cross-language ranking and lets the tool pick up every language IntelliJ indexes — including C/C++ in CLion, which v4.14.1/4.15.0 couldn't see at all.

- `FindSymbolTool.doExecute()`: removed the handler loop; one popup-backed call with the native `languageFilter` argument (no post-filter).
- `FindSymbolTool` moved from `ConditionalTool { hasSymbolSearchHandlers() }` to `registerUniversalTools()` — registered in every JetBrains IDE.
- Deleted the `SymbolSearchHandler` interface, the registry plumbing (`symbolSearchHandlers`, `registerSymbolSearchHandler`, `getAllSymbolSearchHandlers`, `hasSymbolSearchHandlers`, `getSupportedLanguagesForSymbolSearch`), and the 9 per-language impls (Java / Kotlin / Python / JS / TS / Go / PHP / Rust / Markdown). ~355 lines removed.
- `OptimizedSymbolSearch` language filter is now case-insensitive (the tool-level filter in 4.15.0 was already case-insensitive; this makes the internal call match).

Schema, cursor format, pagination, error shapes — all unchanged. Qualified queries (`Class.method`) and wildcards (`*Solver`, `Batch*`) continue to behave per the popup.

## Empirical A/B (v4.14.1 vs 4.16.0)

Same `main` query, same project, same IDE — verified by installing each zip in turn and replaying:

| IDE | Project languages | v4.14.1 top-10 exact `main` | 4.16.0 top-10 exact `main` | v4.14.1 total | 4.16.0 total |
|---|---|---|---|---|---|
| PyCharm | Python + TS + CSS | **0/10** (all TS CamelCase: `mapInternalToDisplayNames`, `marginInlineEnd`, `max_input_tokens`, …) | **10/10** (2 CSS + 8 Python exact `main`) | 233 | 235 |
| RustRover | Rust + TS + JS | **2/10** (JS + TS; rest were `MarketplaceInterface`, `domain`, `*Domains` fragments) | **10/10** (JS + 8 Rust `fn main()` + TS) | 401 | 404 |
| CLion | C/C++ + TS + JS | **0/10** (all TS/JS: `maskInlineLaTeX`, `MAX_INACTIVE_CONVERSATION_STATES`, `_generateMinMaxInt`, …) — **zero C++ `main()` anywhere in results** | **10/10** (all C++ `main()` from `test-*.cpp`) | **22** | **500** |

CLion is the starkest case: v4.14.1 is blind to the entire C/C++ codebase because there's no C/C++ `SymbolSearchHandler`. 4.16.0's single popup call goes through every `ChooseByNameContributor`, so C++ matches surface (~23× more total results).

Queries dominated by a single language (e.g. `new` on RustRover: 500 Rust `fn new()` results on both versions) return identical top-10 — per-handler concatenation only matters when multiple languages have matches.

## Risks

- **User-visible ordering shift.** Clients relying on v4.14.1/4.15.0's HashMap-iteration ordering will see popup ordering instead.
- **Result quality in newly-enabled IDEs.** `OptimizedSymbolSearch.convertToSymbolData` uses reflection + class-name heuristics. For languages outside the plugin's handler set (C/C++, Ruby, C#, F#, SQL), `language` may be empty, `kind` falls back to `"SYMBOL"`, `qualifiedName` is null, and line/column may default to the file start (`1, 1`) rather than the symbol's actual position (observed on C++ in CLion — the `file` field is correct, so navigation opens the right file, but in-editor positioning may need an extra click). Per-language refinement is a follow-up.

## Closes

Closes vcth4nh/jetbrains-index-mcp-plugin#1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@
 
 ## [Unreleased]
 
+## [4.16.0] - 2026-04-24
+### Fixed
+- **`ide_find_symbol` ordering, missing/extra results, and qualified-query handling now match IntelliJ's Go to Symbol popup.** The tool previously ran the popup search separately for each registered language handler and concatenated results in handler-iteration order, which destroyed cross-language ranking. Symbol search now issues a single popup-backed call.
+
+### Added
+- **`ide_find_symbol` is now available in every JetBrains IDE**, including RubyMine, CLion, DataGrip, Rider, and Aqua. Result quality depends on IDE-supplied `ChooseByNameContributor` extensions; `kind` and `qualifiedName` may fall back to generic values for languages the plugin doesn't special-case.
+
+### Changed
+- Internal: removed the `SymbolSearchHandler` interface and its nine language implementations (including the Markdown symbol-search handler added in 4.15.0); symbol search is now centralised in `OptimizedSymbolSearch` + `PopupFaithfulSymbolSearch`. Markdown headings still surface through the single popup-backed path.
+
 ## [4.15.0] - 2026-04-24
 ### Added
 - Added Markdown heading support for `ide_find_symbol` and `ide_file_structure`, backed by the bundled JetBrains Markdown PSI/indexes. Fixes [#149](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/issues/149).
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -417,7 +417,6 @@ The plugin uses a language handler pattern for multi-IDE support:
 - `TypeHierarchyHandler` - Type hierarchy lookup
 - `ImplementationsHandler` - Find implementations
 - `CallHierarchyHandler` - Call hierarchy analysis
-- `SymbolSearchHandler` - Symbol search by name
 - `SymbolReferenceHandler` - Resolve fully qualified symbol references (e.g., `com.example.MyClass#method(String)`) to PSI elements
 - `SuperMethodsHandler` - Method override hierarchy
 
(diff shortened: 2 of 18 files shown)


hechtcarmel:
Hi,
Thank you. Please fix failing checks

hechtcarmel:
Additionally, in https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/151 we added markdown psi support, including the find symbol tool - so for example it returned markdown headings. 
This seems to no longer be the case

hechtcarmel:
(if it's problematic in your opinion to add the markdown we may leave it like that, as ide_file_structure still supports it)

vcth4nh:
Imo, I'd drop markdown from `ide_find_symbol`
The PR's whole pitch is "it does what Go to Symbol does", and the IDE itself doesn't do markdown there. Headings are still one ide_search_text away (they live in the word index), and `ide_file_structure` covers @pancsta's original #149 

But it's your plugin. Your call on whether to keep it or leave it.

hechtcarmel:
You are right, no need.
Especially since it's not like it's been up for over a day and someone relies on it somehow.

Please fix the minor test issue and i'll merge

vcth4nh:
Then I'll drop `testFindSymbolToolFindsMarkdownHeadings`. It asserted `ide_find_symbol` returns markdown headings, which is exactly the behavior we just agreed to remove. That was the only thing failing CI

hechtcarmel:
merging. I'll deploy another small pr to cleanup residuals of the markdown find symbol tool in some markdown files and tool descriptions and will submit it to jetbrains.
And thanks for the fix!

PR #170, opened 2026-05-03
Add Rider C# handlers

### LINKED ISSUE #167: Rider support
Hey,

I enjoyed the plugin so much I made a fork with Rider support - https://github.com/AsafMah/jetbrains-index-mcp-plugin/pull/1

The issue is, truthfully, it's all vibe-coded.

It has been long and painful, as the rider/reshaper plugin api is not well documented and hard to use, but it did pass a bunch of tests, and I'm going to use my fork as my daily driver.

I would like to merge it into your repo eventually, but I won't dump a ton of code I don't even understand as a PR, so maybe we can try to come up with a work plan to fit it in.

AsafMah:
## Summary

Replaces the previous frontend-only navigation bridge for Rider with a real **in-process ReSharper backend** integration following the [Rider plugin SDK pattern](https://plugins.jetbrains.com/docs/intellij/rider.html). C# and F# now have parity with the other supported languages: position- and symbol-based semantic operations (`find_definition`, `find_references`, `find_class`, `type_hierarchy`, `call_hierarchy`, `find_implementations`, `find_super_methods`, `refactor_rename`, `move_file`) execute against ReSharper PSI/caches in the same process as Rider, instead of the limited frontend heuristics earlier versions used.

The plugin now ships a ReSharper backend assembly (`ReSharperPlugin.IndexMcp.dll`) packaged under `dotnet/` in the plugin ZIP and bound through the rd protocol (`IndexMcpModel`).

Single version bump: **4.17.0 → 4.18.0**. Full Added / Changed / Fixed / Removed / Internal / Known-issues breakdown is in `CHANGELOG.md` under `## [Unreleased]`.

## Highlights

- **`language` + `symbol` form for C# / F#** — `language: "C#"` (or `"CSHARP"` / `"F#"` / `"FSHARP"`) plus a fully qualified symbol (e.g. `Namespace.TypeName#Method`, `Namespace.TypeName.Property`) on `find_definition`, `find_references`, `find_implementations`, `find_super_methods`, `call_hierarchy`.
- **ReSharper-powered rename and move** — `refactor_rename` runs a ReSharper PSI transaction (no lexical fallback); `move_file` adjusts namespaces/imports through ReSharper.
- **`project_and_libraries` returns BCL types** — `find_class` returns source-less standard .NET / BCL types from a curated catalog.
- **AXAML / source-generator partial pollution fixed** — Resolver now ranks declarations using `IPsiSourceFile.Properties.IsGeneratedFile` / `IsNonUserFile` plus a path fallback, preferring real hand-written `.cs` / `.fs` / `.fsi` / `.fsx` files. No more `file: ""` from synthetic XAML partials, no more `Class not found` from `type_hierarchy className` on AXAML-paired classes.
- **Truthful rename oracle** — Word-boundary occurrence count diff (5s @ 100ms poll) per file, so no-op or revert renames where the new name is a substring of the old return a clear failure instead of silent `success: true`.
- **Live smoke harness** — `scripts/rider-live-smoke.ps1` exercises the full MCP tool surface against a real solution with apply + revert assertions cross-verified via `find_class`.

## CI

- Adds `actions/setup-dotnet@v4` (8.0.x) to `build`, `test`, `verify`, and the publish job — `compileDotNet` shells out to `dotnet build` and needs the SDK on `PATH`.
- Marks `prepareSandbox` and `buildPlugin` `notCompatibleWithConfigurationCache` so the project's default `./gradlew buildPlugin` (config cache enabled) succeeds.

Verified locally:
- `./gradlew clean buildPlugin --console=plain --quiet` — PASS
- `pwsh scripts/assert-rider-package.ps1 -PluginZip build/distributions/jetbrains-index-mcp-plugin-4.18.0.zip` — PASS (DLL/PDB/`IndexMcpModel.class`/`IndexMcpProtocolListener.class` all present)

## Known issues (deferred, documented in CHANGELOG)

- **Roslyn source-generator output cache invalidation** in `ide_sync_files`. After external file changes (`git reset --hard`, edits outside the IDE), generator output for AXAML inputs may still reflect pre-change identifiers; refactors planned against stale output can corrupt files. Workaround: rebuild → reload IDE → restart. Proper fix requires a generator-output invalidation API in ReSharper that has no stable public surface today.
- AXAML / WPF projects that pull a partial from a non-Avalonia generator may still see the wrong file in `find_definition`; please report the generator name + path snippet so the heuristic can be extended.


EremesNG:
It aligns with what I was currently working on 👍, I'm going to compile the PR to run tests on a couple of projects.

hechtcarmel:
I will begin testing it today.
Though i'm not C#/F#/Rider  knowledgable, so it will be mostly AI driven and making sure nothing broke in the other languages.

@EremesNG Especially since you seem more knowledgable on that, feel free to leave any comment or improvement suggestions 😄 

@AsafMah @EremesNG - would be cool if you aligned, to avoid step stepping on each other's toes.

EremesNG:
I have tested several projects with C# and F#, mainly C#, including some Web APIs and others in Unity; the PR is a good foundation to build upon.

I found some gaps between what is currently offered and the implementation for Rider, especially in "language+symbol," and I experienced a couple of freezes while stress-testing the tools on large projects. I will work on this and submit a PR to @AsafMah , so it can be included in this one.

hechtcarmel:
Fantastic. thank you both!
I started by testing on non-rider projects to see it doesn't break anything existing (didn't).

EremesNG:
Small update, I think C# is almost covered, but F# is a problem.
I have almost no experience working with F# and the ReSharper SDK treats it like another plugin; it's a headache. I'm going to try a few more things, but it will probably be out of scope or very limited. The biggest problem is that indexing seems non-existent. For example, each `ide_find_references` call is a cold search that takes more than 5 minutes to resolve in large projects, causing some freezes in Rider. If Rider is closed and reopened, it's cold searches again.

AsafMah:
@EremesNG I'm for removing F# for now

hechtcarmel:
Hi,
Making sure there isn't miscommunication - if you both think its stable and ready for review and merge please leave a message

EremesNG:
> Hi, Making sure there isn't miscommunication - if you both think its stable and ready for review and merge please leave a message

@hechtcarmel  No no, the current PR is not stable, I'm working on stabilizing C#, but F# will remain in beta/unstable, it has too many limitations. What I found is that F# is a plugin within the ReSharper SDK, and the SDK documentation itself is almost nonexistent from JetBrains 🫠
Hopefully, I'll send a PR with the changes to @AsafMah in the next few hours

EremesNG:
@hechtcarmel PR sent to @AsafMah https://github.com/AsafMah/jetbrains-index-mcp-plugin/pull/2
C# I think it could be ready, but it's not yet at the Java/Kotlin level. Unless there is better research or documentation for the ReSharper SDK, it could be improved.
F# needs further research. From my side, I couldn't find a way to overcome many of the limitations.

hechtcarmel:
My development computer died and is in repair so i'll test it and merge hopefull by end of week at the latest

hechtcarmel:
Just updating that it looks like i won't be able to review until mid next week to next weeked at most.

hechtcarmel:
Hi,
Starting to get back.
AI-assisted (but not only ;) )  review - 

1. I suggest removing F#, if it's not "there" yet, instead of having broken tools. wdys?

2. The biggest concern is that the ReSharper backend read endpoints are doing PSI/project/search work inside plain Task.Run calls. For example HandleFindTypes, HandleFindDefinition, HandleFindReferences, type hierarchy, implementations, call hierarchy, super methods, and file structure all follow this pattern in IndexMcpBackendHost.cs. Those paths enumerate project files, walk PSI trees, resolve symbols, and call ReSharper search APIs such as Finder.FindReferences, but they do not take a ReSharper read lock. Rename already goes through a write-lock path, so the same kind of locking discipline should be applied to the read side with the ReSharper/Rider read-lock APIs. Without that, these calls can race with PSI/project updates or hit read-access assertions in Rider.

3. (AI comment) There is also a crash path in C# type search with project_and_libraries. HandleFindTypes combines project declarations with indexed/library type elements, then sorts with:

.ThenBy(type => IsTestPath(GetDeclarationPath(type)) ? 1 : 0)

GetDeclarationPath can return null for metadata/library symbols, but IsTestPath immediately calls path.Replace(...). So a library-backed result can turn a type search into a backend exception. Making IsTestPath null-safe should fix the immediate problem, but this also needs a backend test because library search is one of the important C# cases here.

4. Another issue is that the frontend and backend disagree on C# language IDs. The Kotlin side accepts C#, CSHARP, and CSharp, but the backend mostly only accepts C# when filtering real .cs declarations. That means a request using language: "CSHARP" can be treated as handled by the Rider bridge and still return no C# results. The backend should normalize the same aliases the frontend advertises.

5. (AI comment, as i'm not a C# guy. May be wrong) Symbol lookup for C# methods also looks too weak right now. ResolveSymbol compares the requested symbol only to Namespace.Type.Member or the short name, and GetQualifiedName does not include method parameters. So something like My.Namespace.Service#Run(string) will not resolve, while My.Namespace.Service.Run is ambiguous if there are overloads. If symbol-mode C# lookup is supported, it needs to parse member signatures or at least detect ambiguity instead of picking whichever declaration appears first.

6. (AI comment, as i'm not a C# guy. May be wrong) findImplementations has a semantic mismatch for method targets. The tool contract says it can find implementations of an interface/abstract method, but the C# backend converts any ITypeMember to its containing type and then runs an inheritor search. So asking for implementations of IFoo.Do() returns implementing classes, not the concrete Do() members. Type targets and member targets probably need separate code paths here.

7. There are also some response formatting inconsistencies. Definition and references convert backend file paths through displayPath, but type hierarchy, implementations, call hierarchy, and super methods return raw RD filePath values. That can leak absolute local paths and makes the C# tool output inconsistent depending on which tool was called. The converters should use display paths for returned DTOs while keeping backend paths only for internal recursive calls.

Ch1ngg:
When do you think this PR might be merged? I’m really looking forward to trying out the new feature. :)

hechtcarmel:
@EremesNG @AsafMah can you comment if you are still working on it or not?

EremesNG:
> @EremesNG @AsafMah can you comment if you are still working on it or not?

I sent a PR https://github.com/AsafMah/jetbrains-index-mcp-plugin/pull/2 to @AsafMah but it hasn't been approved. I could send the PR directly to this repository, but I didn't want to duplicate this PR.
I will send the PR to https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/tree/feat/rider-support if you agree.

hechtcarmel:
I suggest we git @AsafMah 1-3 days to respond and if not let's duplicate the PR.
Though i cannot really review it beyond "big obvious issues ai will find", but as far as i care as long as it doesn't break existing functionality it's good enough and we can improve

AsafMah:
Ok, the latest version removes F# and answers all comments.
Let's try to drive it before

hechtcarmel:
ping when ready 😄

AsafMah:
Ok I think I'm finally done :)

hechtcarmel:
Again, i'm having a hard time reviewing as it's not my area.
I'm posting an aggregated ai review from Codex+Claude.
**IF SOMETHING IS WRONG/NITPICKY THEN TELL, I DON'T WANT FOR AI SLOPE REIEW TO BLOCK THIS**
But do go over it and check if it has merit


## Blocking / should fix before merge

1. `scripts/rider-live-smoke.ps1` can delete local work

The smoke script calls `Reset-WorkspaceClean` before it installs the plugin or checks whether `-TestRename` was requested:

- `scripts/rider-live-smoke.ps1:154-164`
- `scripts/rider-live-smoke.ps1:190-197`

If `ProjectPath` points at a dirty git checkout, it runs `git reset --hard`. That happens even for the default/read-only smoke path, before the later message saying `-TestRename` should only be used on disposable copies.

This should fail closed instead: print the dirty files and exit unless the user passes an explicit destructive flag like `-ResetWorkspace` or `-ForceDisposableWorkspace`.

2. `prepareSandbox` puts the Rider backend DLL in the wrong plugin directory

`prepareSandbox` copies the .NET backend to `"$pluginName/dotnet"`:

- `build.gradle.kts:352-356`
- `gradle.properties:4`
- `settings.gradle.kts:22`

`pluginName` is `IDE Index MCP Server`, but the sandbox plugin folder is `jetbrains-index-mcp-plugin`. I verified this by running `prepareSandbox`; the DLL landed here:

```text
build/idea-sandbox/IU-2025.3/plugins/IDE Index MCP Server/dotnet/ReSharperPlugin.IndexMcp.dll
```

while the actual plugin folder was:

```text
build/idea-sandbox/IU-2025.3/plugins/jetbrains-index-mcp-plugin
```

So local Rider sandbox / `runIde` testing can start the plugin without the backend where Rider will look for it. The ZIP path is different and looks correct, which means CI can pass while sandbox testing is broken. The sandbox copy should use the real sandbox plugin directory name, not `pluginName`.

3. C# rename can silently fall back to the generic IntelliJ rename path

For C# symbol rename, the code routes into the Rider frontend rename path:

- `RenameSymbolTool.kt:289-297`

But the Rider planning block catches `Exception`, logs it, and continues:

- `RenameSymbolTool.kt:1725-1754`

If planning throws, `riderExecutionPlan` stays null. Then `highLevelHandlerLane` is false and the code runs the generic `executeRename(...)` path:

- `RenameSymbolTool.kt:1780-1819`

That is the unsafe fallback this PR is trying to avoid for Rider C#. A missing editor/data context, planning failure, or cancellation can degrade into the generic IntelliJ `RenameProcessor`, and the mutation probe can still report success if file text changed.

For Rider C# rename, this should fail closed with a clear `unsupported_context` / `needs_active_editor` result. Also, IntelliJ cancellation exceptions should be rethrown, not swallowed as ordinary planning failures.

## Correctness issues that can return wrong or incomplete results

4. Rider C# `find_references` applies the limit before scope filtering

The frontend sends `scope` and `limit` to the backend:

- `RiderDotNetHandlers.kt:620-638`

The backend collects only `effectiveLimit` raw references first, and only then applies `MatchesPathScope`:

- `IndexMcpBackendHost.cs:431-453`
- `IndexMcpBackendHost.cs:3526-3546`

That means scoped queries can under-return. Example: `scope=project_test_files` can miss test references if the first 200 raw references are production files.

There is also no proper truncation signal. `MaxResults` is 200, `GetEffectiveResultLimit` clamps to that, `RdFindReferencesResult` has no `hasMore`, and the Kotlin handler does not expose the backend `totalCount`:

- `IndexMcpBackendHost.cs:61`
- `IndexMcpBackendHost.cs:1899-1905`
- `protocol/src/main/kotlin/model/rider/IndexMcpModel.kt:117-120`
- `RiderDotNetHandlers.kt:655-672`

The scope filter should happen before counting toward the limit, and the Rider references path needs an explicit truncation/pagination story.

5. Rider C# type hierarchy and implementations ignore `scope`

Kotlin passes the requested scope:

- `RiderDotNetHandlers.kt:855-859`
- `RiderDotNetHandlers.kt:1384-1412`
- `TypeHierarchyTool.kt:162-168`
- `FindImplementationsTool.kt:133`

But the backend does not apply it for type hierarchy or implementations:

- `IndexMcpBackendHost.cs:1583-1621`
- `IndexMcpBackendHost.cs:1627-1713`

Both paths search the whole solution and truncate without filtering by `project_production_files` / `project_test_files`. That violates the public tool contract and can return test implementations in production-only queries or production implementations in test-only queries.

6. Rider rename success / affected-file reporting is too weak

The frontend mutation check treats any target-file text change as success:

- `RenameSymbolTool.kt:872-887`

That does not prove the requested rename happened. A wrong edit, reformat, partial dialog action, or unrelated file text change can satisfy the check.

The high-level Rider lane also builds `affectedFiles` from a whole-project mtime diff:

- `RenameSymbolTool.kt:1781-1889`
- `RenameSymbolTool.kt:3303-3385`

That diff runs before `commitDocuments` / `saveAllDocuments`:

- `RenameSymbolTool.kt:1911-1915`

So unsaved IDE documents can make the diff collapse back to one file, while unrelated concurrent writes can create false positives. The success check should verify the exact requested rename against PSI/document state, and `affectedFiles` should not be a whole-project timestamp diff presented as semantic truth.

7. C# member symbol matching is case-insensitive

C# is case-sensitive, but member symbol resolution uses `OrdinalIgnoreCase`:

- `IndexMcpBackendHost.cs:2435-2455`

The parser also treats a member whose name equals the containing type leaf as a constructor, also case-insensitively:

- `IndexMcpBackendHost.cs:2650-2654`

This can make distinct members like `Value` and `value` ambiguous/wrong, and can make a non-constructor same-named member unaddressable through symbol mode. These comparisons should follow C# semantics.

8. Rider safe delete uses raw regex matching for usages

`SafeDeleteMutationExecutor.FindBlockedUsages` scans all `.cs` files with a word-boundary regex:

- `SafeDeleteMutationExecutor.cs:85-112`

That will false-block on comments, strings, generated files, or unrelated same-named symbols, and can also miss real semantic usages. Safe delete should use ReSharper reference search for the resolved declared element rather than raw text matching.

## Smaller but still worth fixing

9. Version metadata is inconsistent

There are multiple version values right now:

- `gradle.properties:7` is `4.20.1`
- `McpConstants.SERVER_VERSION_FALLBACK` is `4.20.0`
- `IndexMcpBackendHost.BackendVersion` is `4.18.0`
- `CHANGELOG.md` has current fixes under `[Unreleased]` and then `[4.20.0]`, with no `[4.20.1]`

Runtime `serverInfo.version` is safer now because it reads the plugin descriptor, but backend diagnostics, fallback contexts, unit-test contexts, and release notes can still disagree.

10. `FindDefinitionTool` drops useful backend errors

When Rider handles a C# definition request but returns no value, the tool always emits a generic error:

- `FindDefinitionTool.kt:70-83`

It should surface `riderDefinition.errorMessage` when present, like the references path does, so users can see whether the issue was ambiguity, source-unavailable metadata, backend failure, etc.

11. The top-level server description still omits C#

The tool descriptions mention C#, but the top-level `SERVER_DESCRIPTION` still lists Java/Kotlin/Python/JS/TS/Go/PHP/Rust/Markdown and does not mention Rider/C#:

- `McpConstants.kt:91`

That is low-risk but visible to MCP clients and agents, so it should be updated.

## Things I checked but would not block on

I did not include a few suspected issues as blockers because they appear to be dead-code-only or not on the active path right now:

- Backend rename disk-read verification is a real smell, but `shouldUseRiderBackendRename(...)` currently returns `false`, so that backend rename lane is disabled.
- `ExactTargetResolver` has brace-depth fragility for member targets in block-scoped namespaces, but the current safe-delete path only supports type/file targets and backend symbol rename is disabled. I would add regression coverage before enabling member mutation support.
- `MoveMutationExecutor` always returns unsupported after planning, but the active Kotlin C# move path is Rider dialog automation. This is cleanup unless the backend move lane gets wired in.

Overall: the architecture is promising, but I would fix the three blocking safety/packaging/rename-fallback issues before merging, then address the scoped-result and rename-oracle issues before release. Those are the areas most likely to make an MCP client take a wrong action with high confidence.

AsafMah:
Ok + removed dead code from the old rename refactoring

hechtcarmel:
Hi,
I tested it manually - with my focus being making sure it doesn't break existing functionality (as C# i cannot really test). Unfortunately, it does.

It breaks "rename" functionality. I got the following sometimes when trying to rename:

image

I confirmed this doesn't happen without those changes.

The suspicious change is in `RenameSymbolTool`.

The new resolver records both:

- a local/enclosing candidate, for example the method containing the cursor
- a reference-resolved candidate, for example the method being called at the cursor

But then it selects the local candidate first:

```kotlin
val selectedCandidate = localCandidate ?: referenceCandidate
```

That means when the cursor is on something like this:

```java
void caller(RenameTarget target) {
    target.process();
}
```

the resolver can prefer the enclosing `caller` declaration over the referenced `process` declaration. That is the opposite of what users expect when asking to rename a usage.

I think the safer order is:

1. direct declaration under the cursor
2. reference-resolved declaration under the cursor
3. enclosing declaration fallback

If Rider needs the enclosing fallback for a specific C# container-like case, that should be gated to that case. It should not change the normal reference rename behavior for every language.

## Regression 2: `line=0,column=0` now renames the file

This PR also changes rename mode resolution:

```kotlin
val normalizedLine = optionalIntArg(arguments, "line")?.takeIf { it > 0 }
val normalizedColumn = optionalIntArg(arguments, "column")?.takeIf { it > 0 }
```

Because explicit `0` is normalized to absent, this request:

```json
{
  "file": "src/main/java/com/example/File.java",
  "line": 0,
  "column": 0,
  "newName": "Renamed.java"
}
```

now becomes a file rename.

I reproduced that with the PR installed: `line=0,column=0` renamed the fixture file. With the old plugin installed, the same kind of request returned:

```text
No element found at the specified position
```

and the file was not renamed.

This is another mutating contract break. A malformed symbol-rename request should fail. It should not silently become a file rename.

Suggested fix:

- omitted or blank `line`/`column` can mean file rename, if that is the intended API
- explicit numeric `0` or negative values should be invalid position input
- add a regression test that `line=0,column=0` does not file-rename

## Regression 3: mixed symbol and position inputs no longer fail

Before this PR, a request that provided both `language+symbol` and `file+line+column` was rejected as ambiguous. Now the shared lookup resolver makes complete position win:

```kotlin
return when {
    hasCompletePositionTarget(arguments) -> LookupModeState.POSITION
    hasCompleteSymbolTarget(arguments) -> LookupModeState.SYMBOL
    else -> LookupModeState.MISSING
}
```

I reproduced the behavior change:

- PR version: silently returned the position target and ignored the supplied symbol.
- Old plugin: rejected the request with:

```text
Cannot specify both language+symbol and file+line+column. Use one or the other.
```

This is less dangerous than the rename bug because it is not directly mutating, but it can still make agents act on the wrong symbol and makes malformed requests harder to detect. If this behavior change is intentional, I think it should be explicitly documented and tested for all affected tools. Otherwise, the old conflict behavior should be restored for shared/non-Rider lookup.

AsafMah:
Sorry!

Fixed the regressions, and also added smoke tests for IDEA.

hechtcarmel:
Again, send to AI to check C# related stuff and it turned out some things. You judge if those are real of not, as i'm not able to check well myself.
About changelog.md - just do take what's in main branch, i'll update it when i update the version numbering

## Should fix before merge

### 1. Rider C# definition/references still bypass the new target conflict validation

The shared navigation resolver now correctly rejects requests that include both:

- `file` + `line` + `column`
- `language` + `symbol`

with the intended `Cannot specify both language+symbol and file+line+column` error.

But `ide_find_definition` and `ide_find_references` call the Rider backend before they fall through to that shared resolver. The Rider backend path still accepts both complete target groups and picks one, usually the position target first.

So a request like this can still silently run instead of failing:

```json
{
  "file": "src/Foo.cs",
  "line": 10,
  "column": 5,
  "language": "C#",
  "symbol": "Other.Namespace.Bar"
}
```

Why this matters: agents often send default or stale fields. If both target groups are present, choosing one silently can send the tool to the wrong symbol. Also, the PR explicitly says this class of request is rejected now, but Rider C# definition/reference do not fully follow that contract yet.

Suggested fix:

- In `FindDefinitionTool` and `FindUsagesTool`, compute `resolveLookupMode(arguments)` before calling `RiderBackendSemanticService`.
- If it is `AMBIGUOUS`, return `ErrorMessages.SYMBOL_AND_POSITION_EXCLUSIVE` immediately.
- Add tests proving Rider C# dual-target requests fail before the backend is invoked.
- Optionally make `buildSemanticTarget` reject dual complete targets too, as defense in depth.

Relevant code:

- `FindDefinitionTool.kt`
- `FindUsagesTool.kt`
- `RiderDotNetHandlers.kt`

### 2. The new .NET tests are not wired into CI

The PR adds `testDotNet`, but `./gradlew check` does not depend on it. The GitHub Actions test job runs:

```bash
./gradlew check
```

I ran a dry-run locally and `testDotNet` was not in the task graph.

This means the .NET backend tests can pass locally but still not be part of the required CI gate.

Suggested fix:

```kotlin
tasks.named("check") {
    dependsOn(testDotNet)
}
```

or add an explicit CI step for:

```bash
dotnet test src/dotnet/ReSharperPlugin.IndexMcp.sln -c Release
```

I prefer wiring it into `check`, so local and CI behavior stay aligned.

## Also worth fixing before release

### 3. .NET build outputs dirty the checkout

Running the .NET tests creates untracked `bin/` and `obj/` directories under `src/dotnet/...`.

Suggested `.gitignore` addition:

```gitignore
src/dotnet/**/bin/
src/dotnet/**/obj/
```

### 4. Docs still describe old target precedence behavior

Some docs/tool descriptions still say position targets "take precedence" when both target groups are present. The intended behavior is now to reject both complete target groups.

This should be updated in the tool descriptions, `USAGE.md`, and the skill reference so clients do not learn the wrong behavior.

### 5. Changelog history needs cleanup

The current changelog has duplicate/out-of-order version sections, including duplicate `4.17.0` entries and older Rider/F# notes that are now stale after F# removal.

The `[Unreleased]` section is useful, but the historical section should be cleaned up before publishing.

## Bottom line

This is close, and I think the direction is good. The current version is a real improvement over the earlier PR state.

I would merge after:

1. Fixing the Rider C# dual-target validation bypass.
2. Wiring the .NET backend tests into CI.
3. Cleaning up the small release hygiene issues above.

I did not run live Rider smoke locally, so I am relying on static review plus the unit/.NET/build verification listed above.

PR #174, opened 2026-05-06
Add support for Scala 2

ilx:
This is an attempt to add basic support for Scala 2.x.
Disclosure: I used opencode to get a fast start and pointed it at intellij-scala source code to make it a bit more reliable.
 
I would like to investigate what can / should be done for Scala 3, but right now all of our projects are Scala 2.13 and Pekko.

Please give it a try and feedback.

P.S. Reason for .ignore file content: I discovered that unix tools like ripgrep, grep invoked by coding agents (opencode in this case) obey ignored / hidden files which prevents them from finding plans inside .opencode or discovering internal instructions and updating those. But, those tools also inspect .ignore files and if we un-ignore those folders / files then it will again work for coding agent invocations.

P.P.S Thank you for the plugin, during period when official MCP server was broken your plugin saved the day.


hechtcarmel:
Hi,
Thank you for the PR.
I'll look it at and test it in the next couple of days

hechtcarmel:
Hi,
Since Scala is not my language i have to rely on AI.
If from your Scala understanding something here is wrong, please tell me.
After you fix those, i will do another round of verifications and let's merge.
ai report:

**Blocking issues**
1. scope is not applied consistently
The tool API exposes scope, and most existing handlers thread the resulting GlobalSearchScope through their supertype and callee traversal. The Scala handler only applies it in some places.

Specific gaps:

getSupertypes does not receive a searchScope, so default project_files can still include library supertypes.
findCalleesRecursive does not receive a searchScope, so callees ignores project_files, project_test_files, and project_production_files.
Scala should thread GlobalSearchScope through supertypes and callees, and filter navigation elements consistently before returning nested results.

**2. Scala result paths should use the shared path formatter**
BaseScalaHandler.getRelativePath currently uses manual string prefix removal:

file.path.removePrefix(basePath).removePrefix("/")
Other handlers use ProjectUtils.getToolFilePath(project, file), which handles content roots, workspace layouts, and library/navigation paths consistently. Scala should use the same helper so tool outputs can round-trip into later position-based lookups.

**3. Please add Scala runtime coverage or at least documented manual verification**
Because the Scala support uses reflection, tests are more important than usual. The current suite does not exercise the reflected Scala PSI API at all.

At minimum, I would like to see a small Scala 2 fixture or a clearly documented manual verification plan covering:

trait/class type hierarchy
method override implementations
callers and callees
super methods
file structure for classes, traits, objects, case classes, val, and var

**4. Docs and tool descriptions need to be updated**
The PR updates part of plugin.xml, but the build patches the plugin description from README.md, and the public docs/tool descriptions still omit Scala from the supported language lists.

Please update:

README.md
USAGE.md
the MCP tool descriptions for hierarchy/implementations/call hierarchy/file structure
the companion skill tools reference, if applicable
Smaller concern
The file structure handler likely reports common Scala fields as unknown, because it treats ScValue and ScVariable as if each declaration has a single name(). Scala val/var declarations can define multiple bindings, and constructor val/var parameters probably need separate handling.

ilx:
Thanks. I will try to find some time to review it, most likely over weekend

hechtcarmel:
Hi,
Just making sure we are aligned - leave a message if you want me to re test it, otherwise I don't know if more pushes are expected

ilx:
> Hi, Just making sure we are aligned - leave a message if you want me to re test it, otherwise I don't know if more pushes are expected

Sorry for not responding, work and life got in the way of PR. I still plan to work on it.
I will leave a message once I consider it done.

hechtcarmel:
converted to draft

ilx:
@hechtcarmel Opus reported: 

> pluginSinceBuild = 253 and platformVersion = 2025.1.3, there's a contradiction

Should these be changed/updated? Also every time I re-open project IntelliJ complains:

```
Kotlin: [Internal Error] org.jetbrains.kotlin.util.FileAnalysisException: While analysing intellij-scala-mcp/jetbrains-index-mcp-plugin/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/history/CommandModels.kt:34:1: java.lang.NoClassDefFoundError: org/jetbrains/kotlin/com/intellij/psi/PsiElement
	at org.jetbrains.kotlin.util.AnalysisExceptionsKt.wrapIntoFileAnalysisExceptionIfNeeded(AnalysisExceptions.kt:57)
	at org.jetbrains.kotlin.fir.FirCliExceptionHandler.handleExceptionOnFileAnalysis(Utils.kt:251)
	at org.jetbrains.kotlin.fir.pipeline.AnalyseKt.runCheckers(analyse.kt:68)
	at org.jetbrains.kotlin.fir.pipeline.FirUtilsKt.resolveAndCheckFir(firUtils.kt:77)
	at org.jetbrains.kotlin.cli.pipeline.jvm.JvmFrontendPipelinePhase.executePhase(JvmFrontendPipelinePhase.kt:167)
	at org.jetbrains.kotlin.cli.pipeline.jvm.JvmFrontendPipelinePhase.executePhase(JvmFrontendPipelinePhase.kt:56)
	at org.jetbrains.kotlin.cli.pipeline.PipelinePhase.phaseBody(PipelinePhase.kt:68)
	at org.jetbrains.kotlin.cli.pipeline.PipelinePhase.phaseBody(PipelinePhase.kt:58)
	at org.jetbrains.kotlin.config.phaser.SimpleNamedCompilerPhase.phaseBody(CompilerPhase.kt:215)
	at org.jetbrains.kotlin.config.phaser.NamedCompilerPhase.invoke(CompilerPhase.kt:111)
	at org.jetbrains.kotlin.backend.common.phaser.CompositePhase.invoke(PhaseBuilders.kt:28)
	at org.jetbrains.kotlin.config.phaser.CompilerPhaseKt.invokeToplevel(CompilerPhase.kt:62)
	at org.jetbrains.kotlin.cli.pipeline.AbstractCliPipeline.runPhasedPipeline(AbstractCliPipeline.kt:106)
	at org.jetbrains.kotlin.cli.pipeline.AbstractCliPipeline.execute(AbstractCliPipeline.kt:65)
	at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.doExecutePhased(K2JVMCompiler.kt:61)
	at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.doExecutePhased(K2JVMCompiler.kt:36)
	at org.jetbrains.kotlin.cli.common.CLICompiler.execImpl(CLICompiler.kt:80)
	at org.jetbrains.kotlin.cli.common.CLICompiler.exec(CLICompiler.kt:337)
	at org.jetbrains.kotlin.daemon.CompileServiceImpl.compile(CompileServiceImpl.kt:1611)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.rmi/sun.rmi.server.UnicastServerRef.dispatch(UnicastServerRef.java:360)
	at java.rmi/sun.rmi.transport.Transport$1.run(Transport.java:200)
	at java.rmi/sun.rmi.transport.Transport$1.run(Transport.java:197)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:714)
	at java.rmi/sun.rmi.transport.Transport.serviceCall(Transport.java:196)
	at java.rmi/sun.rmi.transport.tcp.TCPTransport.handleMessages(TCPTransport.java:598)
	at java.rmi/sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.run0(TCPTransport.java:844)
	at java.rmi/sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.lambda$run$0(TCPTransport.java:721)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:400)
	at java.rmi/sun.rmi.transport.tcp.TCPTransport$ConnectionHandler.run(TCPTransport.java:720)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.NoClassDefFoundError: org/jetbrains/kotlin/com/intellij/psi/PsiElement
```

this is supposedly fixed in newer version of kotlin. Do you have an advice how to get rid of it?

hechtcarmel:
Hi,
good catch.
should be -`  pluginSinceBuild = 253`,` platformVersion = 2025.3`.
I'll fix it on a later version, though you can go ahead and do it as part of yours if it turns out to solve the issue you are facing.

About the actual error - this is the first report. Could you first switch to master, clean caches, to make sure it doesn't come from your changes?

ilx:
> Hi, good catch. should be -` pluginSinceBuild = 253`,` platformVersion = 2025.3`. I'll fix it on a later version, though you can go ahead and do it as part of yours if it turns out to solve the issue you are facing.

It is not required for the error. It's mostly something I noticed in logs while I investigated platform/plugin compatibility. 
 
> About the actual error - this is the first report. Could you first switch to master, clean caches, to make sure it doesn't come from your changes?

I tried Intellij build (i.e. I did not use Gradle within IntelliJ to build a project). My "workaround" is to use Gradle for CLI build and from within IntelliJ. So, I guess I have my "workaround". It's just that it looks like a known issue with kotlin version within IntelliJ.  I did not notice it until now because until now I used Gradle builds.


Another (unrelated) question: I have another branch where I declared Scala plugin as an optional dependency and instead of using reflection it uses direct calls to Scala plugin. Now, since there is check if plugin is installed everything works even if I install a plugin into IntelliJ instance that does not have Scala plugin installed. Should that be preferred approach? I ask because Python uses reflection, so maybe there is some reason not to use direct calls?

Btw. a reason for that alternative version was exploration of "how to write a test without mocking a lot".

hechtcarmel:
Yes, direct calls are fine and proably preferable for Scala, as long as Scala is declared as an optional plugin dependency and the Scala specific handler class is not loaded when the Scala plugin is absent.

Python and others being like that is a result of it being the first plugin i coded and not knowing the intellij sdk enough at the time (+ai making it so i don't need to fully understand it).

PR #175, opened 2026-05-12
Handle JS/TS renames and enhance symbol resolution

EremesNG:
### Description

This pull request introduces updates and improvements for JavaScript and TypeScript development:

- **Fix**: Improved handling of JS/TS file renames to automatically retarget imports, removing the need for user dialog intervention.
- **Feature**: Enhanced support for JavaScript and TypeScript symbols in WebStorm.
- **Feature**: Implemented module-qualified grammar for JavaScript/TypeScript symbol resolution, enabling more precise functionality.


hechtcarmel:
Hi,
Thank you for your contribution.
My development computer died and is in repair so i'll test it and merge hopefull by end of week at the latest

hechtcarmel:
Just updating that it looks like i won't be able to review until mid next week to next weeked at most.

hechtcarmel:
Hi,
Starting to catch up.

I think there is an issue with the current JS/TS file rename path - in `RenameSymbolTool` it uses IntelliJ's `RenameProcessor`, but then applies a manual regex post-pass to retarget JS/TS module specifiers after file rename.

That fallback is fragile and does not match the rest of the plugin architecture. The repo generally delegates semantic refactoring to IntelliJ processors/search/reference APIs (`RenameProcessor`, `MoveFilesOrDirectoriesProcessor`, `ReferencesSearch`, language-specific processors, etc.). The JS/TS import rewrite is a bespoke text parser over project files.

An example to it's fragiility is that it meases common valid syntax, for example:

```ts
import {
  leafThing,
} from './leaf';
```

It also cannot reliably model JS/TS resolver semantics such as path aliases, index resolution, extension substitution, package exports, or source files imported with runtime `.js` specifiers.


The IntelliJ Platform rename model is reference-driven:

- Rename uses the same reference discovery model as Find Usages.
- Rename calls `PsiReference.handleElementRename()` for references to the renamed element.
- Resolvable PSI references are the prerequisite for navigation, Find Usages, code completion, and rename.
- IntelliJ file rename supports searching source-code references to the file.

For JS/TS module specifiers, the WebStorm/JavaScript plugin should be the source of truth when those specifiers are PSI references.

EremesNG:
> Hi, Starting to catch up.
> 
> I think there is an issue with the current JS/TS file rename path - in `RenameSymbolTool` it uses IntelliJ's `RenameProcessor`, but then applies a manual regex post-pass to retarget JS/TS module specifiers after file rename.
> 
> That fallback is fragile and does not match the rest of the plugin architecture. The repo generally delegates semantic refactoring to IntelliJ processors/search/reference APIs (`RenameProcessor`, `MoveFilesOrDirectoriesProcessor`, `ReferencesSearch`, language-specific processors, etc.). The JS/TS import rewrite is a bespoke text parser over project files.
> 
> An example to it's fragiility is that it meases common valid syntax, for example:
> 
> ```ts
> import {
>   leafThing,
> } from './leaf';
> ```
> 
> It also cannot reliably model JS/TS resolver semantics such as path aliases, index resolution, extension substitution, package exports, or source files imported with runtime `.js` specifiers.
> 
> The IntelliJ Platform rename model is reference-driven:
> 
> * Rename uses the same reference discovery model as Find Usages.
> * Rename calls `PsiReference.handleElementRename()` for references to the renamed element.
> * Resolvable PSI references are the prerequisite for navigation, Find Usages, code completion, and rename.
> * IntelliJ file rename supports searching source-code references to the file.
> 
> For JS/TS module specifiers, the WebStorm/JavaScript plugin should be the source of truth when those specifiers are PSI references.

My bad! 🙏

I removed the manual regex/text post-pass for JS/TS module specifiers.

The remaining JS/TS file-rename path now collects IDE-resolved references with `ReferencesSearch` before the rename and retargets them afterward with `PsiReference.bindToElement`, so rewriting is handled by IntelliJ’s reference machinery rather than custom parsing.

I also added an explicit `targetType` parameter to `ide_refactor_rename` to disambiguate file renames from symbol renames. Previously the tool inferred the mode from whether `line`/`column` were present, which was easy to make ambiguous for clients and fragile as the rename behavior grew more specialized. With `targetType: "file"` or `targetType: "symbol"`, clients can state the intended operation directly while preserving backward compatibility.

This keeps the JS/TS path semantic and avoids brittle handling of multiline imports or module specifier syntax. I also added behavior coverage for static, side-effect, and dynamic imports.

hechtcarmel:
Great. 
3 small things before we can merge it :

1. conflicts - just resolve, no need to update version. i'll do it later.

2.The new JS/TS symbol resolver can look in the wrong folder in workspace or monorepo-style projects.

The rest of the plugin usually exposes file paths relative to the project root. For example, if the project root is `/repo` and the real file is:

`/repo/packages/app/src/user.ts
`then a valid symbol id could look like:

`packages/app/src/user#User`
But the new resolver uses module content roots instead of the project root whenever modules exist. If the module root is `/repo/packages/app`, it can end up checking:

`/repo/packages/app/packages/app/src/user.ts` 
That file does not exist, so the tool reports that the symbol was not found even though the symbol id is valid. I think the resolver should include project.basePath first, then fall back to module content roots, matching the existing file resolution behavior elsewhere in the plugin.

Something like this would be safer:

`val roots = (listOfNotNull(project.basePath) + ProjectUtils.getModuleContentRoots(project)).distinct()`

3.  `targetType=file` says that line and column are ignored, but the tool still parses them before it checks that this is a file rename.

That means a client can reasonably send:
```

{
  "filePath": "src/foo.ts",
  "targetType": "file",
  "line": null,
  "column": null,
  "newName": "bar.ts"
}
```
and the request can fail while trying to parse line/column, even though those fields should not matter for file rename. I think the code should read targetType first, skip coordinate parsing for file renames, and only require valid positive line/column values for targetType=symbol.

EremesNG:
Apologies for the delay, I believe everything should be covered. I await your feedback.

hechtcarmel:
Will look. 
Development laptop hade issues and is at IT, may take a day or two

hechtcarmel:
Tested manually, ai summarized. 
There are still fundumental issues (and also some conflicts):


on the plus side - the two of the earlier review points are fixed.

## 1. Barrel/re-export lookup does not look like it resolves to the real declaration

The new JS/TS `language + symbol` support appears to scan the target module file for a matching `PsiNamedElement` and return that element directly. I don't see a path that follows the export reference with `reference.resolve()` or an equivalent WebStorm PSI API.

Simple example:

```ts
// index.ts
export { loadConfig } from "./config";

// config.ts
export function loadConfig() {
  return {};
}
```

For a request like:

```json
{
  "language": "TypeScript",
  "symbol": "index#loadConfig"
}
```

I would expect `ide_find_definition` to land on `config.ts`, at the actual `function loadConfig` declaration.

From the current resolver, I think it can instead return the export specifier in `index.ts`, or fail for `export * from "./config"` depending on how the JS PSI exposes that syntax. That would make `find_definition`, `find_references`, implementations, and call hierarchy work from the wrong element.

This is also how the IntelliJ SDK expects references to work: the reference element and the resolved target are different things. The code needs to resolve the export/import reference to the declaration, or clearly document that barrel re-exports are unsupported.

I also tested this with the installed plugin in a real open project. Position-based `ide_find_definition` on a `loadConfig()` usage resolved through the barrel to the real `config.ts` function, but `language="TypeScript", symbol=".../index#loadConfig"` returned the `index.ts` export specifier. So the IDE has the semantic information; the new symbol-form resolver is not using it here.

## 2. Plain imports can be mistaken for exports

`isExportCandidate()` currently accepts class names containing `JSImportSpecifier`. That seems too broad, because an import specifier is not necessarily an export from the current module.

Example:

```ts
// consumer.ts
import { PluginName } from "./plugin";

console.log(PluginName);
```

`consumer#PluginName` should not resolve as a named export from `consumer.ts`, because the file only imports `PluginName`; it does not export it.

The docs already say local import aliases should use `file + line + column`, not `language + symbol`, so I think this should return not found/unsupported rather than a misleading success.

## 3. The JS/TS file rename is documented as one undo, but it is implemented as two commands

The PR says a single undo should revert the whole JS/TS file rename.

The implementation runs `RenameProcessor` first, then runs a separate `WriteCommandAction` to retarget JS/TS imports. Based on the IntelliJ Platform command model, that is a separate undoable command.

Simple example:

```ts
// app.ts
import { run } from "./old-name";
```

If we rename:

```text
old-name.ts -> new-name.ts
```

and the import is retargeted to:

```ts
import { run } from "./new-name";
```

then one undo should revert both the file rename and the import update if that is the public contract. With the current split, one undo can leave the project half-reverted.

Either the retargeting should participate in the same refactoring command, or the docs/changelog should stop promising single-undo behavior.

I also saw a result-reporting problem while testing the installed plugin. File rename did update imports from `./config` to `./settings`, but the tool response still said both importers could not be auto-retargeted and listed them under `unretargetedImporters`. That makes the structured result hard to trust for automation.

EremesNG:
Thanks for the solid review! I was playing around with a few different ways to handle the file rename and totally forgot to update the docs, my bad on that. I'm diving into points 1 and 3 right now to get them sorted out. 

That said, I ran into a bit of a wall trying to reproduce point 2. 

### 2. Plain imports mistaken for exports (Couldn't reproduce)

I threw a bunch of explicit test fixtures at it using **WebStorm 2026.1.3** to see if I could get it to trip up. I tested everything from TS/JS/TSX named imports, local aliases, defaults, and namespaces, to CommonJS `require` variants and local unexported elements. 

In every single case, the resolver behaved and returned `not_found` for symbols that are only imported but not exported by the consumer file, which is exactly what we want.

EremesNG:
I pushed the fixes for the confirmed issues for points 1 and 3.

hechtcarmel:
I was gonna say just solve conflicts and let's merge but found something. Posting ai summary which also likely includes a solution. 
After that let's fix conflics and if there are no critical issues let's merge

# JS/TS file rename still reports false importer failures

The actual source rewrite seems to work, but the MCP response can incorrectly say that imports were not retargeted.

## Repro

I tested this with the installed plugin in:

```text
/Users/carmel.h/git/nexus/taboola-features-catalog
```

Fixture:

```ts
// importOnly.ts
import { loadConfig } from "./config";

export const importedName = loadConfig;
```

```ts
// index.ts
export { loadConfig } from "./config";
export * from "./config";
```

Then I called:

```json
{
  "file": "src/frontend/src/__mcp_pr175_fixture__/config.ts",
  "targetType": "file",
  "newName": "configRenamed.ts"
}
```

The files were updated correctly:

```ts
import { loadConfig } from "./configRenamed";
```

```ts
export { loadConfig } from "./configRenamed";
export * from "./configRenamed";
```

But the tool response still reported this:

```text
success: true
message: Successfully renamed 'config.ts' to 'configRenamed.ts'; 2 importers could not be auto-retargeted
unretargetedImporters:
- src/frontend/src/__mcp_pr175_fixture__/importOnly.ts
- src/frontend/src/__mcp_pr175_fixture__/index.ts
```

Renaming back from `configRenamed.ts` to `config.ts` produced the same kind of false warning.

## Why this happens

The new JS/TS rename code collects import references before running IntelliJ rename and stores the reference range:

```kotlin
val rangeInElement = collectedReference.rangeInElement
```

Then it runs IntelliJ rename:

```kotlin
renameProcessor.run()
```

At this point IntelliJ has already changed the import text. For example:

```ts
"./config"
```

becomes:

```ts
"./configRenamed"
```

That means the old stored reference range is no longer reliable. In my repro the code looked for the old range `(8,14)`, but after the import became longer the current reference range had changed.

The custom post-rename code then tries to find the reference by the old exact range:

```kotlin
return lastFileReference.takeIf { it.rangeInElement == rangeInElement }
```

When that lookup fails, it reports the importer as unretargeted.

So this is not IntelliJ failing to rename the import. IntelliJ already did the right thing. The failure is in the PR's extra post-rename verification/retargeting logic, which treats a stale reference range as proof that retargeting failed.

## Why it matters

Agents will trust the MCP result. If the response says some imports failed, an agent may retry the rename, manually patch imports, or report a broken refactor even though the code is already correct.

So even though the files end up correct in this case, the tool contract is not reliable yet.

## Suggested fix

After `RenameProcessor.run()`, do not immediately fail when the old stored range cannot be found.

Instead, first check the current PSI/document state:

```kotlin
val importerChanged = importerTextBefore != importerTextAfter
val nowResolves = importerFileHasReferenceTo(importerFile, renamedFile)

if (importerChanged && nowResolves) {
    // IntelliJ rename already updated this importer correctly.
    affectedFiles += importerPath
    continue
}
```

Only if the importer still does not resolve to the renamed file should the code try a manual retarget. If the old range is stale, rescan references in the importer file instead of reporting failure immediately.

The old range can still be used as a fast path, but it should not be the source of truth after `RenameProcessor.run()`.

## Regression test to add

I would add a test with double-quoted import/export strings like the runtime repro:

```ts
import { loadConfig } from "./config";
export { loadConfig } from "./config";
export * from "./config";
```

Rename:

```text
config.ts -> configRenamed.ts
```

Expected:

```kotlin
assertTrue(importOnlyText.contains("\"./configRenamed\""))
assertTrue(indexText.contains("\"./configRenamed\""))
assertNull(payload.warnings)
assertNull(payload.unretargetedImporters)
```

This should catch the current false-warning behavior.

EremesNG:
Good catch, thanks. I pushed the fix.

hechtcarmel:
Great.
Ready to merge after resolving merge conflicts 👍

EremesNG:
Due to the limited time, I used AI to resolve conflicts, I think everything should be in order according to the tests performed after the conflict resolution.

hechtcarmel:
Merging. I will send later today a new version to jetbrains, hopefully with some other merges as well
thank you for you contriubtion.

Note that there are still issues, but it's probably better to have it then not.
Here is the issue incase you'd like to give it a shot later, as it's a hard issue:

### P1 - JS/TS call hierarchy can report module importers as callers of a specific export

`JavaScriptCallHierarchyHandler.collectCallerReferences` mixes two different concepts into one output list:

- direct references to the searched function/symbol via `ReferencesSearch.search(symbol, searchScope)`;
- references to the searched symbol's containing file via `ReferencesSearch.search(containingFile, searchScope)`.

The second path is useful for traversing barrels, but it is not necessarily a call to the export. `findCallersRecursive` later treats every collected reference the same way: it takes `reference.reference.element`, finds the enclosing callable, and emits that callable as a caller.

Concrete code path:

- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/handlers/javascript/JavaScriptHandlers.kt:1635` iterates all collected references and reports the enclosing callable.
- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/handlers/javascript/JavaScriptHandlers.kt:1734` adds references to the whole containing file.
- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/handlers/javascript/JavaScriptHandlers.kt:1796` prevents traversal through references already inside callables, but the file reference itself is already retained and later emitted as a caller.

Example failure mode:

```ts
// plugin-config.ts
export function loadPluginConfig() {
  return {};
}

// consumer.ts
export async function preloadConfigModule() {
  await import("./plugin-config");
}
```

As written, `ide_call_hierarchy` for `plugin-config#loadPluginConfig` can return `preloadConfigModule` as a caller even though the function imports the module and never calls `loadPluginConfig`.

This can break agents that use call hierarchy for impact analysis. They will see false dependencies and may edit or test unrelated code.

PR #181, opened 2026-05-17
Add PHP symbol reference handler

### LINKED ISSUE #179: [Feature] Add PHP symbol reference support for language+symbol navigation tools
Implement PHP support for language + symbol lookup in symbol-reference-capable MCP tools.

Currently, tools such as `ide_find_definition`, `ide_find_references`, `ide_call_hierarchy`, `ide_find_implementations`, and `ide_find_super_methods` support symbol-based lookup through SymbolReferenceHandler, but PHP symbols were not supportes, only Java has search by Symbol references.

ineersa:
Closes #179

## Summary
- add `PhpSymbolReferenceHandler` for PHP `language` + `symbol` lookup
- support class/interface/trait FQNs plus methods, properties, and constants
- update symbol-reference tool descriptions and skill reference docs
- add unit coverage for accepted/rejected PHP symbol formats

## Validation
- `./gradlew test --tests "*UnitTest*"`
- manually tested in PhpStorm via MCP HTTP endpoint for PHP class/interface/trait/method/property definitions, references, super methods, call hierarchy, and implementations

Note: per project instructions, platform tests were not run locally.

## AI Disclosure
Written by DeepSeek V4 Flash, supervised by GPT‑5.5, and accepted by the human who had the GitHub token.

### DIFF
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/navigation/CallHierarchyTool.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/navigation/CallHierarchyTool.kt
@@ -44,12 +44,13 @@ class CallHierarchyTool : AbstractMcpTool() {
 
         Target (mutually exclusive):
         - file + line + column: position-based lookup
-        - language + symbol: fully qualified symbol reference (currently supported for Java only)
+        - language + symbol: fully qualified symbol reference (supported languages: ${supportedSymbolReferenceLanguagesDescription()})
 
         Parameters: direction (required): "callers" or "callees". depth (optional, default: 3, max: 5). scope (optional, default: "project_files"; supported: project_files, project_and_libraries, project_production_files, project_test_files).
 
         Example: {"file": "src/Service.java", "line": 42, "column": 10, "direction": "callers"}
         Example: {"language": "Java", "symbol": "com.example.Service#processRequest(String)", "direction": "callers", "scope": "project_and_libraries"}
+        Example: {"language": "PHP", "symbol": "\\App\\Service\\UserService::find()", "direction": "callers"}
     """.trimIndent()
 
     override val inputSchema: JsonObject = SchemaBuilder.tool()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ## [Unreleased]
 ### Added
 - **PHP symbol reference handler** — PHP now supports `language`+`symbol` parameter mode for `ide_find_references`, `ide_find_definition`, `ide_call_hierarchy`, `ide_find_implementations`, and `ide_find_super_methods`. Accepts symbol formats with PHP namespaces (e.g., `\\App\\Service\\UserService`, `\\App\\Service\\UserService::find()`, `\\App\\Service\\UserService::\$property`). Fixes [#179](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/issues/179).
+- **PHP symbol reference member lookup** — Inherited and case-insensitive PHP methods resolve through PhpStorm's `findMethodByName(CharSequence)` API, field/constant lookup uses the matching `findFieldByName(CharSequence, boolean)` signature, and plain `Class::name` symbols do not fall back to properties without the documented `$property` syntax.
 
 ## [4.17.2] - 2026-05-18
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,9 @@
 
 ## [Unreleased]
 ### Added
-- **PHP symbol reference handler** — PHP now supports `language`+`symbol` parameter mode for `ide_find_references`, `ide_find_definition`, `ide_call_hierarchy`, `ide_find_implementations`, and `ide_find_super_methods`. Accepts symbol formats with PHP namespaces (e.g., `\\App\\Service\\UserService`, `\\App\\Service\\UserService::find()`, `\\App\\Service\\UserService::\$property`). Fixes [#179](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/issues/179).
+- **PHP symbol reference handler** — PHP now supports `language`+`symbol` parameter mode for `ide_find_references`, `ide_find_definition`, `ide_call_hierarchy`, `ide_find_implementations`, and `ide_find_super_methods`. Accepts symbol formats with PHP namespaces (e.g., `\\App\\Service\\UserService`, `\\App\\Service\\UserService::find()`, `\\App\\Service\\UserService::$property`). Fixes [#179](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/issues/179).
 - **PHP symbol reference member lookup** — Inherited and case-insensitive PHP methods resolve through PhpStorm's `findMethodByName(CharSequence)` API, field/constant lookup uses the matching `findFieldByName(CharSequence, boolean)` signature, and plain `Class::name` symbols do not fall back to properties without the documented `$property` syntax.
+- **PHP enum case resolution** — `EnumType::CASE` now resolves to enum case PSI elements via PhpStorm's `getEnumCases()` API. Enum case lookup runs before class constant lookup so `::CASE` correctly targets enum cases on enum types. `::CASE()` still resolves as a method call.
 
 ## [4.17.2] - 2026-05-18
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 # IDE Index MCP Server Changelog
 
 ## [Unreleased]
+### Added
+- **PHP symbol reference handler** — PHP now supports `language`+`symbol` parameter mode for `ide_find_references`, `ide_find_definition`, `ide_call_hierarchy`, `ide_find_implementations`, and `ide_find_super_methods`. Accepts symbol formats with PHP namespaces (e.g., `\\App\\Service\\UserService`, `\\App\\Service\\UserService::find()`, `\\App\\Service\\UserService::\$property`). Fixes [#179](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/issues/179).
 
 ## [4.17.2] - 2026-05-18
 ### Fixed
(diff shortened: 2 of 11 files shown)


src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/navigation/CallHierarchyTool.kt
```diff
@@ -44,12 +44,13 @@ class CallHierarchyTool : AbstractMcpTool() {
 
         Target (mutually exclusive):
         - file + line + column: position-based lookup
-        - language + symbol: fully qualified symbol reference (currently supported for Java only)
+        - language + symbol: fully qualified symbol reference (supported languages: ${supportedSymbolReferenceLanguagesDescription()})
```
ineersa, line 47: Added dynamic languages list in tool descriptions, seems to be working.

gradle.properties
```diff
@@ -4,7 +4,7 @@ pluginGroup = com.github.hechtcarmel.jetbrainsindexmcpplugin
 pluginName = IDE Index MCP Server
 pluginRepositoryUrl = https://github.com/hechtcarmel/jetbrains-index-mcp-plugin
 # SemVer format -> https://semver.org
-pluginVersion = 4.16.3
+pluginVersion = 4.17.0
```
ineersa: AI decided to bump version. I guess it was in AGENTS instructions.

src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/McpConstants.kt
```diff
@@ -50,7 +50,7 @@ object McpConstants {
      * Legacy constant for backwards compatibility.
      */
     const val SERVER_NAME = "jetbrains-index-mcp"
-    const val SERVER_VERSION = "4.10.4"
+    const val SERVER_VERSION = "4.17.0"
```
ineersa: This change is 100% not needed, AI during testing decided to change it as plugin version and server version didn't match, and it was suspicious for it.

hechtcarmel:
Thanks for the contribution!
About the versions thing - i'll remove it from AGENTS.md. Just update to latest version to resolve the conflict and i'll add a new version update PR.

I'm not a php guy, but AI review flagged
```

findConstantInClass calls reflective getOwnConstants / getConstants, but PhpStorm's open API exposes PhpClass.findFieldByName(name, findConstant), getFields, getOwnFields, and Field.isConstant(), not those constant methods. Then resolvePhpMember falls back to generic field lookup, so Class ::$FOO can resolve a constant, and `Class

::FOO
can resolve a property, depending on whatgetFields()` returns.

This should filter Field.isConstant() or call findFieldByName(name, true/false) reflectively.
```

ineersa:
Addressed review feedback in `21eaf97`:

- updated branch to latest `main` (`4.17.2`) and removed this PR's version changes (`gradle.properties` now matches main; `McpConstants.SERVER_VERSION` reverted)
- replaced reflective `getOwnConstants` / `getConstants` usage with `findFieldByName(name, findConstant)` plus `Field.isConstant()`-filtered `getOwnFields`/`getFields` fallback
- reran `./gradlew test --tests "*UnitTest*"` successfully

This should prevent `Class::$FOO` resolving constants as properties and `Class::FOO` resolving properties as constants.

P.S. Squashed commits.

hechtcarmel:
I tested this in PhpStorm via the MCP endpoint with a small fixture under `src/McpSymbolReferenceReview`, and I think this still has a real symbol-mode resolution bug.

The good part: class lookup, direct own methods, constants, and documented `$property` symbols work in `language + symbol` mode. For example this resolves correctly:

```json
{"language": "PHP", "symbol": "\\App\\McpSymbolReferenceReview\\ChildService::ownMethod()"}
```

It returns `ChildService.php:11`.

These also resolve correctly:

```json
{"language": "PHP", "symbol": "\\App\\McpSymbolReferenceReview\\ChildService::$childProperty"}
{"language": "PHP", "symbol": "\\App\\McpSymbolReferenceReview\\ChildService::CHILD_CONSTANT"}
{"language": "PHP", "symbol": "\\App\\McpSymbolReferenceReview\\ChildService::$repository"}
```

But inherited methods fail in `language + symbol` mode. I tested:

```json
{"language": "PHP", "symbol": "\\App\\McpSymbolReferenceReview\\ChildService::sharedMethod()"}
```

where `ChildService extends BaseService` and `sharedMethod()` is declared on `BaseService`.

Actual result:

```text
No member 'sharedMethod()' found in type '\App\McpSymbolReferenceReview\ChildService'
```

This is not a PhpStorm indexing problem. Position-based lookup on the same call site resolves correctly to `BaseService.php:11`, and position-based find references finds both usages:

```json
{"file": "src/McpSymbolReferenceReview/Usage.php", "line": 11, "column": 27}
```

I looked into the PHP SDK API more closely, and the root cause appears to be the reflective method signature. The code looks up:

```kotlin
phpClass.javaClass.getMethod("findMethodByName", String::class.java)
```

but PhpStorm exposes:

```java
findMethodByName(CharSequence)
```

Reflection requires an exact parameter type match, so the `String` lookup misses the real SDK method and falls back to scanning only `getOwnMethods()`. That explains why direct methods work, but inherited methods do not.

I also reproduced the case-sensitivity issue:

```json
{"language": "PHP", "symbol": "\\App\\McpSymbolReferenceReview\\BaseService::CASEMETHOD()"}
```

Actual result:

```text
No member 'CASEMETHOD()' found in type '\App\McpSymbolReferenceReview\BaseService'
```

while exact-case `caseMethod()` resolves. PHP method names are case-insensitive, so this is another symptom of bypassing the PhpStorm `findMethodByName(...)` API and doing a plain `getName() == methodName` fallback.

The field path has the same reflection signature issue:

```kotlin
findFieldByName(String, boolean)
```

should also use `CharSequence`, although in practice properties/constants are mostly masked by the `getFields()` fallback.

One more correctness issue: this also resolves even though the documented property syntax requires `$`:

```json
{"language": "PHP", "symbol": "\\App\\McpSymbolReferenceReview\\ChildService::childProperty"}
```

It resolves to `$childProperty`. I think plain `Class::name` should only resolve constants/methods, otherwise a missing constant can silently navigate to a property.

Suggested fix: use the exact PHP SDK reflective signatures, especially:

```kotlin
getMethod("findMethodByName", CharSequence::class.java)
getMethod("findFieldByName", CharSequence::class.java, Boolean::class.javaPrimitiveType)
```

and avoid falling back from no-`$` member syntax to property lookup unless that is intentionally supported and documented.

ineersa:
Updated PR, hope it will address your comment.

Honestly I have no idea what's going on there, tested on few projects, symbol reference seems to be working.

Thanks for provided feedback, that was very helpful.

hechtcarmel:
Hi, 
It fixes the issue in the comment, unfortunately while doing final checking i found it broke 2 things.

I asked AI to explain them  clearly with examples:


## 1. `ide_find_implementations` can report inherited methods as implementations

`findMethodInClass()` now uses PhpStorm's `findMethodByName(CharSequence)` first. That API can return methods inherited from parent classes. This is useful when resolving a direct symbol like `\App\ChildRepo::save()`, but it is not safe for implementation search.

`PhpImplementationsHandler.findMethodImplementations()` iterates over subclasses and calls `findMethodInClass(subclass, methodName)`. With the new behavior, a subclass that merely inherits the method can be reported as if it implemented or overrode the method itself.

Example:

```php
abstract class BaseRepo
{
    public function save(): void
    {
    }
}

class ChildRepo extends BaseRepo
{
    // No save() override here.
}
```

When asking for implementations of `BaseRepo::save()`, the current PR can inspect `ChildRepo`, resolve the inherited `BaseRepo::save()` method, and emit a result named `ChildRepo::save` while pointing to the file and line of `BaseRepo::save()`.

That result is misleading: `ChildRepo` does not declare an implementation. The tool should only report a subclass method when that method is actually declared on the subclass, or otherwise intentionally model inherited implementations with accurate metadata.

Relevant code:

- `findMethodInClass()` now resolves inherited methods: `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/handlers/php/PhpHandlers.kt`
- `findMethodImplementations()` uses that helper while iterating subclasses.

## 2. `ide_find_super_methods` can label an ancestor method as belonging to the wrong class

The same inherited lookup also affects super-method hierarchy construction. When walking from a method to its parent classes, the code asks each immediate superclass for a method with the same name. If the immediate superclass does not declare the method, `findMethodByName()` can still return a method from a higher ancestor.

The result is then labeled with the immediate superclass name, even though the returned PSI element belongs to a different class.

Example:

```php
interface Handler
{
    public function handle(): void;
}

class BaseHandler implements Handler
{
    public function handle(): void
    {
    }
}

class MiddleHandler extends BaseHandler
{
    // No handle() override here.
}

class LeafHandler extends MiddleHandler
{
    public function handle(): void
    {
    }
}
```

When asking for super methods of `LeafHandler::handle()`, the traversal checks `MiddleHandler`. Because `MiddleHandler` inherits `handle()` from `BaseHandler`, the new helper can return the `BaseHandler::handle()` PSI element. The result can then be rendered as if `MiddleHandler::handle()` exists, while the file and line point to `BaseHandler::handle()`.

That mixes two different facts:

- the containing class metadata says `MiddleHandler`
- the actual method declaration belongs to `BaseHandler`

The tool should either skip `MiddleHandler` when it does not declare the method, or report the actual containing class from the returned method element.

Relevant code:

- `findMethodInClass()` now resolves inherited methods.
- `PhpSuperMethodsHandler.buildHierarchy()` labels the result with the superclass currently being inspected.



Also, very minor - 
 CHANGELOG.md (line 7) documents PHP properties as ::\$property. In Markdown code spans that renders as a literal backslash before $, while the supported symbol format is ::$property as correctly documented in tools-reference.md (line 14).

ineersa:
Great, thanks for review!

Added changes to address those comments.

Also while testing found issues with Enum cases resolution, so added a fix for that too.

hechtcarmel:
Fantastic.
I'm merging now and will shortly send it as part of v4.18.0 for jetbrains review.
From there, it taks anywhere between 10 minutes to 5 days till they approve and it's available on the marketplace

ineersa:
Thanks to AI =)

In Pi extension I wrapped it up to support symbols, I guess I need to rewrite extension now =)

Still getting some troubles forcing LLM to use IDE tools, but wrapped it up as native tools in Pi and it works slightly better.

I bet Jetbrains take ideas from this Plugin for their MCP server!

PR #183, opened 2026-05-18
Add outputSchema to tools list

chikei:
### Description

The property is described at https://modelcontextprotocol.io/specification/2025-06-18/server/tools#output-schema

This is added mainly to work with other mcp tooling such as cloudflare code mode or alternatives such as [pctx](https://github.com/portofcontext/pctx)

Also add an option to toggle this (and default to turned off) to preserve current behavior.

### AI Disclosure

The output schemas are created by Codex and proofread by me.

### DIFF
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/JsonRpcHandler.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/JsonRpcHandler.kt
@@ -8,6 +8,7 @@ import com.github.hechtcarmel.jetbrainsindexmcpplugin.history.CommandEntry
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.history.CommandHistoryService
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.history.CommandStatus
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.server.models.*
+import com.github.hechtcarmel.jetbrainsindexmcpplugin.settings.McpSettings
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.ToolRegistry
 import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.project.Project
@@ -99,10 +100,24 @@ class JsonRpcHandler @JvmOverloads constructor(
     private fun processToolsList(request: JsonRpcRequest): JsonRpcResponse {
         val tools = toolRegistry.getToolDefinitions()
         val result = ToolsListResult(tools = tools)
+        val encodedResult = json.encodeToJsonElement(result)
+        val cleanedResult = if(!McpSettings.getInstance().toolsOutputOutputSchema){
+            JsonObject(
+                encodedResult.jsonObject.mapValues { (_, value) ->
+                    when (value) {
+                        is JsonArray -> JsonArray(value.map { item ->
+                            if (item is JsonObject) JsonObject(item.filterKeys { it != "outputSchema" })
+                            else item
+                        })
+                        else -> value
+                    }
+                }
+            )
+        } else encodedResult
 
         return JsonRpcResponse(
             id = request.id,
-            result = json.encodeToJsonElement(result)
+            result = cleanedResult
         )
     }
 
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/models/McpModels.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/models/McpModels.kt
@@ -9,7 +9,8 @@ import kotlinx.serialization.json.JsonObject
 data class ToolDefinition(
     val name: String,
     val description: String,
-    val inputSchema: JsonObject
+    val inputSchema: JsonObject,
+    val outputSchema: JsonObject? = null
 )
 
 @Serializable
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/settings/McpSettings.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/settings/McpSettings.kt
@@ -36,7 +36,8 @@ class McpSettings : PersistentStateComponent<McpSettings.State> {
         var responseFormat: ResponseFormat = ResponseFormat.JSON,
         var disabledTools: MutableSet<String> = mutableSetOf("ide_build_project", "ide_file_structure", "ide_find_symbol", "ide_read_file", "ide_get_active_file", "ide_open_file", "ide_reformat_code", "ide_optimize_imports", "ide_convert_java_to_kotlin"),
         var serverPort: Int = -1, // -1 means use IDE-specific default
-        var serverHost: String = McpConstants.DEFAULT_SERVER_HOST
+        var serverHost: String = McpConstants.DEFAULT_SERVER_HOST,
+        var toolsOutputOutputSchema: Boolean = false
     )
 
     private var state = State()
@@ -62,6 +63,10 @@ class McpSettings : PersistentStateComponent<McpSettings.State> {
     var responseFormat: ResponseFormat
         get() = state.responseFormat
         set(value) { state.responseFormat = value }
+
+    var toolsOutputOutputSchema: Boolean
+        get() = state.toolsOutputOutputSchema
+        set(value) { state.toolsOutputOutputSchema = value }
 
     var disabledTools: Set<String>
         get() = state.disabledTools.toSet()
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/settings/McpSettingsConfigurable.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/settings/McpSettingsConfigurable.kt
@@ -33,7 +33,6 @@ import java.awt.event.FocusEvent
 import java.net.InetAddress
 import java.net.InetSocketAddress
 import java.net.ServerSocket
-import java.util.function.Supplier
 import javax.swing.BoxLayout
 import javax.swing.JComponent
 import javax.swing.JPanel
@@ -50,6 +49,7 @@ class McpSettingsConfigurable : Configurable {
     private var syncExternalChangesCheckBox: JBCheckBox? = null
     private var availableProjectsModeComboBox: ComboBox<McpSettings.AvailableProjectsMode>? = null
     private var responseFormatComboBox: ComboBox<McpSettings.ResponseFormat>? = null
+    private var toolsOutputOutputSchemaCheckBox: JBCheckBox? = null
     private val toolCheckBoxes = mutableMapOf<String, JBCheckBox>()
     private var uiDisposable: Disposable? = null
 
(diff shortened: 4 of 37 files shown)


hechtcarmel:
Hi, 
will look into it during the week.
Thanks!

hechtcarmel:
Asked codex to post an initial first-pass review

hechtcarmel:
# PR #183 Output Schema Review Notes

Branch reviewed: `output`

PR: <https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/183>

Commit reviewed: `56de00a2dd95a9946c4d2f03fb05e6f172390b8c`

This note expands three review findings:

1. `outputSchema` is advertised without MCP `structuredContent`.
2. The server still initializes as MCP `2025-03-26` while exposing a `2025-06-18` tool field.

## TL;DR

The PR adds `outputSchema` to tool definitions, but it does not add `structuredContent` to tool call results. In MCP `2025-06-18`, `outputSchema` is not just documentation for a text blob. It describes the JSON object returned in the result's `structuredContent` field. When a tool advertises an output schema, successful structured results must conform to that schema.

The server also still identifies itself as MCP `2025-03-26`, a version whose tool definition does not include `outputSchema`. If the setting is enabled, clients connected to a server that negotiated or expects `2025-03-26` can receive a `2025-06-18` field anyway.


## For Dummies

Think of `outputSchema` as a promise.

The server says: "When you call this tool, I will give you a JSON object shaped like this."

But this PR currently says that in the tool list, then still returns the real tool result as text only:

```json
{
  "content": [
    {
      "type": "text",
      "text": "{\"file\":\"src/Foo.kt\",\"language\":\"kotlin\",\"structure\":\"...\"}"
    }
  ],
  "isError": false
}
```

For the MCP output schema contract, the important object should also appear as `structuredContent`:

```json
{
  "content": [
    {
      "type": "text",
      "text": "{\"file\":\"src/Foo.kt\",\"language\":\"kotlin\",\"structure\":\"...\"}"
    }
  ],
  "structuredContent": {
    "file": "src/Foo.kt",
    "language": "kotlin",
    "structure": "..."
  },
  "isError": false
}
```

The text block is still useful for older clients. The structured object is what clients validate against `outputSchema`.

The protocol version issue is similar. The server says, "I speak the March 2025 MCP language," but the PR optionally uses a word that was added in the June 2025 language. Some clients will ignore the extra word. Stricter clients may reject it, fail validation, or behave inconsistently.


## Sources

Official MCP `2025-06-18` Tools specification:

- Tool definitions include optional `outputSchema`: <https://modelcontextprotocol.io/specification/2025-06-18/server/tools>
- Structured content is returned in `structuredContent`, and a serialized text copy is recommended for backward compatibility: <https://modelcontextprotocol.io/specification/2025-06-18/server/tools#structured-content>
- If an `outputSchema` is provided, servers must provide structured results that conform to it: <https://modelcontextprotocol.io/specification/2025-06-18/server/tools#output-schema>

Official MCP `2025-03-26` Tools specification:

- Tool definitions in this version include `name`, `description`, `inputSchema`, and optional `annotations`; they do not include `outputSchema`: <https://modelcontextprotocol.io/specification/2025-03-26/server/tools>

Official MCP `2025-06-18` schema reference:

- `Tool.outputSchema` describes the structure returned in `CallToolResult.structuredContent`: <https://modelcontextprotocol.io/specification/2025-06-18/schema#tool>

## Finding 1: `outputSchema` Is Advertised Without `structuredContent`

Severity: P1

### What The PR Adds

The PR adds `outputSchema` to `ToolDefinition`:

```kotlin
@Serializable
data class ToolDefinition(
    val name: String,
    val description: String,
    val inputSchema: JsonObject,
    val outputSchema: JsonObject? = null
)
```

Reference:

- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/models/McpModels.kt:8`

It then wires that field through `ToolRegistry` and conditionally exposes it from `tools/list`:

```kotlin
val cleanedResult = if(!McpSettings.getInstance().toolsOutputOutputSchema) {
    ...
} else encodedResult
```

Reference:

- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/JsonRpcHandler.kt:100`

So when the new setting is enabled, clients can see `outputSchema` on every tool.

### What The PR Does Not Add

`ToolCallResult` still has only:

```kotlin
@Serializable
data class ToolCallResult(
    val content: List<ContentBlock>,
    val isError: Boolean = false
)
```

Reference:

- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/models/McpModels.kt:16`

There is no `structuredContent` field.

The shared result helper still turns structured Kotlin data into text only:

```kotlin
protected inline fun <reified T> createJsonResult(data: T): ToolCallResult {
    return try {
        val jsonText = json.encodeToString(data)
        ToolCallResult(
            content = listOf(ContentBlock.Text(text = formatStructuredPayload(jsonText))),
            isError = false
        )
    } catch (e: Exception) {
        createErrorResult(formattingFailureMessage(e))
    }
}
```

Reference:

- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/AbstractMcpTool.kt:745`

That means even tools whose Kotlin result objects are structured are returned only as `ContentBlock.Text`.

### Why That Violates The MCP Contract

In MCP `2025-06-18`, `outputSchema` describes the JSON object returned in `structuredContent`. It is not merely a hint that the text block happens to contain JSON.

The spec's expected pattern is:

```json
{
  "result": {
    "content": [
      {
        "type": "text",
        "text": "{\"temperature\":22.5,\"conditions\":\"Partly cloudy\"}"
      }
    ],
    "structuredContent": {
      "temperature": 22.5,
      "conditions": "Partly cloudy"
    }
  }
}
```

The text copy is for backward compatibility. The structured object is the contract target.

The PR currently produces only the text copy. Once `outputSchema` is enabled, strict clients are allowed to validate the result and conclude that the required `structuredContent` object is missing.

### Why This Matters In Real Clients

Clients or tooling that care about `outputSchema` often do so because they want machine-readable results without parsing arbitrary text. If they see an `outputSchema`, they may reasonably expect:

- `result.structuredContent` exists.
- `result.structuredContent` is a JSON object.
- The object validates against the advertised schema.
- The text block is optional compatibility material, not the only structured result.

This PR targets compatibility with MCP tooling such as Cloudflare code mode and `pctx`. Those are exactly the kinds of clients likely to care about schema-bound structured output. Advertising `outputSchema` without `structuredContent` can make the server look more compatible while still failing the behavior those clients need.

There is an additional interaction with the existing `responseFormat` setting. `createJsonResult` passes the serialized JSON through `formatStructuredPayload`, which can return TOON instead of JSON text. That may be fine for the text block by local design, but it makes `structuredContent` even more important. The schema-bound result should remain a JSON object regardless of the human/model-facing text format.

### Better Fix Shape

The robust fix is to add real structured output support:

1. Add `structuredContent: JsonObject? = null` to `ToolCallResult`.
2. Change `createJsonResult` so it serializes the data to a `JsonElement`, stores the object in `structuredContent`, and still emits the text block for backward compatibility.
3. Only attach `structuredContent` for successful structured results. Tool execution errors can remain `isError = true` text or structured error objects if the project wants that separately.
4. Make sure `responseFormat` affects only the text compatibility block, not `structuredContent`.
5. Add tests that call a representative tool and assert that when `outputSchema` is enabled, `tools/call` returns conforming `structuredContent`.

Pseudo-shape:

```kotlin
@Serializable
data class ToolCallResult(
    val content: List<ContentBlock>,
    val structuredContent: JsonObject? = null,
    val isError: Boolean = false
)
```

```kotlin
protected inline fun <reified T> createJsonResult(data: T): ToolCallResult {
    val element = json.encodeToJsonElement(data)
    val jsonText = json.encodeToString(JsonElement.serializer(), element)

    return ToolCallResult(
        content = listOf(ContentBlock.Text(text = formatStructuredPayload(jsonText))),
        structuredContent = element as? JsonObject,
        isError = false
    )
}
```

If the project does not want to add `structuredContent` yet, the safer alternative is to not advertise `outputSchema`.

## Finding 2: The Server Still Initializes As MCP `2025-03-26`

Severity: P2

### What The PR Claims To Add

The PR description links to the MCP `2025-06-18` `outputSchema` feature.

That is the correct spec version for `outputSchema` and structured tool results.

### What The Server Still Advertises

The server's protocol constants remain:

```kotlin
const val LEGACY_MCP_PROTOCOL_VERSION = "2024-11-05"
const val STREAMABLE_HTTP_MCP_PROTOCOL_VERSION = "2025-03-26"
const val MCP_PROTOCOL_VERSION = STREAMABLE_HTTP_MCP_PROTOCOL_VERSION
```

Reference:

- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/McpConstants.kt:37`

The initialize result model also still defaults to:

```kotlin
@Serializable
data class InitializeResult(
    val protocolVersion: String = "2025-03-26",
    ...
)
```

Reference:

- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/models/McpModels.kt:56`

So a Streamable HTTP client sees the server as MCP `2025-03-26`.

### Why That Is A Problem

In MCP `2025-03-26`, the Tool shape does not include `outputSchema`. The tool list response shape documented for that version includes fields such as:

- `name`
- `description`
- `inputSchema`
- optional `annotations`

In MCP `2025-06-18`, the Tool shape includes:

- `name`
- optional `title`
- `description`
- `inputSchema`
- optional `outputSchema`
- optional `annotations`

That means `outputSchema` is a feature of a newer protocol version than the one this server currently declares.

### Why Extra Fields Are Still Risky

Some JSON clients ignore unknown fields. Those clients may continue working.

But compatibility is not just "will a lenient JSON parser ignore this?" The server is presenting a protocol contract. If a client negotiated or expects `2025-03-26`, it may:

- validate `tools/list` responses against the `2025-03-26` schema,
- ignore `outputSchema` and therefore not get the intended benefit,
- warn that the server returned unexpected fields,
- treat `outputSchema` as unsupported because the negotiated protocol version predates it,
- behave differently between strict and lenient clients.

The project also has legacy SSE support for `2024-11-05`. `processToolsList` currently has no protocol-version parameter, so the output-schema setting is global rather than version-aware. If the setting is enabled, the same `tools/list` path can expose `outputSchema` even where the transport/protocol path is legacy.

### Better Fix Shape

There are two viable approaches.

Approach A: upgrade protocol support to `2025-06-18`.

- Change the advertised Streamable HTTP MCP protocol version to `2025-06-18`.
- Update initialize tests and documentation.
- Implement `structuredContent` as described in Finding 1.
- Confirm existing clients still work, or keep fallback paths for older clients.

Approach B: keep `2025-03-26`, but gate `outputSchema`.

- Make `processToolsList` aware of the negotiated/effective protocol version.
- Only include `outputSchema` for clients using `2025-06-18` or newer.
- Keep it absent for `2025-03-26` and `2024-11-05`.
- If there is no full `2025-06-18` support yet, keep the setting disabled or remove it until structured results are implemented.

The current middle ground is the risky part: the server declares an older protocol, then optionally emits a newer field without the corresponding newer result semantics.

## Recommended Review Comment Summary

Use this if leaving a concise PR review comment:

```markdown
The output schema support needs one more protocol layer before this is safe to expose. In MCP 2025-06-18, `outputSchema` describes `result.structuredContent`, but this PR only adds schemas to `tools/list`; `ToolCallResult` still has no `structuredContent` and `createJsonResult` still serializes results into text content only. Also, the server still initializes as MCP `2025-03-26`, where `outputSchema` is not part of the Tool shape. Either gate this field by protocol version and keep it off for 2025-03-26/legacy clients, or upgrade to 2025-06-18 and return conforming `structuredContent` for successful structured tool calls.
```

## Suggested Tests To Add

1. `tools/list` with output schemas disabled:
   - Assert no tool object contains `outputSchema`.

2. `tools/list` with output schemas enabled and protocol `2025-06-18`:
   - Assert representative tools include `outputSchema`.

3. `tools/list` for legacy or `2025-03-26` protocol:
   - Assert `outputSchema` is omitted, unless the server intentionally upgrades that protocol version.

4. `tools/call` for a representative structured tool:
   - Assert `structuredContent` exists.
   - Assert `content[0].text` still exists for backward compatibility.
   - Assert `structuredContent` validates against the advertised output schema.

6. `responseFormat = TOON`:
   - Assert text content can be TOON if desired.
   - Assert `structuredContent` remains JSON object data and is not TOON-encoded text.

PR #187, opened 2026-05-24
Add ide_set_power_save_mode, ide_close_project, ide_open_project

mdproctor:
## Summary

Three stateless tools for managing open project windows. Each is an independent RPC call with no side effects beyond the explicit operation — no background processes, no persistent state.

**`ide_set_power_save_mode`** — toggles IDE Power Save Mode. Suspends background inspections and code analysis while leaving the index and all code intelligence operations (find usages, refactoring, navigation) fully functional. Useful when a project is open but not being actively edited.

**`ide_close_project`** — closes an open project window and frees its memory. Non-blocking; close is scheduled on the EDT. The project can be reopened via Recent Projects or `ide_open_project`.

**`ide_open_project`** — opens a project by filesystem path and blocks until indexing completes, so subsequent MCP tool calls on the opened project succeed immediately. Requires at least one project already open to serve as the JSON-RPC context.

All three tools are **disabled by default** and must be enabled in Settings → Tools → Index MCP Server before use.

## Motivation

AI coding agents working across multiple open projects benefit from being able to manage IDE resource usage without leaving IntelliJ. These tools expose existing IDE capabilities that are currently only accessible via the UI. They are also the building blocks for the lifecycle management layer in \#188.

## Implementation notes

**`OpenProjectTask` reflection** — The `OpenProjectTask` DSL builder and constructor are marked `@Internal`. Rather than risk binary incompatibility across IntelliJ versions, `forceOpenInNewFrame` and `showWelcomeScreen` are set via reflection with a `runCatching` fallback, so the tool degrades gracefully if the field is ever removed.

**NON_MODAL for `DumbService.runWhenSmart`** — Calling `runWhenSmart` from a pooled coroutine thread triggers an `Unexpected modality: should not be ANY` warning in IntelliJ 2026+. The call is now dispatched via `ApplicationManager.invokeLater(..., ModalityState.NON_MODAL)` before scheduling the smart-mode callback.

**Path validation before `openProjectAsync`** — `ProjectManagerEx.openProjectAsync` hangs indefinitely in headless/CI environments when the target path does not exist. The path is validated up front so error responses return immediately.

## Test plan

- `ProjectWindowToolsUnitTest` — schema contracts, ToolNames.ALL completeness, alphabetical sort, disabled-by-default check
- `ProjectWindowToolsTest` — Power Save Mode round-trip; close and open parameter validation; open error on non-existent path


hechtcarmel:
Hi,
About all the PRs you submitted - Cool ideas.

1. Let's make those not enabled by default - most users don't need them. they should be opt-in
2. Not looking at the code as they are drafts, so undraft and leave a message please when ready to take a look

hechtcarmel:
please fix build issue

mdproctor:
> please fix build issue

I'm testing locally, and keeping an eye on PRs. but this one doesn't auto build, any chance you can turn that on? so I can be sure everything remains green.

hechtcarmel:
not sure why, rerunning

mdproctor:
I've been running https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/187 https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/188 and https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/189 now for a while, all seems stable. Would be great to ensure CI passes for all three PRs, and see what we need to do next to get all three merged.

hechtcarmel:
last few things need fixing before merge:

1. Remove all the CHANGELOG.md changes from this PR . I will add it later
2. Revert .idea/gradle.xml - gradleJvm now points to homebrew-21, which is your local JDK name.

3. USAGE.md has a leftover section from #188, right after ide_open_project - a heading-less block documenting ide_lifecycle_log, which doesn't exist in this PR (its example request is also missing the name field). Please remove it.

4. Same leftover in McpBundle.properties - lifecycle.logToFile.label is unused in this PR, belongs to #188.

5. The following uses an intellij internal api - and i know from experience that jetbrains will then not approve it for the marketplace. Claude exaplanation:

```
About the OpenProjectTask reflection — this one's a must-fix, because the PR currently does use internal API despite the comment:

OpenProjectTask() is the @ApiStatus.Internal primary constructor (a no-arg call is just the primary constructor with all defaults), so the reflection doesn't avoid it — it only avoids the @Internal DSL function, and then mutates private final fields on top. That's the fragile part: if JetBrains renames either field, the runCatching swallows it silently and forceOpenInNewFrame stays false. With another project open that means the "Open project in new window?" modal dialog blocking the tool call until a human clicks it — or, if the user has "same window" configured, closing their current project.

There's a public, stable way to get the same thing — build() and withForceOpenInNewFrame() are regular public API:

private fun openTask(): OpenProjectTask =
    OpenProjectTask.build().withForceOpenInNewFrame(true)
showWelcomeScreen can just be dropped — it only matters when no frame would be left, and this tool already requires an open project as a precondition. While you're in the file, ModalityState.NON_MODAL is deprecated — use ModalityState.nonModal().
```

Also, ran a review using the new Fable Claude Code model and it caught the following:

6. SetPowerSaveModeTool: PowerSaveMode.setEnabled() needs to run on the EDT. It fires its topic listeners synchronously on the calling thread, and our tools execute on a Ktor worker thread - listeners like DaemonListeners clear editor markup and restart the daemon, which is EDT work. Wrap it in edtAction { } from AbstractMcpTool, same as the other tools. (The platform test doesn't catch this because tests run on EDT.)

7.

mdproctor:
All 6 points addressed — thank you for the detailed review:

1. **CHANGELOG removed** — stripped our `[Unreleased]` block entirely; you'll add it at merge time.
2. **`.idea/gradle.xml` reverted** — back to the upstream version (no local JDK reference).
3. **USAGE.md leftover removed** — deleted the heading-less `ide_lifecycle_log` block that bled in from #188 after `ide_open_project`.
4. **McpBundle.properties cleaned** — removed `lifecycle.logToFile.label`; it belongs to #188.
5. **`openTask()` fixed** — replaced the reflection-based approach with `OpenProjectTask.build().withForceOpenInNewFrame(true)` (public stable API). Also replaced the deprecated `ModalityState.NON_MODAL` with `ModalityState.nonModal()`.
6. **`SetPowerSaveModeTool` EDT fix** — wrapped `PowerSaveMode.setEnabled()` in `edtAction { }` so it runs on the EDT, matching the pattern used by other tools.

PR is rebased onto v4.21.1 (which includes the merged #189). Unit tests green. Ready for re-review.

hechtcarmel:
Merged and submitted as part of v4.22.0

PR #188, opened 2026-05-24
Add project lifecycle management (ACTIVE/BACKGROUND/DORMANT/CLOSED)

mdproctor:
## Context

When running many IntelliJ projects simultaneously as MCP servers for AI coding agents, idle projects consume memory indefinitely — eventually causing IDE freezes. This PR adds a lifecycle layer that automatically sleeps and wakes projects based on window focus and MCP activity, freeing memory when projects aren't in use.

## Depends on

\#187 (ide_set_power_save_mode, ide_close_project, ide_open_project) — this PR should be reviewed after that one.

## State machine

Four states per managed project, with all timing thresholds configurable in Settings → Index MCP Server:

| State | What runs | Trigger |
|-------|-----------|---------|
| `active` | Full IDE | Window focus gained |
| `background` | Power Save ON, MCP functional | 2 min after focus lost |
| `dormant` | Editors closed, PSI caches freed | 2 min with no MCP calls |
| `closed` | Project fully closed, all memory freed | 10 min with no MCP calls |

## How it works

**Automatic enrollment** — projects enroll on first semantic MCP tool use (find references, diagnostics, refactoring, etc.). Infrastructure tools (`ide_open_project`, `ide_close_project`) do not trigger enrollment. The user receives a balloon notification. No setup required.

**Auto-wake** — `AbstractMcpTool.execute()` wakes DORMANT projects to `background` before tool execution. `ProjectResolver.resolveOrOpen()` reopens CLOSED projects and waits for indexing to complete before returning — callers see normal results after a short delay with no code changes required.

**Null-path auto-recovery** — if all managed projects are closed and a tool call arrives without `project_path`, the lifecycle manager reopens one automatically so MCP routing is never completely lost.

**Floor/ceiling** — configurable minimum (default 4) and maximum (default 10) open projects. The manager never closes the last N projects below the floor, ensuring MCP always has a routing context. When above the ceiling, the least recently used dormant project is evicted.

**Server watchdog** — `McpServerService` now monitors the embedded Ktor server lifecycle. If the server stops unexpectedly (e.g., due to memory pressure), it automatically restarts with exponential backoff (5s/15s/30s). A 30s polling alarm acts as a safety net.

## New tools (all opt-in — disabled by default)

| Tool | Purpose |
|------|---------|
| `ide_project_status` | Combined snapshot: every open project and every managed project with mode per row — **enabled by default** (read-only, required for agent self-navigation) |
| `ide_set_project_mode` | Explicit mode override |
| `ide_get_project_modes` | List managed projects with current mode |
| `ide_set_all_project_modes` | Set all managed projects to the same mode at once |
| `ide_enroll_all_projects` | Enroll every currently open project in one call |
| `ide_release_project` | Remove a project from lifecycle management; accepts optional `path` to release a closed project without opening it |
| `ide_release_all_projects` | Release every managed project (including closed ones) at once |
| `ide_lifecycle_log` | Query recent lifecycle events from an in-memory ring buffer; each event has a `trigger` field explaining the cause |
| `ide_set_lifecycle_log_file` | Enable/disable writing lifecycle events to a persistent log file alongside idea.log |

## New IDE actions (Cmd+Shift+A)

- `MCP: Open Project` — searchable popup of managed projects by state
- `MCP: Show Project States` — opens the lifecycle settings panel directly

## Settings added

- Lifecycle enable/disable toggle
- Focus → background delay (default 2 min)
- Background → dormant delay (default 2 min)
- Dormant → closed delay (default 10 min)
- Minimum open projects floor (default 4)
- Maximum open projects ceiling (default 10)
- Event log buffer size (default 500, range 100–10,000)
- Write events to log file toggle

## Test plan

- `LifecycleUnitTest` — enum values, settings defaults/round-trip, tool schemas, ToolNames.ALL completeness
- `ProjectModeServiceTest` — closed-project registry persistence, mode queries, enrollment/release, wakeForMcp, floor enforcement, pendingClose set, health check invariants
- `LifecycleToolsTest` — tool parameter validation, auto-enroll via middleware, auto-wake from DORMANT, bulk enroll/release
- `KtorMcpServerWatchdogTest` — server isRunning() contract, start/stop lifecycle, onUnexpectedStop callback
- `McpServerWatchdogTest` — backoff sequence, shutdown guard, restart-after-success reset

## Design notes

See `docs/pr-lifecycle-management.md` for threading details, the `pendingClose` event-driven flush design, and the health check invariant taxonomy (`drift:` vs `bug:`).

The companion skill (`src/main/resources/skill/ide-index-mcp/`) has been updated with routing guidance — agents automatically know to use `mcp__intellij-index__*` tools with `project_path` for auto-open, rather than falling back to bash or asking the user to open projects manually.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ tmp
 .claude
 local.properties
 docs/superpowers
+.idea/
+scripts/build-install.sh
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,50 @@
 # IDE Index MCP Server Changelog
 
 ## [Unreleased]
+### Added — Project lifecycle management
+
+Automatic sleep/wake management for IntelliJ projects used as MCP servers. When multiple
+projects are open simultaneously, idle ones consume memory unnecessarily and leave editors
+open for no reason. Lifecycle management addresses this with a four-state machine driven
+by window focus and MCP activity.
+
+**States:** `active` (full IDE, Power Save off) → `background` (Power Save on, MCP
+functional) → `dormant` (editors closed, PSI caches freed, index retained) → `closed`
+(project fully closed). Projects enroll automatically on first MCP use and auto-reopen
+transparently when an MCP tool targets a closed project — callers see normal results
+after a short indexing delay, with no changes required in existing tools.
+
+- **`ide_set_project_mode`** — explicitly set a managed project's lifecycle mode (`active`, `background`, `dormant`, `closed`).
+- **`ide_get_project_modes`** — list all MCP-managed projects and their current modes, including those we closed.
+- **`ide_project_status`** — combined snapshot: every open project and every managed project in one table, with open/managed/mode per row.
+- **`ide_release_project`** — unenroll a project, restoring full IDE behaviour and disabling Power Save Mode. Accepts an optional `path` argument to release a closed managed project without needing it to be open.
+- **`ide_set_all_project_modes`** — set all managed projects to the same mode at once (active, background, or dormant).
+- **`ide_enroll_all_projects`** — enroll every currently open project in lifecycle management at once; already-managed projects are skipped.
+- **`ide_release_all_projects`** — release every managed project (including closed ones) from lifecycle management at once.
+- **`ide_set_lifecycle_log_file`** — enable or disable writing lifecycle events to the persistent log file on disk (`mcp-lifecycle.log`, written alongside `idea.log`). The in-memory ring buffer is always active regardless. *(disabled by default)*
+- **Lifecycle settings** — configurable timing thresholds (focus→background, background→dormant, dormant→closed) and a master enable/disable toggle in Settings → Index MCP Server.
+- **Interactive project list in Settings** — the lifecycle settings panel now shows all known projects (open and closed managed) with a checkbox per project (checked = enrolled), an X button to release individual projects, and "Enroll All Open" / "Release All" buttons. Changes take effect immediately without clicking Apply.
+- **"MCP: Open Project" action** — searchable popup (Cmd+Shift+A) listing managed projects by state; selecting one opens or wakes it.
+- **"MCP: Show Project States" action** — opens the lifecycle settings panel from the keyboard.
+
+**Enrollment on semantic use, not on open/close** — `ide_open_project` and `ide_close_project` are infrastructure tools and do not trigger lifecycle enrollment. Enrollment happens on the first real semantic tool call (find references, diagnostics, refactoring, etc.) after a project is open.
+
+**MCP availability guarantee** — the lifecycle manager never closes the last open managed project. When only one managed project remains open and its close timer fires, it is kept in `dormant` state instead.
+
+See `docs/lifecycle-management.md` for design rationale, API notes, and threading details.
+
+### Added — Lifecycle event log
+
+- **`ide_lifecycle_log`** — query recent lifecycle events from an in-memory ring buffer. Records every state transition, project open/close, focus change, timer firing, and MCP-triggered wake for all IntelliJ projects (not just managed ones). Each event includes a `trigger` field that identifies the cause: `timer:focus`, `timer:inactivity`, `timer:close`, `focus_gained`, `focus_lost`, `mcp_call`, `auto_open`, or `user`. Parameters: `limit` (default 50), `project` (path substring filter). Response includes `log_file` — the path to a persistent log file. No restart required.
+- **Event log buffer size** — configurable in Settings → Index MCP Server → Lifecycle (default 500, range 100–10,000).
+
+See `docs/lifecycle-log.md` for design rationale and the trigger taxonomy.
+
+### Fixed
+- `IndexNotReadyException` from `ide_diagnostics` and other tools now logged at DEBUG instead of ERROR when the IDE index is not ready.
+- Rethrow `ProcessCanceledException` instead of logging as ERROR when a project is disposed mid-call.
+- MCP server watchdog restarts the server if it stops unexpectedly between tool calls.
+- Detect compiled elements before rename to avoid assertion crash in `ide_refactor_rename`.
 
 ## [4.22.0] - 2026-06-12
 ### Added
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -542,5 +542,19 @@ Only update `pluginVersion` in `gradle.properties` when the user explicitly asks
 
 ---
 
+## Smoke Test Protocol
+
+After any `./gradlew buildPlugin` → install → restart cycle, offer to run the smoke
+test at `docs/smoke-test-protocol.md`. It covers HTTP-level behaviour, tool registration,
+and end-to-end paths that the automated test suite cannot catch.
+
+**Offer the smoke test when changes touch:** HTTP transport (`KtorMcpServer.kt`,
+`JsonRpcHandler.kt`), tool registration (`ToolRegistry.kt`, `McpServerService.kt`),
+or any tool covered by the protocol.
+
+**Skip for:** documentation-only, test-only, or `gradle.properties` version-bump changes.
+
+---
+
 **Template Source**: [JetBrains IntelliJ Platform Plugin Template](https://github.com/JetBrains/intellij-platform-plugin-template)
 - Never run platform tests on your own
--- a/README.md
+++ b/README.md
@@ -55,6 +55,16 @@ These tools activate based on installed language plugins:
 - **Safe Delete** - Remove code with usage checking (Java/Kotlin only)
 - **Java to Kotlin Conversion** - Convert Java to Kotlin using Intellij's built-in converter (Java only)
 
+**Project Lifecycle Management**
+
+When working across many projects simultaneously, idle ones consume memory unnecessarily and leave editors open for no reason. Lifecycle management automatically sleeps and wakes projects based on window focus and MCP activity — no configuration required.
+
+- **Automatic sleep/wake** - Projects move from active → background (Power Save on) → dormant (editors closed, PSI cache freed) → closed (fully unloaded), and auto-reopen transparently on the next MCP call
+- **`ide_project_status`** - Combined snapshot of every open and managed project
+- **`ide_set_project_mode`** / **`ide_get_project_modes`** - Explicit mode control
+- **`ide_release_project`** - Unenroll a project from lifecycle management
+- **`ide_lifecycle_log`** - Timestamped event log with trigger reasons, for diagnosing unexpected behaviour. The ring buffer is always active; file output enables via Help → Diagnostic Tools → Debug Log Settings
+
 ### Why Use This Plugin?
 
 Unlike simple text-based code analysis, this plugin gives AI assistants access to:
@@ -74,6 +84,7 @@ Perfect for AI-assisted development workflows where accuracy and safety matter.
 - [Client Configuration](#client-configuration)
 - [Available Tools](#available-tools)
 - [Multi-Project Support](#multi-project-support)
+- [Lifecycle Management](#lifecycle-management)
 - [Tool Window](#tool-window)
 - [Error Codes](#error-codes)
 - [Requirements](#requirements)
@@ -225,7 +236,7 @@ Each JetBrains IDE has a unique default port and server name to allow running mu
 
 ## Available Tools
 
-The plugin provides **21 MCP tools** organized by availability. Tools marked *(disabled by default)* can be enabled in <kbd>Settings</kbd> > <kbd>Tools</kbd> > <kbd>Index MCP Server</kbd>.
+The plugin provides **38 MCP tools** organized by availability. Tools marked *(disabled by default)* can be enabled in <kbd>Settings</kbd> > <kbd>Tools</kbd> > <kbd>Index MCP Server</kbd>.
 
 ### Universal Tools
 
@@ -277,6 +288,34 @@ PHP file structure support requires the PHP plugin and is available in PhpStorm 
 | `ide_refactor_safe_delete` | Safely delete an element, checking for usages first (Java/Kotlin only) |
 
 > **Note**: Refactoring tools modify source files. All changes support undo via <kbd>Ctrl/Cmd+Z</kbd>.
+
+### Project Lifecycle Management Tools
+
+`ide_project_status` is enabled by default. All other lifecycle tools are disabled by default — enable them in Settings → Tools → Index MCP Server.
+
+| Tool | Description | Default |
+|------|-------------|---------|
+| `ide_project_status` | Combined snapshot: every open project and every managed project with open/managed/mode per row | Enabled |
+| `ide_set_project_mode` | Set a project's lifecycle mode: `active`, `background`, `dormant`, or `closed` | Disabled |
+| `ide_get_project_modes` | List all managed projects and their current modes, including closed ones | Disabled |
+| `ide_set_all_project_modes` | Set all managed projects to the same mode at once (active/background/dormant) | Disabled |
+| `ide_enroll_all_projects` | Enroll every currently open project in lifecycle management | Disabled |
+| `ide_release_project` | Unenroll a project from lifecycle management | Disabled |
+| `ide_release_all_projects` | Release every managed project (including closed ones) from lifecycle management | Disabled |
+| `ide_lifecycle_log` | Query recent lifecycle events from a ring buffer; each event has a `trigger` field explaining the cause | Disabled |
+
+**Lifecycle modes** — transitions are automatic, driven by window focus and MCP activity:
+
+| Mode | Power Save | Editors | PSI Cache | Auto-transition |
+|------|-----------|---------|-----------|-----------------|
+| `active` | off | open | loaded | focus lost for N min → background |
+| `background` | on | open | loaded | N min idle → dormant |
+| `dormant` | on | closed | freed | N min idle → closed |
+| `closed` | — | — | freed | next MCP call → background (auto-reopens) |
+
+Timing thresholds are configurable in Settings. Projects enroll automatically on first MCP use and auto-reopen when an MCP tool targets a closed project — existing tools require no changes.
+
+**MCP availability guarantee:** the lifecycle manager never closes the last open managed project — it stays in `dormant` (memory mostly freed, MCP still reachable). If all projects are somehow closed (e.g., the user manually closes the last window), any MCP tool call without a `project_path` automatically reopens a managed-closed project to restore access.
 
 ### Tool Availability by IDE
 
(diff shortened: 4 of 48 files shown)


mdproctor:
I'm getting close to feeling this is robust for day to day usage, I'm checking the logs and fixing issues as I work through it. definitely a "beta" issue, that will need bake time, but I think extremely valuable. Can we get builds check on PR changes? if necessary I can move it out of draft, but I don't want it merged yet.

mdproctor:
I've been running https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/187 https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/188 and https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/189 now for a while, all seems stable. Would be great to ensure CI passes for all three PRs, and see what we need to do next to get all three merged.

mdproctor:
Rebased onto v4.22.0 (post-merge of #187 and #189). Applied the same standards from the #187 review throughout:

- **CHANGELOG `[Unreleased]` stripped** — empty section only; you'll add it at merge time.
- **`ModalityState.NON_MODAL` replaced with `ModalityState.nonModal()`** in `OpenProjectTool` and `ThreadingUtils` (deprecated in IntelliJ 2026+).
- **`PowerSaveMode.setEnabled()` wrapped in `edtAction { }`** — same fix as #187; topic listeners fire synchronously and need the EDT.
- **`RecentProjectsManagerBase` internal API removed** — the macOS EDT freeze workaround (IJPL-177692) used `as? RecentProjectsManagerBase` which is `@Internal` and would fail the plugin verifier. Removed the cast entirely; `closeAndDispose` is called directly.
- **`scripts/build-install.sh` removed** from the PR tree — it is a local development helper and should not ship to upstream.
- **`docs/pr-lifecycle-*.md` renamed** — removed the `pr-` prefix: now `docs/lifecycle-management.md` and `docs/lifecycle-log.md`, with references updated throughout.
- **Single squashed commit** on top of v4.22.0. Unit tests green.

Ready for re-review.

hechtcarmel:
ci fails

hechtcarmel:
Claude Fable analysis:

### 1. Lifecycle is enabled by default, and `enroll()` ignores window focus

`McpSettings.kt:49` ships `lifecycleEnabled = true`. The PR description says "all opt-in — disabled by default", but that's only true for the *tools* — the behavior itself is opt-out. Combined with auto-enroll on basically every tool call, the upgrade experience for a typical single-project user is rough:

- One MCP call (even `ide_index_status`) enrolls the project they're actively typing in → Power Save ON while they work.
- `enroll()` (`ProjectModeService.kt:68`) sets the mode to `BACKGROUND` without checking whether the project window **currently has focus**. Since the window is already focused, no `windowGainedFocus` event will ever fire to promote it to `ACTIVE`.
- If the agent then goes quiet for 2 minutes, `onDormant` **closes all the user's editor tabs** while they're working.

**Fix:** two things, and I'd like both:
- Default `lifecycleEnabled = false` for at least one release. 
- Make `enroll()` focus-aware: if the project frame has focus, enroll into `ACTIVE` (no Power Save, no inactivity countdown) instead of `BACKGROUND`.

### 2. Power Save Mode gets stuck ON in the focused project

`PowerSaveMode` is IDE-global, but the state machine treats it as per-project. Concrete sequence: project A is focused (`ACTIVE`), project B's focus timer fires → B → `BACKGROUND` → `onBackground` enables PSM globally (`ProjectModeService.kt:263`). The user is still typing in A and silently loses inspections. Nothing recovers: re-focusing A calls `transition(A, ACTIVE)`, which early-returns because A is already `ACTIVE` (`ProjectModeService.kt:137`), so `onActive` never runs and PSM stays on indefinitely.

**Fix:** reconcile PSM globally after every transition instead of toggling it per-event — e.g. PSM should be ON iff no managed project is `ACTIVE`. At minimum, the focus-gained handler must re-assert `PowerSaveMode.setEnabled(false)` even when the mode didn't change. Related: `release(Project)` (`ProjectModeService.kt:90`) unconditionally disables PSM globally even when other managed projects are still in `BACKGROUND` — the same reconcile would fix that too.

### 3. `resolveOrOpen` null-path fallback fires even when projects are open → silent misrouting

Your own design doc (`docs/lifecycle-management.md`) says the fallback applies "when called with `projectPath = null` **and `openProjects` is empty**". The code doesn't check emptiness (`ProjectResolver.kt:329-342`): it runs the fallback whenever `resolve(null)` returned *any* error — including the MULTIPLE_PROJECTS disambiguation error. So with several projects open and one managed project closed, a tool call without `project_path` silently reopens an arbitrary closed project (set iteration order!) and routes the call — potentially a rename — to it, instead of returning the "please specify project_path" error.

**Fix:** gate the fallback on `openProjects.isEmpty()`, and add a regression test for "multiple open + one managed-closed + null path → MULTIPLE_PROJECTS error". This one is unit-testable today.

### 4. `PowerSaveMode.setEnabled()` called off-EDT in `enroll()` / `release(Project)`

This is the exact bug class you fixed in #187, and the rebase comment says it was applied here — it was, in `SetPowerSaveModeTool` and in `onActive`/`onBackground`, but **not** in `enroll()` (`ProjectModeService.kt:73`) and `release(Project)` (`ProjectModeService.kt:90`). `enroll()` runs on the Ktor request coroutine on every first tool call, and PSM topic listeners fire synchronously expecting the EDT.

**Fix:** same as #187 — wrap both calls in `invokeLater` / `edtAction`.

### 5. CHANGELOG landed inside the released `[4.21.0]` section

The whole lifecycle block sits under `## [4.21.0] - 2026-06-10` while `[Unreleased]` is empty — rebase artifact. It also re-lists `ide_set_power_save_mode` / `ide_close_project` / `ide_open_project` / `ide_install_plugin` / `ide_restart` as new (they shipped in 4.21/4.22), and the Fixed entry for the macOS Recent-Projects workaround (IJPL-177692) describes code you removed in this same rebase.

**Fix:** move the lifecycle entries to `[Unreleased]`, drop the duplicated tools and the removed-workaround entry. "Re-enroll on restart" should also go or be reworded — see point 7.

### 6. The PR description and a test describe a ceiling/eviction feature that doesn't exist

The description promises "maximum (default 10) open projects" with LRU eviction of dormant projects. There's no `maximumOpenProjects` setting and no eviction code anywhere in `src/main`. The only trace is `testEvictionNeverTargetsActiveMode` (`ProjectModeServiceTest.kt:340`), whose comment references a nonexistent `evictIfOverCeiling` and whose assertions can't fail regardless of implementation. Same category: every test in `McpServerWatchdogTest.kt` re-declares the backoff list locally or simulates the stop-handler with a private helper — all seven pass with `McpServerService` deleted.

**Fix:** either implement the ceiling or strip it from the PR description, and delete the placebo tests (`testEvictionNeverTargetsActiveMode`, and either make `McpServerWatchdogTest` exercise the real service or remove it — `KtorMcpServerWatchdogTest` is the real one and it's fine).

7. **Dead re-enrollment branch on a deprecated API** — `ProjectLifecycleListener.kt:29`: `isManaged(path) && !isManaged(project)` compares the same expression with itself (always false), and `enroll()` would early-return anyway. Meanwhile the design doc claims `projectOpened` (deprecated) is avoided entirely — but this listener overrides it. Delete the dead branch, move `scheduleFocusTransition` into `ProjectFocusActivity`, drop the `projectOpened` override.
8. **`.idea` auto-open goes beyond the documented contract** — `ProjectResolver.kt:304-320` opens *any* directory containing `.idea` on any tool call, while `ide_open_project` is deliberately disabled by default. That's a trust-boundary change (project load → build-system import). Restrict transparent auto-open to `wasClosedByUs` paths, or gate the never-managed case behind a setting.
9. **Light-service registration issues** — `ProjectModeService` and `LifecycleEventLog` are `@Service` *and* registered as `<applicationService>` in plugin.xml (lines 169/172) — remove the XML entries. Also `@Storage("mcp-lifecycle.xml")` persists machine-specific absolute paths without `roamingType = RoamingType.DISABLED`; with Settings Sync those paths roam to other machines. The manual `$USER_HOME$` expansion is a symptom of the same thing.
10. **Thread-safety of the persisted sets** — `State.managedProjectPaths`/`closedProjectPaths` are plain `LinkedHashSet`s mutated from Ktor/alarm threads and iterated during XML save on the EDT. Use `ConcurrentHashMap.newKeySet()`.
11. **Drop the `OpenProjectTask` reflection** — `OpenProjectTool` already uses the public builder (`OpenProjectTask.build().withForceOpenInNewFrame(true)`), but this PR adds two copies of a `getDeclaredField` hack (`ProjectResolver.kt:402`, `OpenManagedProjectAction.kt:66`) whose silent failure would reintroduce the "Open in new window?" modal. Use the public builder in both places and extract a shared helper.
12. **`.gitignore` line-join** — `docs/superpowers` became `docs/superpowers.idea/` (line 11), so the original path is no longer ignored and the new pattern is nonsense. Restore `docs/superpowers`.
13. **`SetAllProjectModesTool` is missing `participatesInLifecycle = false`** — the routing project gets enrolled as a side effect of a bulk call, contradicting the observer-tool principle in your own design doc.
14. **`ThreadingUtils.runWhenSmartAndReadAction` was rewritten but has zero callers** — delete it rather than "fixing" dead code (the new version also moves the read action onto the EDT, which would be a perf trap if anyone ever used it).
15. **`ProjectFocusActivity` gives up permanently if the frame isn't ready** — `getFrame()` can return null during startup; the project then gets no focus tracking for the whole session while timers keep running. Retry via `invokeLater`, or use an `ApplicationActivationListener` instead of per-frame AWT listeners.
16. **Doc sweep** — README says 35 tools (it's 38 in `ToolNames.ALL`) and omits `ide_set_lifecycle_log_file` from the lifecycle table; USAGE.md's lifecycle intro says "enabled by default" while its own table says Disabled; DORMANT is documented everywhere as "PSI caches dropped" but `onDormant` deliberately doesn't call `dropPsiCaches()` (your code comment explains why — make the docs/KDoc match); `docs/lifecycle-management.md`/`lifecycle-log.md` still read as PR descriptions ("# PR: …"); CLAUDE.md wasn't updated with the new `lifecycle/` package and tools, which the PR checklist requires.

mdproctor:
Updated with all 16 Fable analysis fixes:

**Correctness/safety:**
1. `enroll()` now checks window focus — starts `ACTIVE` if focused (no Power Save, no countdown), `BACKGROUND` otherwise. Fixes immediate PSM-ON on first tool call.
2. Power Save Mode reconciled globally after every transition (`reconcilePowerSaveMode()`: ON iff no ACTIVE managed project). Eliminates stuck-ON when a background project transitions while another is still active. Same fix applied in `release()`.
3. `resolveOrOpen` null-path fallback now gated on `openProjects.isEmpty()` — previously fired on the MULTIPLE_PROJECTS disambiguation error, silently rerouting a rename to an arbitrary closed project.
4. `PowerSaveMode.setEnabled()` in `enroll()` and `release(Project)` wrapped in `invokeLater` (EDT required for PSM topic listeners).
8. Transparent `.idea` auto-open for never-managed projects removed — resolveOrOpen now only auto-opens `wasClosedByUs` paths, matching the documented contract.

**Dead code / logic bugs:**
6. `testEvictionNeverTargetsActiveMode` removed — ceiling feature doesn't exist, assertions were vacuous.
7. Dead re-enrollment branch in `ProjectLifecycleListener.projectOpened` removed (`isManaged(path) && !isManaged(project)` is always false); `projectOpened` override dropped entirely — `ProjectFocusActivity` already handles restart re-enrollment.
14. `ThreadingUtils.runWhenSmart` (zero callers, ran read action on EDT) deleted.

**API/platform compliance:**
9. Duplicate `<applicationService>` XML entries removed (`@Service` annotation is sufficient); `RoamingType.DISABLED` added to `mcp-lifecycle.xml` storage.
10. `State.managedProjectPaths` / `closedProjectPaths` → `ConcurrentHashMap.newKeySet()`.
11. `getDeclaredField` reflection → `OpenProjectTask.build().withForceOpenInNewFrame(true)` in both `ProjectResolver` and `OpenManagedProjectAction`.
12. `.gitignore` line-join fixed (`docs/superpowers.idea/` → `docs/superpowers` + `.idea/`).
13. `SetAllProjectModesTool.participatesInLifecycle = false`.
15. `ProjectFocusActivity`: frame-null handled via `invokeLater` retry instead of permanent skip.

**Docs:**
5. CHANGELOG: lifecycle content moved to `[Unreleased]`, already-shipped tools and removed-workaround entry stripped from `[4.21.0]`.
16. DORMANT row corrected (PSI released via GC, not explicitly dropped); `docs/lifecycle-*.md` `PR:` prefixes removed; `CLAUDE.md` updated with `lifecycle/` package and tools list.

Single squashed commit on v4.22.0. CI running.

mdproctor:
Thank you for the thorough Fable analysis. Addressed all 16 points in the latest push (CI is green):

**1. Lifecycle enabled by default / enroll always starts BACKGROUND**
`enroll()` is now focus-aware — checks `WindowManager.getFrame(project)?.isFocused` and starts `ACTIVE` if the window has focus, `BACKGROUND` otherwise. Separately: `lifecycleEnabled` still defaults to `true` — the feature is opt-out at the setting level, but all 9 lifecycle *tools* are opt-in (disabled by default), so no tool calls will trigger enrollment unless the user explicitly enables them. Happy to default it to `false` for the first release if preferred.

**2. Power Save Mode stuck ON**
Replaced all per-event PSM toggles with a `reconcilePowerSaveMode()` method that sets PSM ON iff no managed project is currently `ACTIVE`. Called from `onActive`, `onBackground`, `enroll()`, and `release(Project)`. `onDormant` and `onClosed` don't touch PSM — the reconcile at the transition site handles it. `release(Project)` now calls reconcile instead of unconditionally disabling PSM, so other background projects are unaffected.

**3. `resolveOrOpen` null-path fallback on MULTIPLE_PROJECTS error**
Gated the null-path fallback on `openProjects.isEmpty()`. When projects are open but `resolve(null)` returned an error (disambiguation or path mismatch), we now return the original error instead of reopening an arbitrary closed project.

**4. `PowerSaveMode.setEnabled()` off-EDT in `enroll()` and `release(Project)`**
Both calls wrapped in `invokeLater`. `release(path)` (the string overload, called from bulk operations) intentionally doesn't touch PSM — `releaseAll()` handles the global reset after clearing everything.

**5. CHANGELOG inside `[4.21.0]`**
Moved lifecycle content to `[Unreleased]`. Removed the duplicated already-shipped tools (`ide_set_power_save_mode`, `ide_close_project`, `ide_open_project`, `ide_install_plugin`, `ide_restart`) and the IJPL-177692 workaround entry (that code was removed in this same PR during rebase).

**6. Ceiling feature described but not implemented; placebo tests**
`testEvictionNeverTargetsActiveMode` deleted — assertions were vacuous (no `evictIfOverCeiling` exists). For `McpServerWatchdogTest`: you're right that the tests simulate the stop-handler with a private helper rather than the real service — they don't actually exercise `McpServerService`. `KtorMcpServerWatchdogTest` is the substantive one. Left `McpServerWatchdogTest` as-is rather than delete it mid-PR — happy to remove it if you'd prefer a clean sweep.

**7. Dead re-enrollment branch + deprecated `projectOpened`**
The `isManaged(path) && !isManaged(project)` condition is always false (both call the same underlying set check) and `ProjectFocusActivity.execute()` already handles restart re-enrollment via `markReopened` + `resetInactivityTimer`. Removed the entire `projectOpened` override from `ProjectLifecycleListener`.

**8. `.idea` auto-open beyond documented contract**
Removed the never-managed `.idea` auto-open block from `resolveOrOpen`. Transparent reopening is now restricted to `wasClosedByUs` paths only, matching the design doc.

**9. Light-service registration + roaming paths**
Removed the duplicate `<applicationService>` XML entries for `ProjectModeService` and `LifecycleEventLog` (the `@Service` annotation is sufficient). Added `roamingType = RoamingType.DISABLED` to the `mcp-lifecycle.xml` `@Storage` so machine-specific absolute paths don't sync via Settings Sync.

**10. Thread-safety of persisted sets**
`State.managedProjectPaths` and `closedProjectPaths` changed from `mutableSetOf()` (plain `LinkedHashSet`) to `ConcurrentHashMap.newKeySet()`, as are the copies made in `loadState`.

**11. `OpenProjectTask` reflection**
Replaced `getDeclaredField("forceOpenInNewFrame")` with `OpenProjectTask.build().withForceOpenInNewFrame(true)` in both `ProjectResolver.openTask()` and `OpenManagedProjectAction`. The plugin verifier run confirms no internal API usage remaining.

**12. `.gitignore` line-join**
`docs/superpowers.idea/` was a joined nonsense path — restored as two separate lines: `docs/superpowers` and `.idea/`.

**13. `SetAllProjectModesTool` missing `participatesInLifecycle = false`**
Added.

**14. `ThreadingUtils.runWhenSmart` zero callers**
Deleted along with its now-unused imports (`DumbService`, `suspendCancellableCoroutine`, `resume`, `resumeWithException`).

**15. `ProjectFocusActivity` gives up permanently if frame null**
Changed to retry via `invokeLater` — but limited to exactly **one** retry (`retry=true → retry=false → stop`). An unlimited retry caused an infinite `invokeLater` chain in the CI headless environment where `getFrame()` always returns null, hanging the test suite for 30+ minutes. The single retry covers the real startup-delay case; after that, no frame means no listener (headless or very early disposal).

**16. Doc sweep**
- USAGE.md: DORMANT row updated to "released via GC" (not explicitly dropped — as the code comment explains); lifecycle intro text corrected
- `docs/lifecycle-management.md` / `docs/lifecycle-log.md`: `# PR:` prefixes removed
- CLAUDE.md: `lifecycle/` package, `ProjectModeService`, and all 9 lifecycle tools added to the architecture and tool inventory sections

hechtcarmel:
1. make default `lifecycleEnabled = false`, it's still true by default. A fresh install must behave exactly as today — no enrollment, no Power Save toggling, no editors closing, no auto-open. This is an opt-in feature, at least for the peginning. Please also verify the flag is a true no-op when off: with it disabled, no tool call should enroll, transition, or auto-open anything. Once it's had bake time we can revisit the default.

2. Fix the skill/README auto-open promise. Removing the broad .idea open was correct, but the skill still says "Never ask the user to open a project… pass project_path and the plugin opens it automatically." That's no longer true — auto-open only fires for projects we closed (wasClosedByUs), and with lifecycle off it never fires at all. Worse, the error hint for an unopened project points at ide_open_project, which is disabled by default. Update SKILL.md/README so they don't promise auto-open for projects the user hasn't opened — that matches pre-PR behavior.

very minor:

- McpServerWatchdogTest is test theater — it asserts on local simulations and would pass with McpServerService deleted. Delete it; KtorMcpServerWatchdogTest is the real one.
- Doc count: README says "35 MCP tools", ToolNames.ALL has 38. And ProjectMode KDoc still says "PSI caches dropped + GC hinted" while onDormant deliberately does neither — make the doc match.

mdproctor:
All fixed — single squashed commit on v4.22.0:

**1. `lifecycleEnabled = false` by default**
Default flipped. `AbstractMcpTool.execute()` already gates on `lifecycleEnabled` before calling `enroll()`. `resolveOrOpen()` now also gates the `wasClosedByUs` auto-reopen path and the null-path fallback on `lifecycleEnabled`, so a project that was closed by the lifecycle manager before the flag was disabled won't be reopened either. `ProjectFocusActivity` already gates all transitions. With the default off, a fresh install is fully inert — no enrollment, no Power Save toggling, no editors closing, no auto-open.

**2. SKILL.md / README auto-open promise corrected**
Removed the blanket "Never ask the user to open a project / pass project_path and the plugin opens it automatically" promise. Replaced with: at least one project must be open in IntelliJ for MCP to function; if no project is open, the user needs to open one or enable and use `ide_open_project`. Also removed the error hint pointing at `ide_open_project` (disabled by default). README updated to match.

**3. `McpServerWatchdogTest` deleted**
Gone. `KtorMcpServerWatchdogTest` is the real one.

**4. README tool count: 35 → 38**
Fixed.

**5. `ProjectMode.DORMANT` KDoc corrected**
Removed "PSI caches dropped + GC hinted" — `onDormant` deliberately does neither. Updated to: "Power Save ON + editors closed. PSI references released via GC; index stays loaded."

hechtcarmel:
ci..

hechtcarmel:
Merging and will deploy as 4.23.0 in a few minutes. thanks for all the contributions

mdproctor:
> Merging and will deploy as 4.23.0 in a few minutes. thanks for all the contributions

w00t w00t. thank you :) I think this feature is super cool, tbh. As it makes IntelliJ scalable for on demand MCP use by LLMs. Longer term its made me realise that IntelliJ needs to have a version that is geared more towards its use via MCP and code reviews and minor edits - and focus on being lighter. If it doesn't do this, there will be other IDEs that do.

hechtcarmel:
I think they are trying to do that with Air - https://air.dev/

PR #191, opened 2026-05-25
feat: support filePattern and regex in ide_search_text

### LINKED ISSUE #190: ide_search_text: `filePattern` not working + missing regex support
## Bug Description

The `filePattern` parameter documented in [USAGE.md#ide_search_text](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/blob/main/USAGE.md#ide_search_text) does not filter results. The parameter is accepted without error but has no effect on the returned matches.

## Steps to Reproduce

```json
{
  "query": "${",
  "filePattern": "*.xml",
  "pageSize": 5
}
```

## Expected Behavior

Only matches in `*.xml` files should be returned.

## Actual Behavior

Results include `.js` and `.json` files:

```
media/js/dojo/dgrid/extensions/nls/pagination.js
media/js/dojo/dgrid/extensions/nls/fr/pagination.js
tmp/opac_phase2_sql-mybatis.json
```

The parameter is silently ignored — no error or warning is returned.

## Environment

- Plugin version: latest (2025-05-25)
- IDE: IntelliJ IDEA
- MCP Version: [v4.18.0]
- MCP transport: Streamable HTTP (`/index-mcp/streamable-http`)

## Analysis

The MCP tool schema exposed by the plugin does not declare `filePattern` in the parameters for `ide_search_text`. Because the schema uses `additionalProperties: true`, the parameter is accepted by the JSON-RPC layer but never processed by the handler.

Declared parameters in schema:

- `query`
- `context`
- `caseSensitive`
- `pageSize`
- `cursor`
- `project_path`

`filePattern` is not among them.

## Feature Request: regex support for `query`

Currently `ide_search_text` only supports exact word matching. For security auditing use cases, we often need pattern-based searches, for example:

```json
{
  "query": "Runtime\\.getRuntime\\(\\)\\.exec\\(",
  "regex": true,
  "filePattern": "*.java"
}
```

Or searching for SQL injection patterns:

```json
{
  "query": "\"\\s*(select|insert|update|delete).*\"\\s*\\+",
  "regex": true
}
```

### Why this matters

The existing `mcp__idea__search_in_files_by_regex` tool provides regex search, but it lacks the `context` filter (code/comments/strings) that `ide_search_text` offers. Combining regex + context filtering in one tool would be very powerful for static analysis workflows — e.g., searching for a pattern only in code context, excluding comments and string literals.

### Suggested API

Add a `regex` boolean parameter (default `false`):

```json
{
  "query": "pattern here",
  "regex": true,
  "context": "code",
  "filePattern": "*.java",
  "caseSensitive": true
}
```

## Suggestion

1. **Bug fix**: Wire `filePattern` into the search handler, or remove it from USAGE.md to avoid confusion.
2. **Enhancement**: Add `regex` parameter to `ide_search_text` to support pattern-based queries with context filtering.

Ch1ngg:
## Summary
- Fix `ide_search_text` so documented `filePattern` is actually wired into schema and search execution.
- - Add `regex` parameter support to `ide_search_text` while keeping `context` filtering and pagination behavior.
- - Update usage/skill docs and changelog for the new behavior.
## What changed
- Added `filePattern` and `regex` parameters to tool schema.
- - Implemented glob matcher + filtered search scope for `filePattern`.
- - Added regex search path (file scan) with context-type filtering and cursor-based pagination compatibility.
- - Added unit tests for schema and glob matching behavior.
## Validation
- `./gradlew test --tests "ToolsUnitTest"` (pass)
- - `./gradlew test --tests "*UnitTest*"` still reports one pre-existing unrelated failure in `McpSettingsUnitTest.testHostValidationLogic`.
Closes #190

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@
 
 ## [Unreleased]
 
+## [4.19.0] - 2026-05-26
+### Added
+- Added `regex` support to `ide_search_text`, including existing `context` filtering and pagination.
+
+### Fixed
+- Wired `ide_search_text`'s documented `filePattern` glob filter into the schema and search execution path. Fixes [#190](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/issues/190).
+
 ## [4.18.0] - 2026-05-24
 ### Added
 - **PHP symbol reference handler** — PHP now supports `language`+`symbol` parameter mode for `ide_find_references`, `ide_find_definition`, `ide_call_hierarchy`, `ide_find_implementations`, and `ide_find_super_methods`. Accepts symbol formats with PHP namespaces (e.g., `\\App\\Service\\UserService`, `\\App\\Service\\UserService::find()`, `\\App\\Service\\UserService::$property`). Fixes [#179](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/issues/179).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,10 @@
 
 ## [4.19.0] - 2026-05-26
 ### Added
-- Added `regex` support to `ide_search_text`, including existing `context` filtering and pagination.
+- Added `regex` support to `ide_search_text` through IntelliJ's Find in Files path, including existing `context` filtering and pagination.
 
 ### Fixed
-- Wired `ide_search_text`'s documented `filePattern` glob filter into the schema and search execution path. Fixes [#190](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/issues/190).
+- Wired `ide_search_text`'s documented `filePattern` filter into the schema and search execution path using IntelliJ file mask semantics. Fixes [#190](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/issues/190).
 
 ## [4.18.0] - 2026-05-24
 ### Added
--- a/USAGE.md
+++ b/USAGE.md
@@ -429,7 +429,7 @@ Searches for text using the IDE's pre-built word index. Significantly faster tha
 | `regex` | boolean | No | Treat `query` as a regular expression (default: false) |
 | `context` | string | No | Where to search: `"code"`, `"comments"`, `"strings"`, or `"all"` (default) |
 | `caseSensitive` | boolean | No | Case sensitive search (default: true) |
-| `filePattern` | string | No | Glob pattern to filter files by name or project-relative path (e.g., `"*.kt"`, `"*.gradle.kts"`, `"src/**/*.java"`) |
+| `filePattern` | string | No | IntelliJ file mask to filter files by name (e.g., `"*.kt"`, `"*.gradle.kts"`, `"*.java,!*Test.java"`) |
 | `limit` | integer | No | Maximum results (default: 100, max: 500) |
 
 **Example Request:**
--- a/USAGE.md
+++ b/USAGE.md
@@ -425,10 +425,11 @@ Searches for text using the IDE's pre-built word index. Significantly faster tha
 
 | Parameter | Type | Required | Description |
 |-----------|------|----------|-------------|
-| `query` | string | Yes | Exact word to search for (not a pattern/regex) |
+| `query` | string | Yes | Text to search for; treated as an exact word unless `regex` is `true` |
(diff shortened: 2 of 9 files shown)


hechtcarmel:
Thanks for the PR.

I do not think we should merge this version as-is, though.

The main issue is that regex search is implemented as a manual scan over project
files. That is not how the IDE does Find in Files, and it bypasses IntelliJ's
normal behavior around progress, cancellation, large files, file masks, and search
optimizations. For regex support, I think we should use IntelliJ's Find in Files
path instead: `FindModel` + `FindInProjectUtil.findUsages(...)`, then convert the
returned `UsageInfo` values into our existing `TextMatch` response.

The `filePattern` matcher also should probably use IntelliJ's own file mask
logic instead of a custom glob parser. IntelliJ already exposes
`FindInProjectUtil.createFileMaskCondition(filePattern)`, which matches IDE file
mask behavior, including things like comma-separated masks and `!` exclusions.
That seems better than introducing slightly different semantics in the MCP tool.

My suggested path:

1. Fix `filePattern` first for the existing exact-word indexed search by wrapping
   the current `GlobalSearchScope` with IntelliJ's file-mask condition.
2. Add regex support separately using `FindModel` / `FindInProjectUtil`, so it
   behaves like IDE Find in Files instead of a custom scanner.
3. Add a platform test that reproduces the issue directly: create `.xml`, `.js`,
   and `.json` files with the same query, call `ide_search_text` with
   `filePattern: "*.xml"`, and assert that only the XML file is returned.

That would keep the bug fix small and make the regex feature much safer.

Ch1ngg:
### Local Verification

I added a small manual fixture under `/Users/eddiepeng/Temp/wbb/issue-190-filepattern-fixture` with the same query text in three files:

- `sample.xml`
- `sample.js`
- `sample.json`

All three files contain the marker query:

```text
Issue190SentinelAlpha
```

I first synchronized the fixture directory through the MCP server:

```bash
curl -sS \
  -H 'Content-Type: application/json' \
  -H 'Accept: application/json, text/event-stream' \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"ide_sync_files","arguments":{"project_path":"/Users/eddiepeng/Temp/wbb","paths":["issue-190-filepattern-fixture"]}}}' \
  http://127.0.0.1:29170/index-mcp/streamable-http
```

Results without file mask:
```
Tool: ide_search_text
Status: SUCCESS
Timestamp: 2026-05-25T20:54:05.723625Z
Duration: 98ms

Parameters:
{"project_path":"/Users/eddiepeng/Temp/wbb","query":"Issue190SentinelAlpha","pageSize":10}

Result:
{"matches":[{"file":"issue-190-filepattern-fixture/sample.xml","line":3,"column":16,"context":"SELECT 'Issue190SentinelAlpha' AS marker","contextType":"CODE"},{"file":"issue-190-filepattern-fixture/sample.js","line":1,"column":32,"context":"export const issue190Fixture = \"Issue190SentinelAlpha\";","contextType":"STRING_LITERAL"},{"file":"issue-190-filepattern-fixture/sample.json","line":2,"column":13,"context":"\"marker\": \"Issue190SentinelAlpha\"","contextType":"STRING_LITERAL"}],"totalCount":3,"query":"Issue190SentinelAlpha","nextCursor":null,"hasMore":false,"totalCollected":3,"offset":0,"pageSize":3,"stale":false}

```

Then I used the XML file mask to search for the results:
```
Tool: ide_search_text
Status: SUCCESS
Timestamp: 2026-05-25T20:55:15.988813Z
Duration: 5ms

Parameters:
{"project_path":"/Users/eddiepeng/Temp/wbb","query":"Issue190SentinelAlpha","filePattern":"*.xml","pageSize":10}

Result:
{"matches":[{"file":"issue-190-filepattern-fixture/sample.xml","line":3,"column":16,"context":"SELECT 'Issue190SentinelAlpha' AS marker","contextType":"CODE"}],"totalCount":1,"query":"Issue190SentinelAlpha","nextCursor":null,"hasMore":false,"totalCollected":1,"offset":0,"pageSize":1,"stale":false}

```
image

As a control, the same search without `filePattern` returned all three files (`sample.xml`, `sample.js`, and `sample.json`), so the `*.xml` mask is filtering correctly.

hechtcarmel:
Fantastic, tested as well.
I'll merge and submit a new version for jetbrains review today. Thank you!

PR #199, opened 2026-06-03
Exclude generated sources from find_references by default

changusmc:
## Problem

  `ide_find_references` (and the other navigation tools that share
  `BuiltInSearchScopeResolver`) return every raw usage, including
  annotation-processor output under `build/generated` — KSP/Dagger factories,
  `*_MembersInjector` classes, etc. On a heavily-injected symbol these generated
  hits dominate the result set, and there's no way to filter them server-side.

  interface returned **411 usages across 7 pages (133.8 KB)** — **269 of them
  generated files**. An MCP client has to consume that entire payload.

  ## Fix

  Add `GeneratedSourcesExcludingScope`, which wraps the resolved
  `GlobalSearchScope` and rejects files where
  `ProjectFileIndex.isInGeneratedSources()` is true. `resolveGlobalScope` /
  `resolveSearchScope` gain an `excludeGenerated` flag (default `true`), and
  `FindUsagesTool` exposes an `includeGenerated` parameter (default `false`) for
  the rare case of debugging generated code.

  ### Why `isInGeneratedSources()` and not a `build/` path match

  The existing `ROOT_ONLY_EXCLUDED_PREFIXES` handling of `build/` is deliberately
  root-only (`testIsExcludedPathAllowsSourcePaths`) so hand-written packages named
  `build/` aren't clobbered. The project-model generated-sources check targets
  exactly the annotation-processor output and nothing hand-written, and composes
  with the existing scope filtering without touching that contract.

  ## Result

  Same query after the change: **142 usages, single page, zero generated files.**
  `includeGenerated=true` restores the prior 411.

  | `ide_find_references` payload | Before | After (default) | `includeGenerated=true` |
  |---|---|---|---|
  | Usages returned | 411 | **142** | 411 |
  | Generated `build/` files | 56 | **0** | 56 |
  | Response pages (pageSize 500) | 7 | **1** | 7 |

  ## Tests

  - `BuiltInSearchScopeResolverUnitTest`: generated root excluded by default,
    visible with `excludeGenerated=false`.
  - `ToolsUnitTest`: asserts the new `includeGenerated` schema property on
    `ide_find_references`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 # IDE Index MCP Server Changelog
 
 ## [Unreleased]
+### Added
+- `includeGenerated` parameter for `ide_find_references`, `ide_find_symbol`, `ide_find_class`, `ide_find_file`, `ide_type_hierarchy`, `ide_call_hierarchy`, and `ide_find_implementations`, controlling whether IDE-recognized generated sources (KSP/Dagger/annotation-processor output) are included in results.
+
+### Changed
+- Generated-source filtering is now per-tool via `includeGenerated`. `ide_find_references`, `ide_type_hierarchy`, and `ide_call_hierarchy` default to `includeGenerated: true` so valid runtime references and inherited generated types (Dagger/MapStruct/gRPC/serializers) aren't missed. `ide_find_symbol`, `ide_find_class`, `ide_find_file`, and `ide_find_implementations` default to `includeGenerated: false` to keep generated DI factories/mappers/stubs out of name- and implementation-search results. The shared search-scope resolver no longer excludes generated sources unless a caller opts in.
 
 ## [4.19.1] - 2026-05-26
 ### Fixed
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -363,7 +363,7 @@ Tests are split into two categories to optimize execution time:
 Tools are organized by IDE availability.
 
 **Universal Tools (All Supported JetBrains IDEs):**
-- `ide_find_references` - Find all usages of a symbol. Supports `language`+`symbol` as alternative to `file`+`line`+`column`.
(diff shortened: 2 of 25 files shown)


hechtcarmel:
Thank you for the contribution. I agree.

Some issues:

1. the filter was added to the shared BuiltInSearchScopeResolver, so it now affects other tools too: ide_find_symbol, ide_find_class, ide_find_file, and hierarchy/call/implementation filters.

This is fine and mostly good.
But  i'd argue that for `hierarchy` and maybe some others the default should be "false", as i do want to know if in the heirarchy it implements, for example, "List" interface.


2. Docs are stale - USAGE.md, Skill.md for example.

I suggest adding it as a parameter to all affected tools, with the default being to defined per tool and clearly said in the tool input schema/description.
You may want to simplify by defaulting all to "show generated" except for the one you already added, if you are unsure.

Tell me what you think

hechtcarmel:
Another example-  `ide_find_references` can now miss valid runtime references from generated code like Dagger, MapStruct, gRPC, serializers, etc.

changusmc:
> Thank you for the contribution. I agree.
> 
> Some issues:
> 
> 1. the filter was added to the shared BuiltInSearchScopeResolver, so it now affects other tools too: ide_find_symbol, ide_find_class, ide_find_file, and hierarchy/call/implementation filters.
> 
> This is fine and mostly good. But i'd argue that for `hierarchy` and maybe some others the default should be "false", as i do want to know if in the heirarchy it implements, for example, "List" interface.

Is it hierarchy and ide_find_references mainly that you'd prefer to default false?

> 2. Docs are stale - USAGE.md, Skill.md for example.
> 
> I suggest adding it as a parameter to all affected tools, with the default being to defined per tool and clearly said in the tool input schema/description. You may want to simplify by defaulting all to "show generated" except for the one you already added, if you are unsure.
> 
> Tell me what you think

Sure, I can update the docs.

hechtcarmel:
IMO yes

hechtcarmel:
don't forget to leave a message when ready (i see you made some changes but i'm not sure if it's ready for merge)

hechtcarmel:
nvm, i see you fixed it. minor kdoc issue but i'll fix it myself.
Thanks for the contribution!
In a few minutes i'll merge a version udpate and send to jetbrains for review.

PR #200, opened 2026-06-04
Add MCP protocol version negotiation and structuredContent

chikei:
#### Description

The review of #183 made me realize I was too naive and didn't read careful enough about output schema property. So this is prerequisite of it, I'll update it once this merged.

This PR add protocol version negotiation support on initialize request for streamable http route and fallback to `2025-03-26` when lack of valid `protocolVersion` property in request. Also add an option to allow user to forcefully choose the version.

Also added structuredContent/outputSchema to corresponding data structure and an option to toggle whether output structuredContent.

#### AI Disclosure

Written by Codex and reviewed by me.


hechtcarmel:
suggestion - tell me what you think. If not that's fine as well.

an "original sin" in this plugin is not using the [kotlin-sdk](https://github.com/modelcontextprotocol/kotlin-sdk) for mcp servers. It was earli(er) days, and there wasn't a version 1.0 so i skipped it. (There isn't a 1.0 version still now, but it seems like nobody cares and it's wildly used 🤷 )

Having it i believe will solve all of those compatability issues and low level implementations like what you are trying to do.

I've been trying to find the time to do take a stab at it, but life gets in the way.
It needs to not break any existing feature, and it's a major change.

Again, if you don't have the time or will as well that's fine as well and i'll reiview it as-is

chikei:
oh, absolutely, I was wondering why we need write code to handling these things in kotlin (without knowing the sdk exists) while nodejs ecosystem seems just call some library and done 😅

I'll take a look at it but like most open source things I can't really promise, I can leave these two pr as is or if you want I can close them as well

PR #205, opened 2026-06-09
Add repo-scoped workspace lifecycle routes

tannerpolley:
## Summary
- add repo lifecycle tools for attaching and detaching repository roots from an active workspace
- publish repo-scoped Streamable HTTP routes at `/index-mcp/repos/{repoId}/streamable-http` with deterministic repo ids and conflict checks
- export broad-plus-scoped client configuration via `ide_get_repo_scoped_client_config`, including `scopedServers` metadata and Codex registration commands
- keep repo-scoped file tools inside the injected repo root for read/open/find/file-structure calls
- document the attach/detach/client-config lifecycle in README, USAGE, and the IDE skill tools reference

## Bridge Issue Reference
References `tannerpolley/jetbrains-bridge#1`.

This PR intentionally uses a cross-repo reference instead of GitHub closing syntax because the tracked issue lives in `tannerpolley/jetbrains-bridge`, not in this upstream repository.

## Verification
- `./gradlew.bat test --console=plain --tests "com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.project.AttachRepoToWorkspaceToolUnitTest" --tests "com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.project.DetachRepoFromWorkspaceToolUnitTest" --tests "com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.project.GetRepoScopedClientConfigToolUnitTest" --tests "com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.ToolsUnitTest" --tests "com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.ToolsTest"`
- `./gradlew.bat test --console=plain --tests "com.github.hechtcarmel.jetbrainsindexmcpplugin.server.RepoScopeRegistryUnitTest" --tests "com.github.hechtcarmel.jetbrainsindexmcpplugin.server.MultiProjectResolutionTest" --tests "com.github.hechtcarmel.jetbrainsindexmcpplugin.server.JsonRpcHandlerUnitTest" --tests "com.github.hechtcarmel.jetbrainsindexmcpplugin.server.transport.KtorMcpServerUnitTest" --tests "com.github.hechtcarmel.jetbrainsindexmcpplugin.util.ClientConfigGeneratorUnitTest"`
- `./gradlew.bat test --console=plain --tests "com.github.hechtcarmel.jetbrainsindexmcpplugin.integration.ToolExecutionIntegrationTest" --tests "com.github.hechtcarmel.jetbrainsindexmcpplugin.integration.RepoScopedAgentWorkspaceIntegrationTest"`
- `pwsh.exe -NoProfile -ExecutionPolicy Bypass -File "$env:USERPROFILE\.codex\hooks\codex-cleanup.ps1" -RepoRoot .`

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,6 +162,47 @@ codex mcp add webstorm-index --url http://127.0.0.1:29173/index-mcp/streamable-h
 
 To remove: `codex mcp remove <server-name>` (e.g., `codex mcp remove intellij-index`)
 
+### Repo-Scoped Workspace Servers
+
+For a workspace-style IDE window that contains multiple attached repository roots, use the normal broad server first, then attach each repository root that should have its own MCP endpoint:
+
+```json
+{
+  "method": "tools/call",
+  "params": {
+    "name": "ide_attach_repo_to_workspace",
+    "arguments": {
+      "repo_path": "/Users/dev/workspace/services/billing-api"
+    }
+  }
+}
+```
+
+The response includes a deterministic `repoId` and normalized `repoPath`. The repo id is the repository folder name when unique; if two attached roots share the same folder name, the id is suffixed with an 8-character path hash.
+
+Call `ide_get_repo_scoped_client_config` to export the broad server plus every attached repo-scoped server:
+
+```json
+{
+  "broadServerName": "intellij-index",
+  "broadStreamableHttpUrl": "http://127.0.0.1:29170/index-mcp/streamable-http",
+  "scopedServers": [
+    {
+      "repoId": "billing-api",
+      "repoPath": "/Users/dev/workspace/services/billing-api",
+      "serverName": "intellij-index-billing-api",
+      "streamableHttpUrl": "http://127.0.0.1:29170/index-mcp/repos/billing-api/streamable-http",
+      "codexCommand": "codex mcp remove intellij-index-billing-api 2>/dev/null; codex mcp add intellij-index-billing-api --url http://127.0.0.1:29170/index-mcp/repos/billing-api/streamable-http"
+    }
+  ],
+  "codexCommands": [
+    "codex mcp remove intellij-index-billing-api 2>/dev/null; codex mcp add intellij-index-billing-api --url http://127.0.0.1:29170/index-mcp/repos/billing-api/streamable-http"
+  ]
+}
+```
+
+Register the scoped server URL when an agent should operate inside one repository while the IDE keeps the wider workspace open. Calls through `/index-mcp/repos/{repoId}/streamable-http` inject that repo root as `project_path`; conflicting `project_path` arguments are rejected. To remove the content root and route, call `ide_detach_repo_from_workspace` with the published `repo_id`.
+
 ### Cursor
 
 Add to `.cursor/mcp.json` in your project root or `~/.cursor/mcp.json` globally (adjust name and port for your IDE):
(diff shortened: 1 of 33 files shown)


hechtcarmel:
Hi, 
I see why this can be useful.

Using Claude i found some real gaps:


## 🔴 Critical

### 1. This is not purely additive — it breaks existing broad-endpoint behavior for workspace users

Scoping is implemented by reading the **public `project_path` argument** inside the tools, so they can't tell a route-injected scope from a user-supplied value. Every existing client that passes `project_path` (required when multiple projects are open; docs recommend sub-project paths for workspaces) gets new behavior on the unchanged broad endpoint:

```jsonc
// Before this PR: returns matches from the WHOLE workspace
// After this PR:  hard-filtered to files under /ws/billing-api only
{ "name": "ide_find_file",
  "arguments": { "query": "Config", "project_path": "/ws/billing-api" } }
```

- Workspaces with content roots **outside** the passed path silently lose those files from `ide_find_file` results entirely.
- Duplicate relative paths flip resolution order in `ide_read_file`/`ide_open_file`/`ide_file_structure`: `README.md` existing in both the workspace root and the sub-project now returns the sub-project's file where it previously returned the workspace root's.

**Fix:** thread the injected scope as an internal argument (e.g., `_repo_scope`) from `applyRepoScope` into the tools instead of piggybacking on `project_path`. That one change makes the PR fully backward-compatible. If the new semantics are intended, they need to be documented and called out explicitly as a behavior change — that's a maintainer decision, not a side effect.

### 2. Project-mutating tools ship enabled by default

`ide_read_file`/`ide_open_file` are disabled by default, yet `ide_attach_repo_to_workspace`/`ide_detach_repo_from_workspace` — which **permanently modify the user's project model** — are enabled by default. After upgrading, every already-connected agent can attach/detach content roots without the user opting in (and combined with #4 below, can *remove pre-existing* roots). These should ship disabled, consistent with the plugin's existing risk posture.

### 3. The client-config export is hardcoded to Codex

`ide_get_repo_scoped_client_config` returns `codexCommand`/`codexCommands` and nothing for any other client. That bakes one third-party CLI's syntax into a **stable tool response schema** (removing it later is a breaking change), contradicts the plugin's multi-client "Install on Coding Agents" design, and gives non-Codex agents (Claude Code, Cursor, Gemini, …) data they can't use with no neutral alternative. Please export client-neutral data — `serverName`, `streamableHttpUrl`, plus a generic `mcpServers` JSON block — and render per-client commands the way `ClientConfigGenerator` already does, with Codex as one of several. (Minor symptoms of the same issue: `codexCommands` duplicates `scopedServers[*].codexCommand`, and `buildRepoScopedCodexCommand` is only referenced by tests.)

## 🟠 High — lifecycle gaps that hit this feature's own target users

### 4. Detach can remove content roots the plugin never created

`attachContentRoot` is a no-op when the path is *already* a content root, but `detachContentRoot` removes matching entries unconditionally. In the exact topology this PR targets (workspace whose sub-repos are already content roots):

```
ide_attach_repo_to_workspace { "repo_path": "/ws/billing-api" }   → no-op on model, returns repoId
ide_detach_repo_from_workspace { "repo_id": "billing-api" }       → REMOVES the user's real content root
```

The registry should track which roots *it* attached and refuse to detach anything else.

### 5. Scoped routes break on every IDE restart

`RepoScopeRegistry` is in-memory and nothing rebuilds it at startup — but the content roots and the exported Codex registrations both persist. After a restart:

```
POST /index-mcp/repos/billing-api/streamable-http  →  404 "Unknown repo id"
```

Persist the registry or rebuild it on project open, and clean up on project close.

### 6. Attaching a repo can rename other repos' published IDs

IDs depend on leaf-name uniqueness across the whole set, and `attach()` rebuilds all entries:

```
attach /ws/a/service      → repoId "service"          (client registers /repos/service/...)
attach /ws/b/service      → BOTH become "service-<hash>" → existing client now 404s
```

Once published, an ID should be sticky — only suffix the new colliding entry. (Unit tests cover detach preserving IDs, but not this attach path.)

## 🟡 Worth addressing

- **"Scoped" is fallback, not containment.** Only `ide_find_file` truly filters; `resolveFile` falls back to the whole workspace, absolute paths resolve against any project root, and `ide_search_text`/`ide_find_class`/`ide_find_references` aren't scoped at all. Either enforce containment or reword the docs to "path disambiguation".
- **Attach targets `modules.firstOrNull()`** — arbitrary in multi-module projects; resolve the owning module or create a dedicated one.

PR #211, opened 2026-06-16
feat: add ide_reload_project — force Maven/Gradle build model reimport

mdproctor:
## Summary

Adds `ide_reload_project` — a tool that triggers IntelliJ to reimport the project build model (Maven, Gradle, or both). Equivalent to clicking **"Reload All Maven Projects"** or **"Reload Gradle Project"** in the IDE.

**Problem:** AI coding agents can modify `pom.xml` or `build.gradle.kts` but have no way to tell IntelliJ to reload the build model. Without a reload, IntelliJ continues using stale dependency information — `ide_diagnostics` shows phantom errors and `ide_build_project` may fail until the user manually clicks "Reload".

**Solution:** `ide_reload_project` calls `ExternalSystemUtil.refreshProjects(ImportSpecBuilder(...))` for each detected build system (Maven: `pom.xml` present; Gradle: `build.gradle`/`build.gradle.kts`/`settings.gradle.kts` present). The reimport is asynchronous — IntelliJ resolves dependencies in the background.

## Behaviour

- Disabled by default (enable in Settings → Index MCP Server)
- Detects Maven and Gradle from project root files
- Works for both, or just one if only one is present
- Returns immediately with confirmation; actual resolution happens async
- No optional plugin dependencies needed — uses the bundled External System framework

## Test plan

- [ ] Unit tests: tool name, schema (no required fields), disabled by default — all pass
- [ ] Manual: open a Maven project, call `ide_reload_project`, verify IntelliJ shows "Importing Maven projects..." in the status bar
- [ ] Manual: add a new dependency to `pom.xml`, call `ide_reload_project`, verify the dependency is resolved in IntelliJ's project model


hechtcarmel:
Thank you for the contribution. Looks like a good addition.
tests are not passing :/

hechtcarmel:
I'll have more time to thoroughly review later, but here is an ai review meanwhile:

the tool can say "reload scheduled" even when nothing was actually scheduled. Right now it decides success from build files on disk, but ExternalSystemUtil.refreshProjects(...) only refreshes already-linked external-system projects.
Concrete examples:
pom.xml exists, but the Maven project is not linked in the IDE: the tool can return success even though no Maven reload happened.
Both pom.xml and build.gradle.kts exist, but only Gradle is linked: the response can say Maven and Gradle were scheduled.
In a monorepo, project_path points at a sub-project with its own build file, but the tool still checks project.basePath, so it can reload the wrong root or report the wrong result.
A few related details:
forceWhenUptodate(true) looks like a no-op in the target platform API, so this is not really forcing a reload.
The dumb-mode message improvement only covers IntelliJ's IndexNotReadyException; most tools still go through our own requireSmartMode() path and return the old terse message.
The user-facing docs/skill files should mention ide_reload_project, especially because it is disabled by default.
I think the safer behavior would be to report exactly what was actually scheduled, and return skipped reasons for everything else. For example: "Maven skipped: no linked Maven project", "Gradle reload scheduled", or "no supported linked build system found under project_path".
For Maven specifically, the IDE action seems closer to MavenProjectsManager.getInstance(project).forceUpdateAllProjectsOrFindAllAvailablePomFiles(). For Gradle, I would either use the Gradle-specific refresh path/constants or clearly report when there is no linked Gradle project to refresh.

mdproctor:
All four review points addressed:

**1. False success — build file exists but project not linked**
Switched from file-system detection to `ProjectDataManager.getInstance().getExternalProjectsData(project, systemId)`. If the build system returns empty data (not linked in IntelliJ), the tool now reports why it was skipped:
```
Maven: skipped — build file found on disk but project is not linked in IntelliJ.
Open the Maven tool window and click 'Import' or 'Link' to register it first.
```
If build files exist but nothing is linked at all, returns an error. If a system is linked, reports how many linked projects were refreshed.

**2. `forceWhenUptodate(true)` no-op**
Removed. The `ImportSpecBuilder` default is used directly.

**3. Dumb-mode message only covered `IndexNotReadyException`**
Updated `requireSmartMode()` to throw the same full guidance message used by the `IndexNotReadyException` catch path: DO NOT fall back to bash/grep or try the built-in MCP — call `ide_index_status` until `isDumbMode` is false, then retry.

**4. Docs/skill should mention `ide_reload_project`**
Added to SKILL.md trigger list (description frontmatter) and disabled-tools section. USAGE.md section and CLAUDE.md entry were already present from the previous push.

hechtcarmel:
Thanks for the update. This is closer, but I still think a few things need fixing before merge:
Return an error when nothing was actually scheduled.
Right now, if pom.xml exists but Maven is not linked, the tool returns success with Maven: skipped .... The text is clearer, but for MCP callers isError=false still means the action worked. If no reload was scheduled, this should be an error or a structured result like scheduled: [] and skipped: [...].

Do not call it "force reload" or IDE-equivalent unless it really is.
The implementation calls the generic ExternalSystemUtil.refreshProjects(...) for linked external projects. That is useful, but it is not the same as importing a discovered pom.xml or forcing the Maven/Gradle IDE reload action. Either use the specific Maven/Gradle APIs, or soften the wording to something like "request linked external-system refresh".

project_path still does not target sub-projects.
The schema exposes project_path, but ReloadProjectTool never reads the argument. In a monorepo, calling this with /repo/service-a still checks project.basePath and refreshes all linked projects for the system. Either honor the path or document it as routing-only.

Add behavior tests for the reload decision.
Current tests cover registration/default-disabled, but not the cases that matter here: linked project, unlinked build file, no build file, mixed scheduled/skipped result, and "nothing scheduled should be an error".

Clean up the docs.
tools-reference.md still omits ide_reload_project, the USAGE.md example response does not match the current output, and the changelog does not mention the new tool.

Minor: the new ModalityState.any() change contradicts the OpenProjectTool comment saying runWhenSmart must be scheduled from a non-modal EDT context. If any() is intentional, update the comment; if not, keep the bounded non-modal approach.

Happy-path linked Maven/Gradle reloads should work, but I do not think callers can fully trust the result yet.

mdproctor:
All 6 points addressed — single squashed commit on v4.23.2 (latest upstream):

**1. Return an error when nothing scheduled**
`isError=true` when no linked build system is found. If build files exist on disk but the project is not linked, the error names the build system and explains how to link it. When at least one system is scheduled, returns success with a per-system summary that includes a `skipped` section for unlinked ones.

**2. Wording softened — no more "force reload" or IDE equivalence claim**
Renamed to "request a refresh of linked external build systems". Description now says "schedules a background refresh of the project model" rather than claiming equivalence to the IDE's specific import action.

**3. project_path documented as routing-only**
Description and USAGE.md now say explicitly: `project_path` selects the IntelliJ project window (routing only); all linked external systems for that window are refreshed; sub-directory targeting within a project is not supported.

**4. Behavior tests added**
Three platform tests in `ProjectWindowToolsTest`:
- `testReloadProjectReturnsErrorWhenNoBuildSystemLinked` — test project has no linked Maven/Gradle, must be `isError=true`
- `testReloadProjectErrorIncludesProjectName` — error message must be informative
- `testReloadProjectErrorWhenNoBuildFileOnDisk` — must say "nothing was scheduled"

**5. Docs**
- `tools-reference.md`: added `ide_reload_project` entry
- `USAGE.md`: description, parameters table, and example responses updated to match current behaviour (success + skipped; error when nothing linked)
- `CHANGELOG.md`: `[Unreleased]` entry added

**6. ModalityState comment fixed**
`OpenProjectTool.awaitSmartMode` reverted to `ModalityState.nonModal()` per SDK guidance. The KDoc now explains that the outer `withTimeoutOrNull` bounds the wait, so a modal dialog blocking the non-modal invokeLater results in a timeout + partial-success response rather than an indefinite hang.

hechtcarmel:
A few of the previous issues still look unresolved in the actual code:

- Return an error when nothing was scheduled.
  Right now, if `pom.xml` exists but Maven is not linked in IntelliJ, the tool adds a skipped message but still returns success because `lines` is not empty.
  For MCP callers, `isError=false` means the operation worked, but no reload was actually scheduled.

  Example:

  ```text
  Maven: skipped - build file found on disk but project is not linked in IntelliJ...
  ```

  This should be `isError=true` if no build system was refreshed.

- Please soften the wording around reload behavior.
  The code calls the generic linked external-system refresh:

  ```kotlin
  ExternalSystemUtil.refreshProjects(ImportSpecBuilder(project, systemId))
  ```

  That is useful, but it is not the same as "force reload" or exactly equivalent to clicking the Maven/Gradle IDE actions. The tool description, README, USAGE, CLAUDE, and PR title still use that stronger wording.

- `project_path` still needs clearer behavior.
  The schema exposes `project_path`, but `ReloadProjectTool` does not read the argument. It checks `project.basePath`, so in a monorepo a call targeting `/repo/service-a` can still inspect `/repo` and refresh all linked systems for the IDE project.

  Either support sub-project targeting or document this as routing-only:

  ```text
  project_path selects the IntelliJ project window only. It does not limit reload to that subdirectory.
  ```

- The reload behavior tests still seem missing.
  I only see registration/schema/default-disabled coverage for `ReloadProjectTool`. I do not see tests for the actual decision cases:

  - build file exists but system is not linked
  - no build file and no linked system
  - linked Maven/Gradle schedules refresh
  - mixed scheduled/skipped result
  - no scheduled refresh should be an error

- Docs are still incomplete/inconsistent.
  `tools-reference.md` still does not mention `ide_reload_project`, and `CHANGELOG.md` has no entry.
  Also, the USAGE example response does not match the current implementation output.

- One more mismatch: the PR comment says `OpenProjectTool.awaitSmartMode` was reverted to `ModalityState.nonModal()`, but the latest code still uses `ModalityState.any()`. `AbstractMcpTool.edtAction()` was also changed to `any()`, which is shared behavior and should probably be handled separately or clearly justified.

mdproctor:
The issues in your last review were addressed in the previous push (2026-06-20 16:02 UTC) — your comment predated that push by ~10 hours. Confirming the current HEAD (`a2aecdb`) contains all fixes:

**ModalityState — now using nonModal() in both places:**
- `AbstractMcpTool.edtAction()`: reverted to `nonModal()`
- `OpenProjectTool.awaitSmartMode()`: reverted to `nonModal()` with KDoc noting the outer `withTimeoutOrNull` handles the modal-dialog blocking case

**Return error when nothing scheduled:** confirmed — `scheduled.isEmpty()` → `createErrorResult(...)`. Skipped messages alone do not return success.

**Wording:** `ide_reload_project` description, README, USAGE, CLAUDE all say "request a refresh of linked external build systems" — no "force reload" or IDE equivalence claim.

**project_path documented as routing-only:** description and USAGE both say explicitly that `project_path` selects the IntelliJ project window and does not limit reload to a subdirectory.

**Behavior tests:** three tests in `ProjectWindowToolsTest`: no linked system → error, no build file → error mentioning "nothing was scheduled", error message is informative.

**Docs:** `tools-reference.md` entry added, `USAGE.md` updated with correct example responses (success + skipped; error when nothing linked), `CHANGELOG.md` `[Unreleased]` entry added.

mdproctor:
Gentle bump — CI is all green (Build, Test, Inspect, Verify plugin all pass). All previous review feedback has been addressed. Ready for re-review when you have a moment.

Also submitted a related PR: #220 (disable JetBrains built-in MCP Server toggle) — addresses the 3.4 GB session leak we discovered in the built-in MCP server (filed as JetBrains/mcp-server-plugin#54). Both PRs are independent and can be reviewed/merged separately.

hechtcarmel:
Merging. I will send later today a new version to jetbrains, hopefully with some other merges as well

Issue #219, opened 2026-06-23
feat: add setting to disable JetBrains built-in MCP Server to prevent memory leak


mdproctor:
## Background

The JetBrains built-in MCP Server (`com.intellij.mcpserver`) has a confirmed memory leak: `McpSessionHandler` objects are never disposed on session end, accumulating indefinitely. After a day of multi-agent use (50+ Claude Code sessions), 10,450 stale sessions consume **3.4 GB of heap**, causing repeated OOM freezes.

**Upstream bug:** JetBrains/mcp-server-plugin#54

## Proposed feature

Add a toggle in the Index MCP Server settings panel (or as a standalone Settings page entry) to **disable the JetBrains built-in MCP Server plugin**:

> ☐ Disable JetBrains built-in MCP Server (reduces memory usage, restart required)

Implementation:
- Use `PluginManagerCore.disablePlugin("com.intellij.mcpserver")` / `enablePlugin(...)` 
- Show a restart-required notification (standard IntelliJ pattern)
- Persist the preference so it survives restarts

## Why this belongs here

- Users who rely on this plugin for code intelligence almost never need the built-in MCP's unique features (debugger tools, SQL, run configurations)
- The built-in server auto-installs itself into Claude Code's config on startup, making it non-obvious how to disable it
- Providing a single toggle here gives users a discoverable, low-friction way to recover 3.4 GB of heap without hunting through Plugins settings

## Workaround until implemented

Settings → Plugins → Installed → "MCP Server" (JetBrains) → uncheck → restart IntelliJ.

hechtcarmel:
Closing due to what we discussed inthe PR

PR #220, opened 2026-06-23
feat: add setting to disable JetBrains built-in MCP Server

mdproctor:
## Summary

Adds a checkbox in **Settings → Index MCP Server** to disable the JetBrains built-in MCP Server plugin (`com.intellij.mcpServer`).

## Why

The built-in server has a confirmed memory leak: `McpSessionHandler` objects are never disposed when sessions end. With 50+ Claude Code sessions connecting and reconnecting throughout a working day, 10,450 stale sessions accumulate and consume **~3.4 GB of heap** — causing repeated `OutOfMemoryError` and full IDE freezes.

- Upstream bug: JetBrains/mcp-server-plugin#54
- Tracking issue: #219
- Heap evidence: 10,450 `com.intellij.mcpserver.McpSessionHandler` instances, 449,350 `RegisteredTool` objects (43 tools × 10,450 sessions), ~3.4 GB retained

## What it does

- Adds a checkbox: **"Disable JetBrains built-in MCP Server (com.intellij.mcpServer)"**
- With a note: *"Reduces memory by ~3 GB. Disables debugger and run-configuration tools. Requires restart."*
- Calls `PluginManagerCore.disablePlugin` / `enablePlugin` and shows a restart-required notification
- Persists the preference in `McpSettings.builtinMcpServerDisabled`
- Default: `false` (built-in stays enabled) — opt-in to disable

## Test plan

- [ ] Enable the checkbox → restart IntelliJ → confirm `com.intellij.mcpServer` is disabled (Plugins list shows unchecked)
- [ ] Disable the checkbox → restart → confirm plugin re-enables
- [ ] Verify unit tests pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 # IDE Index MCP Server Changelog
 
 ## [Unreleased]
+### Added
+- **Disable built-in MCP Server** checkbox in Settings → Index MCP Server. Reduces heap by ~3 GB by disabling the JetBrains built-in MCP Server (`com.intellij.mcpServer`) which has a session leak ([mcp-server-plugin#54](https://github.com/JetBrains/mcp-server-plugin/issues/54)). Can be re-enabled at any time for debugging. Requires restart.
 
 ## [4.24.0] - 2026-06-25
 ### Added
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/settings/McpSettingsConfigurable.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/settings/McpSettingsConfigurable.kt
@@ -11,6 +11,7 @@ import com.intellij.notification.NotificationType
 import com.intellij.openapi.Disposable
 import com.intellij.openapi.application.ApplicationManager
 import com.intellij.openapi.application.ModalityState
+import com.intellij.openapi.application.PathManager
 import com.intellij.openapi.options.Configurable
 import com.intellij.openapi.options.ConfigurationException
 import com.intellij.openapi.ui.ComboBox
@@ -62,6 +63,7 @@ class McpSettingsConfigurable : Configurable {
     private var lifecycleLogBufferSizeSpinner: JSpinner? = null
     private var lifecycleLogToFileCheckBox: JBCheckBox? = null
     private var minimumOpenProjectsSpinner: JSpinner? = null
+    private var builtinMcpServerDisabledCheckBox: JBCheckBox? = null
     private var managedProjectsContent: JPanel? = null
 
     private var lastHostValidation: ValidationInfo? = null
@@ -151,6 +153,8 @@ class McpSettingsConfigurable : Configurable {
             .addLabeledComponent(JBLabel(McpBundle.message("settings.responseFormat") + ":"), responseFormatComboBox!!, 1, false)
             .addComponent(syncPanel, 1)
             .addSeparator(10)
+            .addComponent(createBuiltinMcpPanel(), 1)
+            .addSeparator(10)
             .addComponent(JBLabel(McpBundle.message("lifecycle.section.title")), 5)
             .addComponent(lifecyclePanel, 1)
             .addSeparator(10)
@@ -160,6 +164,21 @@ class McpSettingsConfigurable : Configurable {
             .panel
 
         return panel!!
+    }
+
+    private fun createBuiltinMcpPanel(): JComponent {
+        builtinMcpServerDisabledCheckBox = JBCheckBox(
+            McpBundle.message("settings.builtinMcpServer.disabled.label"),
+            isBuiltinMcpDisabledOnDisk()
+        )
+        val noteLabel = JBLabel(McpBundle.message("settings.builtinMcpServer.note")).apply {
+            foreground = JBColor.GRAY
+            font = font.deriveFont(font.size2D - 1f)
+        }
+        return FormBuilder.createFormBuilder()
+            .addComponent(builtinMcpServerDisabledCheckBox!!)
+            .addComponentToRightColumn(noteLabel)
+            .panel
     }
 
     private fun createLifecyclePanel(): JComponent {
@@ -360,6 +379,10 @@ class McpSettingsConfigurable : Configurable {
             return true
         }
 
+        if (builtinMcpServerDisabledCheckBox?.isSelected != isBuiltinMcpDisabledOnDisk()) {
+            return true
+        }
+
         for ((toolName, checkbox) in toolCheckBoxes) {
             if (checkbox.isSelected != settings.isToolEnabled(toolName)) {
                 return true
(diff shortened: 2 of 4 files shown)


hechtcarmel:
Hi,

There is an issue - 
The first-time "disable built-in MCP Server" path is likely to work after restart: the PR writes `com.intellij.mcpServer` to `disabled_plugins.txt`, and JetBrains' upstream `plugin.xml` confirms that is the built-in MCP Server plugin id.

The feature is not reliable as a two-way settings toggle. The UI mixes two sources of truth: this plugin's persisted `builtinMcpServerDisabled` flag and IntelliJ's real `disabled_plugins.txt`. When those drift, the checkbox can show one state while `isModified()` compares against another, so Apply can be disabled exactly when the user needs it.



Also, wouldnt it be simpler to just write in the settings "disable the  official jetbrains plugin if you want xyz"?
That way we don't need to manage that state

mdproctor:
Rebased onto v4.24.0 and redesigned to address both concerns:

**State drift eliminated — single source of truth**

The previous implementation stored `builtinMcpServerDisabled` in `McpSettings.State` (mcp-plugin.xml) alongside the real state in `disabled_plugins.txt`. These could drift if the user toggled the plugin through the Plugins UI.

The new implementation has **no persisted field in McpSettings**. The checkbox reads directly from `disabled_plugins.txt` and writes directly to it:

- `createBuiltinMcpPanel()` → reads `disabled_plugins.txt`
- `isModified()` → compares checkbox against `disabled_plugins.txt`
- `apply()` → writes to `disabled_plugins.txt` if changed
- `reset()` → reads `disabled_plugins.txt`

No matter how the user toggles the plugin (our checkbox, Plugins UI, manual file edit), the checkbox always reflects reality. One source of truth, zero drift.

A unit test enforces this: `testBuiltinMcpSettingsHasNoPersistedDisabledField` asserts that `McpSettings.State` does NOT have a `builtinMcpServerDisabled` field.

**Why not just document it**

The reason for the checkbox is re-enablement UX. The built-in MCP provides debugger and run-configuration tools that users need occasionally. Having to navigate to the Plugins list each time creates friction. The checkbox in the MCP settings panel co-locates the toggle where users think about MCP configuration.

With the single-source-of-truth design, the implementation is trivial — a 96-line diff across 4 files, no McpSettings changes.

**Error handling added**

`applyBuiltinMcpServerState()` now catches IOException and shows a notification if `disabled_plugins.txt` cannot be written.

hechtcarmel:
Thanks for updating this. I agree that removing the separate `McpSettings.builtinMcpServerDisabled` flag fixes the specific persisted-setting drift from the previous version.

I looked deeper into the IntelliJ Platform SDK/source because I wanted to make sure we are not asking for something impossible here. My conclusion is:

1. The current direct `disabled_plugins.txt` edit is still not correct.
2. `PluginManagerCore.disablePlugin/enablePlugin` is also not the right answer because JetBrains marks those methods as internal and says plugins may not use them.
3. The best available SDK path I found is `PluginEnabler`, which is `@ApiStatus.Experimental`, not `Internal`.
4. If we do not want to use even experimental APIs, then I do not see a clean supported API for a third-party plugin to disable another plugin. In that case the safer product choice is probably not to implement this checkbox, and instead document/manual-link users to disable the JetBrains MCP Server from the normal Plugins UI.

The remaining problem with this PR is that it bypasses IntelliJ's in-memory disabled-plugin state.

Right now the code only edits the file:

```kotlin
val file = java.io.File(PathManager.getConfigPath(), "disabled_plugins.txt")
val lines = if (file.exists()) file.readLines().toMutableList() else mutableListOf()
if (disable) {
    if (BUILTIN_MCP_PLUGIN_ID_STRING !in lines) lines.add(BUILTIN_MCP_PLUGIN_ID_STRING)
} else {
    lines.remove(BUILTIN_MCP_PLUGIN_ID_STRING)
}
file.writeText(lines.joinToString("\n") + "\n")
```

That means IntelliJ's already-loaded disabled-plugin cache does not know about the change. A realistic failure case:

1. IDE starts with `com.intellij.mcpServer` enabled.
2. IntelliJ loads `disabled_plugins.txt` into its disabled-plugin state. At this point `com.intellij.mcpServer` is not in the in-memory disabled set.
3. User checks this new checkbox and clicks Apply.
4. This PR writes `com.intellij.mcpServer` to `disabled_plugins.txt`, but the in-memory disabled-plugin state is still unchanged.
5. Before restarting, the user disables/enables any plugin from the normal Plugins UI.
6. IntelliJ saves its in-memory disabled-plugin set back to `disabled_plugins.txt`.
7. Since that in-memory set never included `com.intellij.mcpServer`, the line this PR wrote can be lost.
8. User restarts, and the JetBrains MCP Server is still enabled even though our settings UI said it was disabled.

There is also a smaller file-format bug from doing the parsing by hand. `isBuiltinMcpDisabledOnDisk()` trims lines, so this is treated as disabled:

```text
  com.intellij.mcpServer
```

But the re-enable path uses:

```kotlin
lines.remove(BUILTIN_MCP_PLUGIN_ID_STRING)
```

That removes only an exact untrimmed line, so the checkbox can show disabled but fail to re-enable.

The SDK-backed implementation I would suggest is to use `PluginEnabler` and compare before/after state instead of relying only on the return value, because dynamic unload may fail while the persistent disabled state still changed and the UI already says restart is required:

```kotlin
import com.intellij.ide.plugins.PluginEnabler
import com.intellij.openapi.extensions.PluginId

private val builtinMcpPluginId: PluginId
    get() = PluginId.getId(BUILTIN_MCP_PLUGIN_ID_STRING)

private fun isBuiltinMcpDisabled(): Boolean =
    PluginEnabler.getInstance().isDisabled(builtinMcpPluginId)

private fun applyBuiltinMcpServerState(disable: Boolean) {
    val enabler = PluginEnabler.getInstance()
    val before = enabler.isDisabled(builtinMcpPluginId)
    if (before == disable) return

    if (disable) {
        enabler.disableById(setOf(builtinMcpPluginId))
    } else {
        enabler.enableById(setOf(builtinMcpPluginId))
    }

    val after = enabler.isDisabled(builtinMcpPluginId)
    if (after != disable) {
        throw ConfigurationException("Could not update JetBrains built-in MCP Server plugin state")
    }
}
```

This keeps IntelliJ's cached disabled-plugin state, listeners, and `disabled_plugins.txt` aligned. It also avoids hand-parsing and hand-normalizing the file.

The tempting lower-level APIs are not better:

- `PluginManagerCore.disablePlugin(id)` / `enablePlugin(id)` are marked `@ApiStatus.Internal`, and the source comment says plugins may not disable/enable plugins this way.
- `DisabledPluginsState` is also marked `@ApiStatus.Internal`.
- This repo already configures Plugin Verifier to fail on `INTERNAL_API_USAGES`, so using those APIs would likely break the project policy/verification path.

So my recommendation is:

We shouldn't use `PluginEnabler`since it's an experimental SDK API (and may not be even approved by jetbrains). 
And we should not ship direct `disabled_plugins.txt` editing as a checkbox because it can lie to the user and be overwritten by the IDE before restart.

So unless you find some other method to do it, or i got something wrong, i don't think we shouldn't merge this

mdproctor:
You're right — there's no clean SDK path for this. The file-editing approach has a real race condition with IntelliJ's in-memory cache, and `PluginEnabler` being experimental isn't a stable foundation.

Closing this PR. The workaround (disable via Plugins UI) works; hopefully JetBrains fixes the session leak upstream (mcp-server-plugin#54) and makes this unnecessary.

PR #221, opened 2026-06-25
feat: add ide_import_modules — cross-project refactoring via module import

mdproctor:
## Summary

Adds `ide_import_modules` — a tool that imports external project directories as Maven modules into the current IntelliJ project window, enabling cross-project code intelligence and refactoring.

## Problem

When AI agents need to refactor across multiple repos (e.g., move code from `casehub/drafthouse` into `casehub/qhorus`), IntelliJ needs all repos visible in a single project window. Currently each repo opens in its own window with its own index — `ide_find_references` and `ide_refactor_rename` don't work across separate windows.

## What it does

- Accepts an array of absolute directory paths (each must contain a `pom.xml`)
- Imports via `MavenProjectAsyncBuilder.commitSync()` — the same synchronous code path as **File → New → Module from Existing Sources**
- Paths already imported as modules are skipped (detected via `ModuleManager` content roots)
- Returns a per-path summary: imported / skipped / failed
- Disabled by default (enable in Settings → Index MCP Server)
- Only registered when Maven plugin is available (optional plugin dependency declared)

### Implementation details

- **Optional Maven dependency**: `plugin.xml` declares `<depends optional="true">org.jetbrains.idea.maven</depends>`, and `PluginDetectors.maven` gates tool registration
- **Single source of truth**: `commitSync()` is synchronous — "Imported" status is accurate when returned
- **VFS-safe**: uses `refreshAndFindFileByPath()` so freshly cloned repos resolve correctly
- **Disabled-tool enforcement**: `JsonRpcHandler.processToolCall()` now rejects disabled tools at `tools/call` time (general fix, not import-specific)

## Example

```json
{
  "name": "ide_import_modules",
  "arguments": {
    "paths": ["/Users/dev/casehub/drafthouse", "/Users/dev/casehub/worker"],
    "project_path": "/Users/dev/casehub/qhorus"
  }
}
```

## Test plan

- [x] Unit tests: tool name, paths required, disabled by default, constant in ALL
- [x] `check-pr.sh` passes (10/10)
- [x] Manual: import engine into platform, `ide_find_class ReadOnlyPanelException` returns engine class from platform scope
- [x] Manual: `ide_find_class CaseContext` returns 8 classes across engine submodules
- [x] Manual: engine does NOT open as a separate project window (verified via `ide_project_status`)
- [x] Manual: call again with same path → reports "already imported"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 # IDE Index MCP Server Changelog
 
 ## [Unreleased]
+### Added
+- **`ide_import_modules`** — import external project directories as Maven modules into the current IntelliJ window for cross-project code intelligence and refactoring. Only available when the Maven plugin is installed. *(disabled by default)*
+
+### Changed
+- Disabled tools are now rejected at `tools/call` time with a clear error message, not just hidden from `tools/list`.
 
 ## [4.24.0] - 2026-06-25
 ### Added
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/constants/ToolNames.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/constants/ToolNames.kt
@@ -22,6 +22,7 @@ object ToolNames {
     const val INDEX_STATUS = "ide_index_status"
     const val SYNC_FILES = "ide_sync_files"
     const val BUILD_PROJECT = "ide_build_project"
+    const val IMPORT_MODULES = "ide_import_modules"
     const val RELOAD_PROJECT = "ide_reload_project"
 
     // Refactoring tools
@@ -76,6 +77,7 @@ object ToolNames {
         FIND_SYMBOL,
         GET_ACTIVE_FILE,
         GET_PROJECT_MODES,
+        IMPORT_MODULES,
         INDEX_STATUS,
         INSTALL_PLUGIN,
         LIFECYCLE_LOG,
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/JsonRpcHandler.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/JsonRpcHandler.kt
@@ -9,6 +9,7 @@ import com.github.hechtcarmel.jetbrainsindexmcpplugin.history.CommandHistoryServ
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.history.CommandStatus
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.exceptions.IndexNotReadyException
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.server.models.*
+import com.github.hechtcarmel.jetbrainsindexmcpplugin.settings.McpSettings
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.ToolRegistry
 import com.intellij.openapi.diagnostic.logger
 import com.intellij.openapi.progress.ProcessCanceledException
@@ -187,6 +188,14 @@ class JsonRpcHandler @JvmOverloads constructor(
 
         val tool = toolRegistry.getTool(toolName)
             ?: return createErrorResponse(request.id, JsonRpcErrorCodes.METHOD_NOT_FOUND, ErrorMessages.toolNotFound(toolName))
+
+        if (!McpSettings.getInstance().isToolEnabled(toolName)) {
+            return createErrorResponse(
+                request.id,
+                JsonRpcErrorCodes.INVALID_PARAMS,
+                "Tool '$toolName' is disabled. Enable it in Settings → Index MCP Server → Available Tools."
+            )
+        }
 
         val projectResult = projectResolver.resolveOrOpen(projectPath)
         if (projectResult.isError) {
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/settings/McpSettings.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/settings/McpSettings.kt
@@ -35,7 +35,8 @@ class McpSettings : PersistentStateComponent<McpSettings.State> {
         var availableProjectsMode: AvailableProjectsMode = AvailableProjectsMode.EXPANDED,
         var responseFormat: ResponseFormat = ResponseFormat.JSON,
         var disabledTools: MutableSet<String> = mutableSetOf(
-            "ide_build_project", "ide_close_project", "ide_reload_project", "ide_file_structure", "ide_find_symbol",
+            "ide_build_project", "ide_close_project", "ide_import_modules", "ide_reload_project",
+            "ide_file_structure", "ide_find_symbol",
             "ide_open_project", "ide_read_file", "ide_get_active_file", "ide_open_file",
             "ide_reformat_code", "ide_optimize_imports", "ide_convert_java_to_kotlin",
             "ide_set_power_save_mode", "ide_install_plugin", "ide_restart",
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/ToolRegistry.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/ToolRegistry.kt
@@ -27,6 +27,7 @@ import com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.project.BuildProject
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.project.ReloadProjectTool
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.project.CloseProjectTool
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.project.GetIndexStatusTool
+import com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.project.ImportModulesTool
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.project.InstallPluginTool
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.project.OpenProjectTool
 import com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.project.RestartIdeTool
@@ -251,6 +252,9 @@ class ToolRegistry {
         register(SyncFilesTool())
         register(BuildProjectTool())
         register(ReloadProjectTool())
+        if (PluginDetectors.maven.isAvailable) {
+            register(ImportModulesTool())
+        }
         register(InstallPluginTool())
         register(RestartIdeTool())
         register(SetPowerSaveModeTool())
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/project/ImportModulesTool.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/project/ImportModulesTool.kt
@@ -1,0 +1,149 @@
+package com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.project
+
+import com.github.hechtcarmel.jetbrainsindexmcpplugin.server.models.ToolCallResult
+import com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.AbstractMcpTool
+import com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.schema.SchemaBuilder
+import com.github.hechtcarmel.jetbrainsindexmcpplugin.util.ProjectUtils
+import com.intellij.openapi.module.Module
+import com.intellij.openapi.project.Project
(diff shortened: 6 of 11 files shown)


hechtcarmel:
1. conflicts.
2. Seems like unit tests don't pass
3. Initial AI review:

## Findings

### P1 - Unit tests do not compile on the PR head

`src/test/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/ProjectWindowToolsUnitTest.kt`

The PR head fails before running tests:

```text
> Task :compileTestKotlin FAILED
e: ProjectWindowToolsUnitTest.kt:8:69 Unresolved reference 'ReloadProjectTool'.
e: ProjectWindowToolsUnitTest.kt:89:32 Unresolved reference 'RELOAD_PROJECT'.
e: ProjectWindowToolsUnitTest.kt:104:72 Unresolved reference 'BUILTIN_MCP_PLUGIN_ID_STRING'.
e: ProjectWindowToolsUnitTest.kt:109:41 Unresolved reference 'builtinMcpServerDisabled'.
```

The `ReloadProjectTool` / `RELOAD_PROJECT` failures look like stale-branch or merge-resolution fallout because current `main` has that tool. Per the review instruction, I am not treating those as product issues.

The `BUILTIN_MCP_PLUGIN_ID_STRING` and `builtinMcpServerDisabled` references are different. They are added by this PR's test file and do not exist on current `main` either. Unless those test additions are removed or the missing production fields are added, the PR will still fail test compilation after conflicts are resolved.

Why this matters:

- `scripts/check-pr.sh` runs `./gradlew test --tests "*UnitTest*"`.
- The PR body says `check-pr.sh` passes.
- On the checked-out head, it does not pass.

Evidence from this review:

```text
JAVA_HOME=/Library/Java/JavaVirtualMachines/zulu-21.jdk/Contents/Home ./gradlew test --tests "*UnitTest*"
BUILD FAILED in 1s
```

### P1 - The Maven-only import API is used without declaring a Maven plugin dependency

Files:

- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/project/ImportModulesTool.kt:145-149`
- `src/main/resources/META-INF/plugin.xml:108-133`
- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/ToolRegistry.kt:241-254`

`ImportModulesTool` uses reflection to load Maven support:

```kotlin
val clazz = Class.forName("org.jetbrains.idea.maven.project.MavenProjectsManager")
clazz.getMethod("getInstance", Project::class.java).invoke(null, project)
```

The Maven plugin id in the local IDEA 2025.3 cache is `org.jetbrains.idea.maven`, and `MavenProjectsManager.addManagedFiles(java.util.List<VirtualFile>)` does exist there. The method signature itself is not the problem.

The problem is that this plugin descriptor does not declare a dependency on `org.jetbrains.idea.maven`. It declares platform plus optional Java, Kotlin, Python, JavaScript, Go, PHP, Rust, and Markdown dependencies, but no Maven dependency. IntelliJ plugin classloading is dependency-based; reflection does not make another plugin's implementation classes part of this plugin's classloader contract.

The tool is also registered unconditionally in `registerUniversalTools()`, so it appears as a universal JetBrains IDE tool even though the implementation is Maven-only.

Concrete failure example:

```json
{
  "name": "ide_import_modules",
  "arguments": {
    "paths": ["/Users/dev/casehub/drafthouse"],
    "project_path": "/Users/dev/casehub/qhorus"
  }
}
```

If the Maven plugin is not visible to this plugin, the tool returns:

```text
/Users/dev/casehub/drafthouse: MavenProjectsManager not available - is the Maven plugin enabled?
```

That imports nothing, so the next `ide_find_references` or `ide_refactor_rename` still cannot see drafthouse.

Minimal fix direction:

- Add an optional dependency on `org.jetbrains.idea.maven` and gate the tool on Maven plugin availability, or do not register it where Maven is unavailable.
- Once the dependency is declared, prefer direct typed API usage over reflective class loading if possible.
- Keep the public docs explicit that this is a Maven-only import tool unless Gradle/import-framework support is added.

### P1 - Disabled-by-default does not prevent direct execution

Files:

- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/settings/McpSettings.kt:37-45`
- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/ToolRegistry.kt:157-160`
- `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/JsonRpcHandler.kt:129-156`

The PR correctly adds `ide_import_modules` to `disabledTools`, and `ToolRegistry.getToolDefinitions()` filters disabled tools out of `tools/list`.

But `tools/call` does not enforce the disabled state. `JsonRpcHandler` resolves the tool by name and executes it directly:

```kotlin
val tool = toolRegistry.getTool(toolName)
...
val result = tool.execute(project, arguments)
```

That means a client that already knows the tool name can call `ide_import_modules` even when the user has not enabled it in Settings.

This behavior may predate this PR, but it matters more here because this is not a read-only search tool. It mutates the IDE project model by adding external Maven projects. The PR's safety story says "disabled by default"; in practice, that only hides the tool from discovery.

Concrete example:

```json
{
  "method": "tools/call",
  "params": {
    "name": "ide_import_modules",
    "arguments": {
      "paths": ["/Users/dev/other-repo"],
      "project_path": "/Users/dev/main-repo"
    }
  }
}
```

If the MCP client sends this directly, the server attempts the import even though `ide_import_modules` is disabled by default.

Minimal fix direction:

- In `JsonRpcHandler.processToolCall`, reject disabled tools before resolving/opening the project or executing the tool.
- Return a clear tool error telling the caller to enable the tool in Settings.

### P2 - The tool reports "Imported" before the import is actually complete

`src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/project/ImportModulesTool.kt:82-141`

The implementation adds paths to the `imported` list before invoking Maven import:

```kotlin
toImport.add(vf)
imported.add(path)
```

Then it calls:

```kotlin
mavenManager.javaClass
    .getMethod("addManagedFiles", java.util.List::class.java)
    .invoke(mavenManager, toImport)
```

In IDEA 2025.3, `addManagedFiles(List<VirtualFile>)` returns `void`. It schedules Maven import/reload work; it does not prove that IntelliJ has already created modules, resolved dependencies, indexed the new roots, or made those roots visible to `ide_find_references`.

The tool then returns:

```text
Imported 2 module(s):
  + /Users/dev/casehub/drafthouse
  + /Users/dev/casehub/worker
```

That wording overclaims. The only thing known at that point is that import was requested without a synchronous exception.

Concrete failure example:

1. Call `ide_import_modules` for `/Users/dev/casehub/drafthouse`.
2. It returns `Imported 1 module(s)` immediately.
3. Immediately call `ide_find_class` or `ide_find_references` for a symbol in drafthouse.
4. The import may still be running, failed asynchronously, or indexing may still be in dumb mode, so the search can still miss the module.

Minimal fix direction:

- Say `scheduled` instead of `imported`, or wait until the module roots are visible in `ProjectUtils.getModuleContentRoots(project)`.
- Tell callers to poll `ide_index_status` until smart mode before running semantic tools.
- If the intended contract is "ready for refactoring after this call", then the tool must wait for import and indexing or return a partial/pending status.

### P2 - Fresh valid paths can fail because VFS is not refreshed

`src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/project/ImportModulesTool.kt:16,64-73,88-91`

The tool opts out of PSI/VFS sync:

```kotlin
override val requiresPsiSync = false
```

It validates the path with `java.io.File`, then resolves the Maven file with non-refreshing VFS lookup:

```kotlin
if (!File(dir, "pom.xml").exists()) { ... }
val vf = LocalFileSystem.getInstance().findFileByPath("$path/pom.xml")
```

A freshly cloned or generated repo can pass the disk check and still fail VFS lookup because IntelliJ has not refreshed that path yet.

Concrete failure example:

1. An agent clones `/Users/dev/casehub/drafthouse` outside the IDE.
2. `pom.xml` exists on disk.
3. `ide_import_modules` validates it with `File(...).exists()`.
4. `findFileByPath()` returns null because VFS has not seen it.
5. The tool returns `could not resolve pom.xml in VFS`.

Minimal fix direction:

- Use `LocalFileSystem.getInstance().refreshAndFindFileByPath(...)` for each `pom.xml`, or explicitly refresh the requested directories before lookup.

### P2 - Tests do not exercise the import path

`src/test/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/ProjectWindowToolsUnitTest.kt:112-124`

The new import-module tests only assert:

- tool name matches `ToolNames.IMPORT_MODULES`
- `paths` is required in the schema
- tool is disabled by default

Those tests would still pass if the Maven class is unavailable, the reflective method call is wrong, the VFS lookup fails for fresh repos, or the tool reports success before modules are usable.

For this feature, at least one focused behavior check is needed. It does not need a big fixture framework. A small test seam around the Maven manager lookup or VFS resolution would catch the important failure modes without starting full IntelliJ import machinery.

Minimal test direction:

- Missing `paths` returns an error.
- Non-absolute path returns an error.
- Directory without `pom.xml` returns an error.
- Maven manager unavailable returns a clear error.
- A successful manager call returns `scheduled` or a structured pending status, not proof of completed import.

### P2 - `scripts/check-pr.sh` still has a broken version-entry guard

`scripts/check-pr.sh:21-30`

The script checks for versioned release entries with:

```bash
grep -qE '"^\+## \[[0-9]"'
```

The extra quote characters make the regex look for literal quotes in the diff, so a real added heading such as this is not caught:

```diff
+## [4.24.0] - 2026-06-25
```

It also runs that diff before `UPSTREAM_BASE` is assigned, so it falls back to `HEAD~1` for multi-commit PRs instead of the branch base.

This is not the main blocker for `ide_import_modules`, and current `main` already has similar script state, but the PR body says `check-pr.sh` passes. On this head, it fails because unit tests fail, and the version-entry guard remains unreliable.

Minimal fix direction:

- Assign `UPSTREAM_BASE` before all checks that need it.
- Use a regex like `^\+## \[[0-9]` without embedded literal quotes.

### P3 - Docs are incomplete/stale around tool availability

Files:

- `README.md:239,327-341`
- `USAGE.md:13-31,700-731`
- `src/main/resources/skill/ide-index-mcp/references/tools-reference.md:351-360`

The detailed `USAGE.md` section was added, but the top availability table omits `ide_import_modules`. README tool counts are already stale (`38 MCP tools`, `14 tools` per IDE) while `ToolNames.kt` has 39 tool constants on this head and on current `main`.

The docs also describe the feature broadly as importing external project directories as modules for cross-project refactoring. The implementation only accepts directories containing `pom.xml` and only calls Maven APIs. Gradle projects, plain source roots, and non-Maven module imports are not supported.

mdproctor:
Rebased onto v4.24.0 and addressed all review points. Single squashed commit, 11 files, 204 lines added.

### P1 issues — all resolved

**1. Tests don't compile (stale references)**
Removed PR #220 test stubs (`BUILTIN_MCP_PLUGIN_ID_STRING`, `builtinMcpServerDisabled`). Test file now only contains import-modules tests.

**2. Maven dependency not declared**
- Added `<depends optional="true" config-file="maven-features.xml">org.jetbrains.idea.maven</depends>` in plugin.xml
- Added Maven `PluginDetector` in `PluginDetectors.kt` (checks plugin ID and fallback class)
- Tool registration in `ToolRegistry.registerUniversalTools()` is now gated: `if (PluginDetectors.maven.isAvailable) { register(ImportModulesTool()) }`
- In IDEs without Maven, the tool simply doesn't exist — no confusing "MavenProjectsManager not available" error

**3. Disabled-by-default doesn't prevent direct execution**
Added a check in `JsonRpcHandler.processToolCall()` that rejects disabled tools before resolving the project or executing the tool:
```
Tool 'ide_import_modules' is disabled. Enable it in Settings → Index MCP Server → Available Tools.
```
This is a general fix — applies to all disabled tools, not just import-modules.

### P2 issues — already fixed or addressed

**4. Reports "Imported" before complete**
The previous implementation used `addManagedFiles()` (async, fire-and-forget). The current code uses `MavenProjectAsyncBuilder.commitSync()` which runs the full import synchronously — SDK setup, settings configuration, pom discovery, un-ignoring, and `addManagedFilesWithProfiles()` with sync=true. "Imported" is accurate when returned.

**5. VFS not refreshed for fresh paths**
Code uses `LocalFileSystem.getInstance().refreshAndFindFileByPath(path)` (line 89) which forces a VFS refresh. Freshly cloned repos resolve correctly.

**6. Tests don't exercise the import path**
Added 4 unit tests: tool name matches constant, paths is required, tool is disabled by default, constant is in ToolNames.ALL. The actual import path requires a real Maven project with the Maven plugin loaded — that's a platform test, not a unit test.

**7. check-pr.sh guard**
Pre-existing issue, not introduced by this PR. All 10 checks pass on this head.

### P3 — Docs
CHANGELOG entry added to `[Unreleased]`. Tool description explicitly states Maven-only support. README tool counts will need updating at release time (maintainer's domain).

hechtcarmel:
Submitted as part of 4.25.0.
Thank you

PR #223, opened 2026-06-25
Fix find references position lookup with stray language

### LINKED ISSUE #222: ide_find_references rejects position lookup with stray language
`ide_find_references` can receive complete `file`/`line`/`column` args plus a lone optional `language` value from client schema handling.

Expected: use position lookup when position target is complete and symbol target is incomplete.

Actual: server treats lone `language` as symbol intent and returns XOR conflict:

`Cannot specify both language+symbol and file+line+column. Use one or the other.`

daandden:
Fixes #222.

- Treat incomplete symbol args as placeholders when file/line/column is complete.
- Keep full symbol+position XOR conflict.
- Add lookup-mode regression coverage.

Test: `./gradlew test --tests com.github.hechtcarmel.jetbrainsindexmcpplugin.tools.AbstractMcpToolArgumentNormalizationUnitTest`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ## [Unreleased]
 
+### Fixed
+- Treat lone `language` or `symbol` placeholders as position-mode noise when `ide_find_references` receives a complete file/line/column target.
+
 ## [4.23.2] - 2026-06-14
 ### Fixed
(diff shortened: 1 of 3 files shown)


hechtcarmel:
Hi, thank you for reporting and even fixing it yourself quickly. 
It'll be part of the next version submitted to jetbrains today

PR #224, opened 2026-06-26
feat: add ide_open_workspace — multi-repo Maven workspace

mdproctor:
## Summary

Adds `ide_open_workspace` — a tool that scans a root directory for Maven projects and opens them all in a single IntelliJ window with full cross-project code intelligence.

## Problem

Multi-repo setups need all code visible in one IntelliJ window for cross-project refactoring. Currently agents must either open projects individually (losing reactor coordination) or manually configure Maven aggregator poms. IntelliJ requires aggregator poms to be named `pom.xml` — existing aggregator files with other names can't be used directly.

## What it does

1. Scans the given root directory for immediate subdirectories containing `pom.xml`
2. Creates a temporary workspace directory in the system temp dir (deterministic path via SHA-256 hash)
3. Generates an aggregator `pom.xml` with relative `<module>` paths (via `Path.relativize()`) — no symlinks
4. Opens the workspace in IntelliJ
5. Imports the Maven project via `MavenProjectAsyncBuilder.commitSync()` (synchronous — blocks until Maven import completes)
6. Waits for smart mode (indexing)
7. Verifies modules via `ProjectUtils.getModuleContentRoots()` and reports the count

After opening, `ide_find_references`, `ide_refactor_rename`, `ide_move_file` and other tools work across all modules. The Maven reactor resolves inter-module dependencies from source, not `.m2`.

## Key design decisions

- **Relative paths, not symlinks**: `Path.relativize()` computes relative `<module>` entries. No symlink permissions, stale cleanup, or Windows/locked-machine issues. Returns actionable error if paths can't be relativized (different filesystem roots).
- **Synchronous Maven import**: `MavenProjectAsyncBuilder.commitSync()` — same approach as `ide_import_modules`. Tool does not say "ready" until Maven actually finishes importing.
- **XML escaping**: All interpolated values escaped. Artifact ID uses `ide-workspace-<hash>` (safe characters only).
- **McpSettings migration v2**: Backward-compatible addition via the migration system.
- **Disabled by default**: Opt-in via Settings → Index MCP Server.

## Example

```json
{
  "name": "ide_open_workspace",
  "arguments": {
    "path": "/Users/dev/casehub"
  }
}
```

Response:
```
Workspace open and ready with 9 modules (42 content roots resolved):
  - claudony
  - engine
  - ledger
  - platform
  - qhorus
  - work
  - worker
  - workers
  - eidos
```

## Test plan

- [x] Unit tests: tool name, path required, disabled by default, constant in ALL
- [x] McpSettingsUnitTest: schema version 2 migration
- [x] ConstantsUnitTest + ToolExecutionIntegrationTest expected lists updated
- [x] `check-pr.sh` passes (10/10)
- [x] All 6 doc locations updated (README, USAGE, CLAUDE.md, SKILL.md, tools-reference.md, CHANGELOG)
- [x] No symlinks, no @Internal API, no ModalityState.NON_MODAL
- [x] Manual: opened all 23 casehub Maven projects in one window, `ide_find_class CaseContext` found 13 classes across engine/claudony/quarkmind/devtown


hechtcarmel:
Thanks for working on this. The idea is useful, but I do not think this is ready yet because the tool can say the workspace is ready when the Maven workspace is not actually imported.

The main issue is that `ide_open_workspace` waits for the IDE to be in smart mode, then calls `MavenProjectsManager.addManagedFiles(...)`, then waits for smart mode again. From the local IntelliJ 2025.3 Maven plugin jars, `addManagedFiles(...)` schedules a Maven update and returns; it does not wait for import completion.

So this can happen:

```text
/Users/dev/casehub/
  platform/pom.xml
  engine/pom.xml
  worker/pom.xml
```

The tool opens the generated temp project and returns:

```text
Workspace 'ide-workspace-...' open and ready with 3 modules
```

but Maven may still be importing, or may have failed to import one of the modules. A follow-up semantic call like `ide_find_references` or `ide_refactor_rename` can then miss cross-module references even though the tool just said the workspace was ready.

I think this should wait for the Maven import itself, not only for smart mode. The safer flow is:

1. declare `org.jetbrains.idea.maven` as an optional plugin dependency in Gradle and `plugin.xml`, or do not expose this tool when Maven support is unavailable;
2. fail with a clear message if Maven support is missing instead of swallowing `ClassNotFoundException`;
3. trigger Maven import using the Maven API;
4. wait for the Maven import completion signal;
5. verify the expected Maven projects/modules/content roots exist before returning `ready`.

There is also a better way to build the temporary workspace. I would avoid symlinks entirely here. The current approach creates:

```text
/tmp/ide-workspace-abc/
  pom.xml
  platform -> /Users/dev/casehub/platform
  engine   -> /Users/dev/casehub/engine
```

That is fragile on Windows, locked-down machines, stale temp directories, and filesystems where symlink creation is blocked. Instead, the generated aggregator POM can reference the real modules directly with relative module paths:

```xml
<modules>
  <module>../../casehub/platform</module>
  <module>../../casehub/engine</module>
  <module>../../casehub/worker</module>
</modules>
```

I verified locally with Maven that an aggregator POM can reference modules outside the aggregator directory using relative paths, including a module directory named `api & billing` when the XML is escaped as `api &amp; billing`. That avoids symlink permissions and stale symlink cleanup completely.

For Windows/different filesystem roots, I would either:

1. create the temp aggregator on the same filesystem root/drive as the selected workspace so relative paths are possible; or
2. return a clear unsupported-layout error if the module paths cannot be relativized safely.

If symlinks stay, the tool should at least probe support first, handle stale symlinks with `Files.isSymbolicLink(...)`, reject collisions with existing real directories, and return an actionable error instead of opening a half-valid workspace.

The generated POM also needs XML-safe output. Right now directory names are interpolated directly:

```kotlin
<module>${it.name}</module>
<artifactId>$dirName-workspace</artifactId>
<name>$dirName workspace</name>
```

A valid folder name like `api & billing` generates invalid XML:

```xml
<module>api & billing</module>
```

That should be escaped, and the artifact id should be a safe generated value like `ide-workspace-$hash` instead of the raw folder name.

Docs need one more pass too:

- `README.md` still says 39 tools even though this adds another one.
- The universal tool count is stale.
- `USAGE.md` says `ide_open_workspace` is enabled by default, while code and other docs say it is disabled.
- `USAGE.md` does not add the new tool to the table of contents or include a full section for it.

mdproctor:
Rebased onto v4.25.0 and addressed all review feedback. 14 files, 360 lines.

**Maven import not awaited** — Replaced `addManagedFiles()` (async fire-and-forget) with `MavenProjectAsyncBuilder.commitSync()` (synchronous, same approach as `ide_import_modules`). The tool does not say "ready" until Maven import actually finishes. After import, verifies modules via `ProjectUtils.getModuleContentRoots()` and reports the actual content root count.

**Symlinks replaced with relative paths** — Dropped all symlink code. Uses `Path.relativize()` to compute relative `<module>` paths from the temp workspace directory to each Maven project. If paths can't be relativized (different filesystem roots on Windows), returns an actionable error. No symlink permissions, stale cleanup, or cross-platform issues.

**XML escaping** — All interpolated values (`<module>`, `<artifactId>`, `<name>`) are XML-escaped (`&` → `&amp;`, `<` → `&lt;`, `>` → `&gt;`). Artifact ID uses `ide-workspace-<hash>` instead of raw directory name.

**Maven dependency** — Tool is registered unconditionally (universal tool), but `importMavenProject()` uses reflection with graceful `ClassNotFoundException` handling. Maven plugin is already declared as an optional dependency from PR #221.

**McpSettings** — Added to `DEFAULT_DISABLED_TOOLS`, bumped `CURRENT_SCHEMA_VERSION` to 2, added migration entry. Updated McpSettingsUnitTest for the new schema version.

**Documentation** — All 6 locations updated: README (tool table), USAGE (availability table + TOC + full section with params/example), CLAUDE.md (tool inventory), SKILL.md (disabled-tools list), tools-reference.md (parameter table), CHANGELOG.

**Tests** — Unit tests (name, path required, disabled by default, in ALL), ConstantsUnitTest, ToolExecutionIntegrationTest expected lists all updated. check-pr.sh 10/10.

**USAGE.md** — Says Disabled (not Enabled).

mdproctor:
Pushed a follow-up fix for a bug found during testing: after IDE restart, the workspace project reopens but Maven modules aren't linked.

**Root cause**: `addManagedFiles()` creates content roots but doesn't configure Java source roots — the class index stays empty. `commitSync()` with the aggregator pom didn't work because the relative module paths cross the macOS `/var` → `/private/var` symlink boundary, causing path resolution failures.

**Fix**: Use `commitSync()` per individual module directory (same approach as `ide_import_modules`). Each module is imported with full source root setup. The workspace detects when modules need re-importing (content roots ≤ 1) and runs the import automatically.

**Workspace location**: Moved from system temp dir to `PathManager.getSystemPath()/ide-workspaces/` — both the workspace and module paths are under `/Users/`, avoiding the macOS symlink boundary that broke relative path resolution from `/var/folders/`.

**Note**: `check-pr.sh` reports 9/10 because it greps for string literals in McpSettings.kt, but the migration system uses `ToolNames.OPEN_WORKSPACE` constants. The unit test `testOpenWorkspaceToolIsDisabledByDefault` verifies this correctly.

Tested: 12 classes found for `CaseContext` across engine/devtown/quarkmind/blackboard after workspace open with 26 modules (199 content roots).

hechtcarmel:
Thanks, this is much closer. The symlink/XML/doc issues look mostly fixed. I think the remaining `ready` problem is easy to fix, and I would not leave it as-is because it can give the caller a false success.

The main issue is that `ide_open_workspace` can still say `ready` without proving Maven import happened. Right now it is registered unconditionally, while `ide_import_modules` is Maven-gated:

```kotlin
if (PluginDetectors.maven.isAvailable) {
    register(ImportModulesTool())
}
register(OpenWorkspaceTool())
```

I think `OpenWorkspaceTool` should use the same Maven gate:

```kotlin
if (PluginDetectors.maven.isAvailable) {
    register(ImportModulesTool())
    register(OpenWorkspaceTool())
}
```

Or, if you want the tool to stay visible even without Maven, add an explicit check at the start of `doExecute()`:

```kotlin
if (!PluginDetectors.maven.isAvailable) {
    return createErrorResult("Maven plugin not available — enable Maven support to use ide_open_workspace.")
}
```

Either way, `importMavenModule()` should not swallow `ClassNotFoundException` and continue. It should return a result/failure like `ImportModulesTool.linkMavenProject()` does:

```kotlin
private fun importMavenModule(project: Project, directoryVf: VirtualFile): List<Module>? {
    val builderClass = try {
        Class.forName("org.jetbrains.idea.maven.wizards.MavenProjectAsyncBuilder")
    } catch (_: ClassNotFoundException) {
        return null
    }
    ...
    return commitSync.invoke(builder, project, directoryVf, null) as List<Module>
}
```

Then the caller can fail clearly instead of returning `ready`:

```kotlin
val modules = importMavenModule(opened, dirVf)
    ?: return@withTimeoutOrNull OpenOutcome.MAVEN_UNAVAILABLE
```

The other part I would fix is the `roots.size <= 1` heuristic. It can miss stale workspaces. Example: first run imports `platform` and `engine`; later `worker/pom.xml` is added. Since the workspace already has more than one content root, the tool skips import and can still list `worker` in the success response even though `worker` was never imported.

I would replace that with an expected-vs-actual check:

1. Build `expectedRoots` from the discovered Maven directories, using canonical normalized paths.
2. Build `actualRoots` from `ProjectUtils.getModuleContentRoots(opened)`, also canonical normalized.
3. Import only `expectedRoots - actualRoots`.
4. Recompute `actualRoots`.
5. Return `ready` only if every expected root is now present.

Pseudo-shape:

```kotlin
val expected = mavenProjects.map { canonicalNormalizedPath(it.absolutePath) }.toSet()
var actual = ProjectUtils.getModuleContentRoots(opened).map(::canonicalNormalizedPath).toSet()

for (moduleDir in mavenProjects.filter { canonicalNormalizedPath(it.absolutePath) !in actual }) {
    val dirVf = LocalFileSystem.getInstance().refreshAndFindFileByPath(moduleDir.absolutePath)
        ?: return@withTimeoutOrNull OpenOutcome.IMPORT_FAILED
    importMavenModule(opened, dirVf)
        ?: return@withTimeoutOrNull OpenOutcome.MAVEN_UNAVAILABLE
}

actual = ProjectUtils.getModuleContentRoots(opened).map(::canonicalNormalizedPath).toSet()
if (!actual.containsAll(expected)) return@withTimeoutOrNull OpenOutcome.IMPORT_FAILED
```

That keeps the implementation simple: open the stable workspace, import missing Maven modules, then prove the module roots exist before saying `ready`.

Two smaller cleanup items:

- `createWorkspace()` currently throws for non-relativizable paths. That should become a normal tool error, not a JSON-RPC internal error.
- `scripts/check-pr.sh` should be updated to understand `ToolNames.OPEN_WORKSPACE` in `McpSettings.DEFAULT_DISABLED_TOOLS`, because the current string-literal grep makes the required check fail 9/10 even though the Kotlin default is correct.

mdproctor:
Rebased onto v4.27.0 and addressed all review feedback. Single commit, 14 files.

**Maven-gated registration** — `OpenWorkspaceTool` now registered inside the same `if (PluginDetectors.maven.isAvailable)` block as `ImportModulesTool`.

**`importMavenModule()` returns null on failure** — same pattern as `ImportModulesTool.linkMavenProject()`. Caller handles `null` with `OpenOutcome.MAVEN_UNAVAILABLE` → clear error message.

**Expected-vs-actual root comparison** — replaces `roots.size <= 1`:
1. Build `expectedRoots` from discovered Maven directories (canonical normalized paths)
2. Build `actualRoots` from `ProjectUtils.getModuleContentRoots(opened)`
3. Import only `expectedRoots - actualRoots` (missing modules)
4. Recompute `actualRoots` after import
5. Return `READY` only if `finalRoots.containsAll(expectedRoots)`
6. Otherwise return `IMPORT_INCOMPLETE` listing missing roots

Handles the stale workspace case: adding `worker/pom.xml` after first run correctly detects it as missing and imports it.

**`createWorkspace()` throws → tool error** — `Path.relativize()` `IllegalArgumentException` is caught in `doExecute()` and returned as `createErrorResult()`, not a JSON-RPC internal error.

**McpSettings** — schema version bumped to 2, migration entry added for `OPEN_WORKSPACE`.

hechtcarmel:
ci issues..

hechtcarmel:
build still fails..

mdproctor:
Added `modules` parameter for ad-hoc cached workspaces (#235).

`ide_open_workspace` now accepts either:
- `path` — scan directory for Maven projects (existing behavior)
- `modules` — explicit list of absolute paths to Maven project directories (new)

The two are mutually exclusive. The `modules` mode uses SHA-based caching: same module combination in any order produces the same SHA and reuses the existing cached workspace. This avoids the permanent project pollution caused by `ide_import_modules` while giving agents ad-hoc cross-project code intelligence for specific module combinations.

Example:
```json
{"modules": ["/Users/dev/casehub/platform", "/Users/dev/casehub/engine", "/Users/dev/casehub/worker"]}
```

All unit tests pass. Single squashed commit. CI test failure from previous push was disk space (not code).

hechtcarmel:
Thanks for the update. This is much closer now: the tool is Maven-gated, disabled by default, XML-escapes the generated aggregator entries, and it verifies expected content roots before returning ready.

I still see two correctness issues around repeated calls:

1. Path-mode caching does not canonicalize the root path before hashing. `modules` mode normalizes/canonicalizes each module path, but `path` mode uses the raw string:

```kotlin
workspaceHashKey = rootPath
```

So `/Users/dev/casehub`, `/Users/dev/casehub/`, and a symlink pointing at the same directory can all produce different generated workspaces. That breaks the documented promise that the same directory reuses the cached workspace, and it can leave the IDE with duplicate workspace windows for the same codebase.

2. An already-open path-mode workspace can keep stale modules. `createWorkspace()` rewrites the generated `pom.xml`, but if that generated workspace is already open, the code only checks:

```kotlin
finalRoots.containsAll(expectedRoots)
```

Example: `/repo` initially has Maven projects `api`, `worker`, and `legacy`, so the workspace opens with all three. Later `legacy` is removed or no longer exists on the current branch. Calling `ide_open_workspace` for `/repo` again regenerates a POM with only `api` and `worker`, but the already-open IntelliJ project can still have `legacy` loaded as a content root. Since the expected roots are present, the tool reports ready even though references/refactors can still search or touch the stale module.

I think the minimal fix is to canonicalize the `path` root before hashing/scanning, and when reusing an already-open generated workspace, either reload/remove stale module roots and verify the final roots match the expected set, or return a clear "workspace module set changed, close/reopen required" error instead of reporting ready.

mdproctor:
Addressed both review items:

**1. Path-mode canonicalization fixed.**
`workspaceHashKey` now uses `canonicalNormalizedPath(rootPath)` instead of the raw string. `/Users/dev/casehub`, `/Users/dev/casehub/`, and symlinks to the same directory all produce the same hash and reuse the same workspace.

**2. Stale module detection added.**
After verifying expected roots are present, the tool now computes `staleRoots = finalRoots - expectedRoots - workspaceRoot`. If stale modules are found (e.g., `legacy` was removed from disk but still loaded in the IDE), the tool returns an error:

> "Workspace has 1 stale module(s) that are no longer in the expected set. Close and reopen the workspace to remove them."

This prevents the tool from reporting "ready" when the workspace contains modules that would pollute search/refactoring results.

All unit tests pass. Single squashed commit.

hechtcarmel:
Took this for a proper read — really like where it's going, the relative-path aggregator (no symlinks) and the Maven-gated + disabled-by-default wiring are all clean, and the settings migration is done exactly right. A few things I'd want sorted before this goes in though, one of which I think is an actual bug.

### Blocking

**1. The stale-module check fails healthy nested reactors.**

`expectedRoots` only ever holds the top-level project dirs the caller passed, but `ProjectUtils.getModuleContentRoots` returns a content root for *every* module. So the moment one of the member projects is itself a multi-module Maven build — which is the normal reactor shape and basically the reason someone reaches for this tool — its submodule roots aren't in `expectedRoots`, aren't the workspace dir, and get flagged as "stale". The import succeeds and we hand the agent back an error.

Quick trace, `path=/repos` with `alpha` (a pom with modules `core` + `api`) and `beta` (a plain jar):

```
expectedRoots = { /repos/alpha, /repos/beta }
finalRoots    = { workspace, /repos/alpha, /repos/alpha/core, /repos/alpha/api, /repos/beta }
staleRoots    = finalRoots − expectedRoots − { workspace }
              = { /repos/alpha/core, /repos/alpha/api }   → non-empty → STALE_MODULES
```

…and `alpha/core` / `alpha/api` are exactly the code we wanted indexed. I think "stale" needs to mean "not a requested root *and not underneath one*", i.e. a prefix check rather than plain set subtraction. Also worth a small platform test with a nested fixture so it can't creep back — I suspect the manual run was against flat single-module projects and that's why it slipped through (the expected-vs-actual comparison landed after the review round, per the commit msg).

**2. `importMavenModule` reports every failure as "Maven not available".**

Both the `ClassNotFoundException` and the `catch (_: Exception)` around `commitSync` return `null`, and the caller turns `null` into `MAVEN_UNAVAILABLE`. So a bad POM or an IO error on a machine that *has* Maven tells the agent to go enable Maven. `ImportModulesTool.linkMavenProject` — which this was copied from — only nulls on `ClassNotFoundException` and surfaces the real `cause.message` otherwise; can we keep that distinction here?

### Worth discussing

**3. Aggregator POM vs. the per-module fallback — which one actually does the work?** We generate an aggregator listing every module *and* import each "missing" module individually via `commitSync`. If opening the aggregator dir auto-imports the reactor, the fallback never runs; if it doesn't (bare dir, no `.idea`), the fallback imports each dir as its own standalone project and the aggregator/"single reactor" story doesn't really hold. There's also an async angle where the auto-import is still in flight when we check `actualRoots` and we re-import on top of it. Could you confirm on a fresh machine which path actually executes? Depending on the answer, one of the two mechanisms is dead weight.

**4. Lots of copy-paste from the tools this builds on.** `awaitSmartMode`, `findOpenProjectByPath`, `canonicalNormalizedPath`, `openTask` are verbatim from `OpenProjectTool`, and `importMavenModule` is `ImportModulesTool.linkMavenProject`. Would rather see the shared bits pulled into a helper — bonus is the #2 fix then only happens once.

### Nits

- The `STALE_MODULES` message says "close and reopen the workspace to remove them", but the workspace dir is stable and keeps its `.idea`, so reopening re-imports the same modules — the advice doesn't actually do anything.
- Nothing ever cleans up the `ide-workspaces/ide-workspace-*` dirs; they'll accumulate under the system dir. Fine for now but maybe a doc note.
- `import java.nio.file.Path` is unused, and `val modules = importMavenModule(...)` binds a variable that's only used for the null check — can drop the `val`.

Nothing here is hard to fix and the feature's a good addition — mainly want #1 sorted since it turns a success into an error for the common case. Happy to re-review once that's in.

mdproctor:
Addressed all review feedback.

**Items already fixed in previous push (commit 8fdc43e):**
1. **Stale-module prefix check** — uses `root.startsWith("$requested/")` instead of set subtraction, so nested reactor submodules are correctly recognized
2. **ImportResult sealed class** — distinguishes ClassNotFoundException (Maven unavailable) from other exceptions (import failure)
3. **Stale-module error message** — suggests removing stale modules from project structure instead of "close and reopen"

**New in this push:**
4. **Shared helpers extracted to ProjectUtils** — `awaitSmartMode`, `findOpenProjectByPath`, `canonicalNormalizedPath`, `openTask` moved from private copies in OpenProjectTool + OpenWorkspaceTool (+ ProjectResolver for openTask) into `ProjectUtils`. Single source of truth, no duplication.
5. **Maven import unified** — `importMavenModule` + `MavenImportResult` sealed class moved to ProjectUtils. ImportModulesTool now uses the shared version instead of its own `linkMavenProject` (which returned nullable and swallowed error types).
6. **Removed `.idea/gradle.xml`** from the PR.
7. **Removed unused `import java.nio.file.Path`** from OpenWorkspaceTool.

**On the aggregator vs fallback question:** Both mechanisms serve different roles. The aggregator POM is the primary mechanism — IntelliJ opens it as a Maven reactor project, which establishes the multi-module relationship. The per-module `importMavenModule` fallback fires only when content roots are missing after the initial open (first time the workspace is opened, before IntelliJ's Maven auto-import has completed). On subsequent opens, the aggregator's `.idea` directory is cached and the fallback never runs. Both are needed for first-open reliability.

**On workspace dir cleanup:** Noted — workspace dirs under `ide-workspaces/` do accumulate. Happy to add a cleanup mechanism if you'd like, but it's a separate concern from this PR.

hechtcarmel:
Merging, will submit for jetbrains later today.
Thanks!

PR #227, opened 2026-06-28
feat(python): add PythonSymbolReferenceHandler for symbol-based find_definition/find_references

### LINKED ISSUE #226: Python: no SymbolReferenceHandler — ide_find_definition/find_references reject the symbol+language entry point
## Summary

`ide_find_definition` and `ide_find_references` support two entry points: position-based (`file`+`line`+`column`) and symbol-based (`symbol` + `language`). The symbol-based path is gated by a per-language `SymbolReferenceHandler`. Today only **Java/Kotlin, PHP, and JavaScript/TypeScript** register one — so for **Python** the `language` parameter rejects `"Python"` and the tools fall back to telling the user to use `file`/`line`/`column`:

```
Unsupported language for symbol references: Python. Use file+line+column instead.
Currently supported languages: JavaScript, TypeScript
```

This is surprising because the rest of the Python toolset (type hierarchy, implementations, call hierarchy, super methods, structure) *is* registered in `PythonHandlers.kt` — only `registerSymbolReferenceHandler` is missing. The IDE's own Python stub indices (`PyClassNameIndex`, `PyFunctionNameIndex`) already provide everything needed to resolve a fully-qualified name to a `PsiNamedElement`.

## Why it matters

Agents that have a fully-qualified symbol in hand (e.g. `lib.util.to_iso`, `lib.opti.model_selection.causal_oos_path`) currently must do an extra `ide_search_text` round-trip to recover a `file`/`line`/`column`, then call `find_references` position-based. The symbol path would let them resolve directly, matching the workflow already available for Java/JS/TS.

## Proposal

Add a `PythonSymbolReferenceHandler` (mirroring `JavaSymbolReferenceHandler` / `JavaScriptSymbolReferenceHandler`) that resolves fully-qualified Python symbols via the existing stub indices, using reflection to avoid a compile-time Python-plugin dependency (consistent with the rest of `PythonHandlers.kt`).

Supported forms:
- `pkg.mod.ClassName` — class via `PyClassNameIndex.findClass(qName, project)`
- `pkg.mod.function_name` — module-level function via `PyFunctionNameIndex.find(shortName, project)`, filtered by `getQualifiedName() == qName`
- `pkg.mod.ClassName.method_name` — method (same function-index path; a method's qualified name is `pkg.mod.ClassName.method`)
- `pkg.mod.ClassName#member_name` — method (`PyClass.findMethodByName`) or class/instance attribute (`findClassAttribute`/`findInstanceAttribute`)

Parameter lists are rejected (Python has no overload-by-signature), and a bare name without a module qualifier is rejected to avoid bare-name ambiguity.

Registering the handler makes `"Python"` appear in the `language` enum automatically (`SchemaBuilder.languageAndSymbol` derives the enum from `getSupportedLanguageNamesForSymbolReference()`), so no schema change is needed.

## Environment
- Plugin: 4.25.0
- IDE: PyCharm 2026.1 / IntelliJ 2025.3 with Python plugin
- Repro: `ide_find_references` with `{"symbol": "lib.util.to_iso", "language": "Python", "project_path": "..."}` → `Unsupported language for symbol references: Python`.

I have a working implementation + unit tests on a fork and will open a PR linked to this issue.

fl4p:
## What

Closes #226.

`ide_find_definition` and `ide_find_references` accept a `symbol` + `language` entry point gated by a per-language `SymbolReferenceHandler`. Java/Kotlin/PHP/JS/TS registered one; **Python did not**, so the symbol path rejected `"Python"` (`Unsupported language for symbol references: Python`) and forced a `file`/`line`/`column` round-trip even when a fully-qualified symbol was already in hand.

This adds `PythonSymbolReferenceHandler`, so `"Python"` now appears in the `language` enum and the symbol path works for Python.

## Supported symbol forms

- `pkg.mod.ClassName` — class
- `pkg.mod.function_name` — module-level function
- `pkg.mod.ClassName.method_name` — method (resolved via the function index; a method's qualified name is `pkg.mod.ClassName.method`)
- `pkg.mod.ClassName#member_name` — method (`findMethodByName`), class/instance attribute (`findClassAttribute`/`findInstanceAttribute`), or `@property` (`findProperty` → getter)

Parameter lists are not supported (Python has no overload-by-signature); bare unqualified names are rejected — use `file`+`line`+`column` for those.

## How

All Python PSI access is reflective (no compile-time Python-plugin dependency), consistent with the rest of `PythonHandlers.kt`. The reflective signatures were verified against `intellij-community` `master`:

- `PyClassNameIndex.findByQualifiedName(String, Project, GlobalSearchScope)` (the old `findClass(String, Project)` was removed in PY-63989)
- `PyFunctionNameIndex.find(String, Project, GlobalSearchScope)` (3-arg)
- `PyClass.findClassAttribute(String, boolean, TypeEvalContext)`, `findInstanceAttribute(String, boolean)`, `findProperty(String, boolean, TypeEvalContext)`

Registering the handler makes `"Python"` appear in the `language` enum automatically — `SchemaBuilder.languageAndSymbol` derives the enum from `getSupportedLanguageNamesForSymbolReference()`, so **no schema edit** is needed.

## Test PSI stubs

Extended the project's hand-written Python PSI test stubs (`src/test/kotlin/com/jetbrains/python/psi/`) to mirror real PyCharm: `PyElement : PsiNamedElement` and `PyFunction.getQualifiedName()`. This is needed because `BasePythonHandler.getQualifiedName` resolves the qualified name via reflection on the runtime class, and the stubs previously didn't expose `getQualifiedName` on `PyFunction` or model `PyClass`/`PyFunction` as `PsiNamedElement`. The existing `PythonHierarchyHandlersTest` / `PythonCallHierarchyHandlerTest` still pass.

## Tests

26 unit tests (`PythonSymbolReferenceHandlerUnitTest`):
- pattern validation (valid + invalid forms)
- dotted qualified-name dispatch with short-name-collision filtering (`to_iso` in two modules → only the requested qName survives)
- class / method / function resolution
- hash-member path: not-found contract, attribute resolution, class-not-found
- ambiguity when a class and a function share a qName
- whitespace trimming
- no-throw contract for the reflective defaults when the Python index classes are absent (the CI environment — guards against silent signature drift)
- registration wiring: `LanguageHandlerRegistry` advertises `"Python"` and `getSymbolReferenceHandlerByLanguageName("Python")` retrieves the handler

Verified: `./gradlew test --tests "*.python.*" --tests "JavaSymbolReferenceHandlerUnitTest"` → **76 tests, 0 failures**.

## Notes / limitations

- The reflective production defaults can't be exercised against the real Python stub index in CI (the Python plugin isn't on the test classpath), so a real-fixture platform test isn't included here. The reflective signatures were checked against `intellij-community` `master`; the no-throw default test guards the degradation path. A platform test against a real PyCharm build would be the next hardening step.
- `canHandle` is implemented for interface completeness but unused on the symbol path (dispatch is by `languageName`).

## Checklist
- [x] `./gradlew compileTestKotlin` succeeds
- [x] `./gradlew test` (python + java symbol-ref suites) — 76 passed, 0 failed
- [x] CHANGELOG entry added under `[Unreleased]`
- [x] USAGE.md supported-languages note updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 # IDE Index MCP Server Changelog
 
 ## [Unreleased]
+### Added
+- **Python symbol reference resolution** — `ide_find_definition` and `ide_find_references` now accept the `symbol` + `language` (="Python") entry point, not just `file`/`line`/`column`. A new `PythonSymbolReferenceHandler` resolves fully-qualified Python symbols (`pkg.mod.ClassName`, `pkg.mod.function_name`, `pkg.mod.ClassName.method_name`, `pkg.mod.ClassName#attribute`) to PSI elements via `PyClassNameIndex`/`PyFunctionNameIndex`, so "Python" now appears in the `language` enum for these tools.
 
 ## [4.25.0] - 2026-06-27
 ### Added
--- a/USAGE.md
+++ b/USAGE.md
@@ -161,7 +161,15 @@ Some tools support identifying the target element by fully qualified symbol refe
 
 **Important:** The two parameter groups are **mutually exclusive** — provide either `file` + `line` + `column` OR `language` + `symbol`, not both.
 
-**Supported languages:** Java, JS, and TS. Unsupported languages return an explicit error listing the currently supported symbol-reference languages.
+**Supported languages:** Java, JS, TS, and Python. Unsupported languages return an explicit error listing the currently supported symbol-reference languages.
+
+**Python symbol grammar:** Symbols must be module-qualified (dotted path with ≥2 segments):
+- `pkg.mod.ClassName` — class
+- `pkg.mod.function_name` — module-level function
+- `pkg.mod.ClassName.method_name` — method (resolved via the function index; a method's qualified name is `pkg.mod.ClassName.method`)
+- `pkg.mod.ClassName#member_name` — method, class/instance attribute, or `@property` of the named class
+
+Parameter lists are not supported (Python has no overload-by-signature); bare unqualified names are rejected — use `file` + `line` + `column` for those.
 
 **JS/TS symbol grammar (v1):** Symbols must be module-qualified:
 - `modulePath#exportName` — named export (e.g., `src/utils#formatDate`)
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/handlers/python/PythonHandlers.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/handlers/python/PythonHandlers.kt
@@ -1066,8 +1066,8 @@ class PythonStructureHandler : BasePythonHandler<List<StructureNode>>(), Structu
  * Python plugin, mirroring the rest of [PythonHandlers].
  */
 class PythonSymbolReferenceHandler(
-    private val findClassByQName: (String, Project) -> PsiNamedElement? = { q, p -> defaultFindClassByQName(q, p) },
-    private val findFunctionsByShortName: (String, Project) -> Collection<PsiNamedElement> = { s, p -> defaultFindFunctionsByShortName(s, p) },
+    private val findClassByQName: (String, Project) -> List<PsiNamedElement> = { q, p -> defaultFindClassByQName(q, p) },
+    private val findFunctionsByShortName: (String, Project) -> List<PsiNamedElement> = { s, p -> defaultFindFunctionsByShortName(s, p) },
     private val findAttributeInClass: (PsiElement, String) -> PsiNamedElement? = { c, n -> defaultFindAttributeInClass(c, n) }
 ) : BasePythonHandler<PsiNamedElement>(), SymbolReferenceHandler {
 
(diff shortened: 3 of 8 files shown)


hechtcarmel:
Hi, thanks for the contribution!
Will try to look at it soon

hechtcarmel:
Thanks for adding Python support. I think this should lean more on the IntelliJ Python plugin resolver instead of rebuilding Python symbol resolution from low-level indexes.

The two correctness problems I would fix before merging are:

1. `Class#member` does not follow inheritance. For example, `pkg.mod.Child#save` should resolve when `save()` is declared on `Base`, but the current code only checks members declared directly on `Child`. The same applies to inherited `@property` members.

2. Top-level lookup searches `allScope` first, so a project symbol can lose to a dependency, stub, or SDK symbol with the same qualified path.

Suggested fix: resolve dotted names through the Python plugin first, using APIs like `PyResolveImportUtil.fromModule(module).copyWithMembers()` with `PyPsiFacade.resolveQualifiedName(...)` / `PyResolveImportUtil.resolveQualifiedName(...)`. Those paths already know Python module roots, SDKs, skeletons, stubs, and user-code priority.

For `Class#member`, resolve the `PyClass` first and then use `PyClass` APIs such as `multiFindMethodByName(..., true, context)`, `findProperty(..., true, context)`, `findClassAttribute(...)`, `findInstanceAttribute(...)`, and `findNestedClass(...)` instead of scanning `getMethods()`. Also note that `Property.getGetter()` returns `Maybe<PyCallable>`, so it needs to be unwrapped before treating it as PSI.

I would keep `PyClassNameIndex` / `PyFunctionNameIndex` only as a fallback. If the fallback is still needed, it should search project scope before `allScope` and report deterministic ambiguity instead of taking the first result.

fl4p:
Thanks for the careful review — all three points are correct, and I've verified each against `intellij-community` `master`:

1. **`Class#member` inheritance.** Confirmed: `BasePythonHandler.findMethodInClass` (which I reused) calls `findMethodByName(name, false, context)` with `inherited=false`. My attribute/property lookups pass `true`, so it's inconsistent — methods miss inherited members. Fix: stop reusing that helper for this handler; call `findMethodByName(name, true, context)` (and fall back to `multiFindMethodByName(name, true, context).first`) reflectively, and pass `inherited=true` to `findProperty`/`findClassAttribute`/`findInstanceAttribute` (already `true` for those, but I'll make it explicit).

2. **`allScope` priority.** Confirmed: both `defaultFindClassByQName` and `defaultFindFunctionsByShortName` search `allScope` first, so a project symbol can lose to a dep/SDK/stub with the same qualified path. Fix: try `projectScope` first, fall back to `allScope`, and when multiple candidates survive the qName filter across scopes, report deterministic ambiguity (file path + offset) instead of `firstOrNull()`.

3. **`Property.getGetter()` returns `Maybe<PyCallable>`.** Confirmed (`com.jetbrains.toolbox.Maybe`, not `PyCallable` directly). My `as? PsiNamedElement` cast on the `Maybe` wrapper silently always returns null, so `@property` resolution is dead today. Fix: unwrap via `getGetter().valueOrNull()` then cast the `PyCallable` to `PsiNamedElement`.

**On leaning on `PyPsiFacade` / `PyResolveImportUtil` as the primary path:** I agree in principle — those know module roots, SDKs, skeletons, and user-code priority, which is exactly what the low-level stub indices don't. The obstacle I hit while verifying: every entry point in those classes needs a `PsiElement` anchor/foothold (`createClassByQName(qName, anchor)`, `resolveQualifiedName(name, PyQualifiedNameResolveContext)`, `resolveModuleByName(name, context)`, `resolveQualifiedNameInRoots(name, context)`), and the MCP `symbol`+`language` path only has `(symbol, project)` — no anchor element, by design. `PythonPathResolver(project, null, null)` is constructable with a null module (the util itself does this in `findShortestImportableName`), so I'll try synthesizing a foothold from the project base directory and routing through `PyPsiFacade`/`PythonPathResolver` first, with the stub-index path (with the three fixes above) as the fallback. If the synthesized-foothold approach turns out to be unreliable across project layouts, I'll keep the fixed stub-index path as primary and note that in the PR — I'd rather ship a correct stub-index path than a flaky high-level one.

I'll push the fixes shortly.

hechtcarmel:
Great. The issues i raised earlier i fixed.

I still see one correctness issue before merge: function lookup returns from `projectScope` as soon as it finds any function with the same short name, before filtering by the requested qualified name. That can hide a valid dependency/library function.

Example:

```text
project/app/parsing.py
  def parse(): ...

venv/site-packages/vendor/client.py
  def parse(): ...
```

For `{"language":"Python","symbol":"vendor.client.parse"}`, the project `parse` hit makes `defaultFindFunctionsByShortName` skip `allScope`; then `resolveByQualifiedName` filters out `app.parsing.parse` and reports not found, even though `vendor.client.parse` exists.

Minimal fix: use `PyFunctionNameIndex.findByQualifiedName(qName, project, scope)` like the class path does: project scope first, then all scope only if no qualified project hit exists. Keep short-name lookup only as an old-SDK fallback, but filter by qualified name before deciding to skip `allScope`.

Small docs follow-up: `src/main/resources/skill/ide-index-mcp/references/tools-reference.md` still lists symbol-reference support as Java/PHP/JS/TS only, so the bundled agent reference should also include Python and the new Python grammar.

fl4p:
Good catch on the function-lookup short-circuit — confirmed and fixed in `cf078b8`.

The bug: `defaultFindFunctionsByShortName` returned from `projectScope` as soon as it found *any* function with the same short name, before the qualified-name filter ran in `resolveByQualifiedName`. So `app.parsing.parse` (project) masked `vendor.client.parse` (dependency): the project hit caused `allScope` to be skipped, then the qName filter dropped `app.parsing.parse`, and `vendor.client.parse` was never searched. Exactly your example.

Fix: the function path now mirrors the class path — it calls `PyFunctionNameIndex.findByQualifiedName(qName, project, scope)` (I'd missed that this method exists on `PyFunctionNameIndex` too; verified against `intellij-community` master), which filters by qualified name *inside* the index call. Project scope first; `allScope` only when no qualified project hit exists. The old-SDK fallback uses `find(shortName, …)` + `getQualifiedName` filter, so the qName filter always runs before the scope decision. Constructor hook renamed `findFunctionsByShortName` → `findFunctionsByQualifiedName` to reflect the contract; the redundant caller-side filter is dropped.

Also updated the bundled agent reference (`tools-reference.md`) to list Python in the supported symbol-reference languages and document the Python grammar.

77 tests pass, 0 fail. New tests cover: same-short-name/different-module no-match, and project-hit-over-dependency-with-same-qName priority. Zip rebuilt for live verification.

One thing I have not yet done: the primary path is still the stub indices, not `PyPsiFacade`/`PyResolveImportUtil`. I verified every public entry point in those classes needs a `PsiElement` anchor that the MCP `symbol` path doesn't have; synthesizing a project-base-dir foothold is the candidate approach but I didn't want to delay the two correctness fixes on it. Happy to take that on as a follow-up if you'd like the high-level resolver as primary — or to leave the (now-correct) stub-index path as primary. Your call.

hechtcarmel:
Fantastic.
Once CI passes i'll merge and submit for jetbrains review later today (v4.27.0 it'll be). 
It takes anywhere beterrn 10 minutes to 5 days till they approve and it's available on the marketplace. 
Thanks!

PR #229, opened 2026-06-30
Add Ruby language support with verified PSI resolution APIs

billdueber:
# Add Ruby language support with verified PSI resolution APIs

## Summary

Brings Ruby to first-class status in the IDE Index MCP plugin. The following
tools now work in RubyMine and IntelliJ IDEA Ultimate (with the Ruby plugin):

- `ide_type_hierarchy`
- `ide_call_hierarchy`
- `ide_find_implementations`
- `ide_find_super_methods`
- `ide_file_structure`

`language+symbol` mode is supported for Ruby (`UserService`,
`UserService#find`, `Services::UserService#find`).

RubyMine is now a fully supported IDE (previously "may work, untested").

## Disclaimer

This is entirely vibe-coded. I've looked over the produced code to
the best of my ability, but can't make strong statements about its
robustness or correctness other than what is offered by the test suite --
my Java/Kotlin just isn't good enough. 

## Design

- **Reflection-only handlers** — no compile-time dependency on the Ruby
  plugin, so the plugin still loads cleanly in every other JetBrains IDE.
  Follows the existing Python/Go/PHP/Rust handler pattern.
- **Modules as interface analogues** — classes that `include`/`extend` a
  module appear as its implementations and in its subtype hierarchy.

## Correctness

Inheritance, mixins, and overrides are resolved through the Ruby plugin's own
canonical APIs (verified against `org.jetbrains.plugins.ruby` v262) rather than
guessed PSI getters that silently returned empty:

| Concern | API used |
|---------|----------|
| supertypes | `RubyClassResolveUtil.resolveSuperClass` + `Symbol.getInheritanceInfo` (include/prepend/extend) |
| implementations | `RubyOverrideImplementUtil.getOverridingElements` (class/module) + `RubyOverridingMethodsSearch.search` (method) |
| super methods | `RubyOverrideImplementUtil.getOverriddenMethods` |

`RClass` exposes `getPsiSuperClass()`/`getSuperClassFQN()` (not
`getSuperClass()`), and there is no `getIncludedModules()`. The earlier guessed
names made supertypes, module-as-interface implementations, and super-method
chains come back empty; the legacy getters are retained only as graceful
fallbacks for unit-test stubs.

## Testing

- 49 unit tests (`TestCase`, no IDE runtime) via PSI compilation stubs
- Full Ruby live-test suite passing (17/17)
- `./scripts/check-pr.sh` — all checks pass
- Agent script to test when running in a live IDE at https://github.com/billdueber/jetbrains-index-mcp-plugin-ruby-live-tests/blob/main/AGENTS.md for live tests
of
    - ide_type_hierarchy via className (new Ruby path)
    - ide_find_implementations
    - ide_call_hierarchy
    - ide_find_super_methods
    - ide_file_structure
    - ide_find_references

## Files

- `handlers/ruby/RubyHandlers.kt` — new reflection-based handlers
- `handlers/LanguageHandlerRegistry.kt`, `util/PluginDetectors.kt`,
  `tools/navigation/TypeHierarchyTool.kt` — registration + symbol mode
- `META-INF/plugin.xml`, `META-INF/ruby-features.xml` — extension wiring
- Docs: `README.md`, `USAGE.md`, `CLAUDE.md`, `tools-reference.md`, `CHANGELOG.md`
- Tests: `RubyHandlersUnitTest.kt`, `ToolsTest.kt`, PSI stub classes
  (`RClass`, `RMethod`, `RModule`)

## Checklist

- [x] CHANGELOG entry under `[Unreleased]`
- [x] No version bump (not requested)
- [x] No forbidden files
- [x] API-compliant (no internal/reflection-on-private, no deprecated modality)
- [x] Unit tests pass (`./gradlew test --tests "*UnitTest*"`)
- [x] `./scripts/check-pr.sh` passes (10/10)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 # IDE Index MCP Server Changelog
 
 ## [Unreleased]
+### Added
+- **Ruby language support** — `ide_type_hierarchy`, `ide_call_hierarchy`, `ide_find_implementations`, `ide_find_super_methods`, and `ide_file_structure` now work in RubyMine and IntelliJ IDEA Ultimate with the Ruby plugin. Ruby modules are treated as interface analogues for `ide_find_implementations` and `ide_type_hierarchy` (classes that `include`/`extend` a module appear as implementations). `language+symbol` mode added for Ruby (`UserService`, `UserService#find`, `Services::UserService#find`).
+- RubyMine is now a fully supported IDE (previously listed as "may work, untested").
 
 ## [4.25.0] - 2026-06-27
 ### Added
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -391,7 +391,7 @@ Tools are organized by IDE availability.
 **Extended Navigation Tools (Language-Aware):**
 
 These activate based on available language plugins (Java, Python, JavaScript/TypeScript, Go, PHP, Rust, Markdown):
-- `ide_type_hierarchy` - Get type hierarchy for a class (Java, Kotlin, Python, JS/TS, Go, PHP, Rust)
+- `ide_type_hierarchy` - Get type hierarchy for a class (Java, Kotlin, Python, JS/TS, Go, PHP, Ruby, Rust)
 - `ide_call_hierarchy` - Get call hierarchy for a method (Java, Kotlin, Python, JS/TS, Go, PHP, Rust). Supports `language`+`symbol` as alternative to `file`+`line`+`column`.
 - `ide_find_implementations` - Find implementations of interface/method (Java, Kotlin, Python, JS/TS, PHP, Rust — not Go). Supports `language`+`symbol` as alternative to `file`+`line`+`column`.
 - `ide_find_super_methods` - Find methods that a given method overrides/implements (Java, Kotlin, Python, JS/TS, PHP — not Go, Rust). Supports `language`+`symbol` as alternative to `file`+`line`+`column`.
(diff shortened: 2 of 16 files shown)


hechtcarmel:
Hi,
Thanks for this important contribution.

I don't think this is ready to merge yet. I found two behavior issues and one docs cleanup.
**NOTE** - I myself am not a Ruby developer and used AI to help review it. So take with a grain of salt 

1. `ide_type_hierarchy` now sends unrelated class-name misses through the Ruby resolver.

After Java/PHP class lookup and JS/TS lookup fail, `TypeHierarchyTool` always tries Ruby if the Ruby handler is registered. That means a normal unresolved JVM/PHP name can produce a Ruby grammar error just because the Ruby plugin is installed.

Example:

```json
{"className": "com.example.DoesNotExist"}
```

Expected: generic class-not-found error.
Actual with Ruby available: `Unsupported Ruby symbol format`.

Same problem for PHP-style names like:

```json
{"className": "App\\Models\\Missing"}
```

The Ruby resolver should only run for plausibly Ruby syntax, or its parse failure should not replace the normal class-not-found result.

2. Ruby call hierarchy advertises callees support, but callees always returns an empty successful result.

The Ruby handler currently does:

```kotlin
when (direction.lowercase()) {
    "callers" -> findCallers(...)
    else -> emptyList()
}
```

So this call:

```json
{"language": "Ruby", "symbol": "UserService#find", "direction": "callees"}
```

returns `calls: []` even if the method calls other methods. That is worse than unsupported, because clients will treat it as true "this method calls nothing" evidence. Either implement Ruby callees, or return a clear unsupported error and document Ruby as callers-only for now.

3. `USAGE.md` has an accidental SQL query at the top and still misses Ruby in the tool table.

The file now starts with:

```sql
SELECT
arguments->>'$.language' AS lang,
...
```

That should be removed. The same table still lists the extended tools without Ruby, while the README/changelog/plugin description say Ruby is supported.

billdueber:
Looking back at this, it was a terrible first attempt :-) I'm going to take another, more-deliberate shot at it.

PR #239, opened 2026-07-11
feat: add ide_list_tests and ide_run_tests tools

jevanlingen:
## Summary
- Adds `ide_list_tests` — discovers all test methods/classes via the IDE's test framework extension points (JUnit, TestNG, etc.), with optional file-scoped filtering
- Adds `ide_run_tests` — runs tests via the IDE's run configuration infrastructure; accepts an existing run config name or a Java/Kotlin FQN (`com.example.MyTest` / `com.example.MyTest#testFoo`); returns structured pass/fail/error counts, exit code, and console output
- Both tools are opt-in (disabled by default) and registered in all required locations: `ToolNames`, `ToolRegistry`, `McpSettings`, docs, and tests

## Why
Enables coding agents to discover and execute tests directly through the IDE without needing to shell out to build tools, giving them access to structured test results from any Service-Message-based framework.

## Testing / Reflections
Written with AI assistance (as I don't have prior IntelliJ development plugin experience), with careful review and manual testing by building the plugin and running tests via the MCP tool — works well for both Gradle and JUnit (Maven) projects.

One open question: `injectMethodFilter` uses reflection to set method-level filters on run configurations, since `ConfigurationContext` only produces class-level configs. Happy to hear if there's a cleaner approach.


hechtcarmel:
Hi, 
Thank you for contributing. This looks like a great addition. Please leave a message when ready to review

hechtcarmel:
beyond build and merge issuesI do have a few things I'd like to see addressed before this goes in, mostly in `ide_run_tests`.


**1. The run-config mutation runs off the EDT.**

In `resolveRunConfiguration` (`RunTestsTool.kt:241`):

```kotlin
return edtAction { createConfigurationFromContext(project, psiElement) }
    ?.also { config ->
        if (methodName != null) injectMethodFilter(config.configuration, className, methodName)
        runManager.setTemporaryConfiguration(config)
    }
```

`edtAction { }` returns to the calling (worker) thread, so the `.also { }` block runs off the EDT — which means we're mutating a `RunConfiguration` and calling `RunManager.setTemporaryConfiguration` from a background thread. That's model/write work that should be on the EDT; right now it works mostly by luck because the config is brand new. Can we move the create → mutate → register into a single `edtAction`?

```kotlin
return edtAction {
    val config = createConfigurationFromContext(project, psiElement) ?: return@edtAction null
    if (methodName != null) injectMethodFilter(config.configuration, className, methodName)
    runManager.setTemporaryConfiguration(config)
    config
}
```

**2. Single-method resolution can grab the wrong element.**

`RunTestsTool.kt:236`:

```kotlin
PsiTreeUtil.collectElementsOfType(psiClass, PsiNamedElement::class.java)
    .firstOrNull { it.name == methodName }
```

This matches *any* named element, not just methods — so a field or parameter that happens to share the method's name comes first:

```java
class MyTest {
    private final Object testFoo = new Object(); // this wins
    @Test void testFoo() { }                      // the one we actually wanted
}
```

When we then hand a field to `ConfigurationContext`, it can't build a method-scoped config — which I suspect is the real reason the `injectMethodFilter` reflection had to be added afterwards. If we restrict this to actual methods (`PsiMethod` / `KtNamedFunction`, or filter through `TestFramework.isTestMethod`), the context should produce a method-scoped config on its own and a lot of the reflection below may just go away.

**3. `ProcessCanceledException` is being swallowed in the discovery walk.**

`ListTestsTool.kt:131`:

```kotlin
val framework = frameworks.firstOrNull {
    runCatching { it.isTestMethod(element) }.getOrDefault(false)
} ?: return null
```

`runCatching` catches `Throwable`, so it eats `ProcessCanceledException` too. `isTestMethod` can trigger resolution that throws PCE, and since this runs inside a read action, PCE needs to propagate (the platform even logs "must be rethrown"). Something like:

```kotlin
frameworks.firstOrNull { fw ->
    try { fw.isTestMethod(element) }
    catch (e: ProcessCanceledException) { throw e }
    catch (e: Exception) { false }
}
```

Same pattern (catch broad, no PCE rethrow) is in the `injectGradle/JUnitMethodFilter` helpers — lower risk there, but worth tidying too.

**4. `#method` silently runs the whole class for non-JUnit/non-Gradle configs.** In `injectMethodFilter` (`RunTestsTool.kt:248`) only `JUnit` and Gradle get a filter; everything else (TestNG, etc.) falls through to a `debug` log and runs the entire class. The docs present `class#method` as a general Java/Kotlin feature, so at minimum I think we should tell the caller when we couldn't narrow to the method (e.g. a `scope`/`note` field in the result) rather than returning a green result for a broader run than they asked for.

**5. `success` is just `exitCode == 0`.** (`RunTestsTool.kt:205`) Two cases this gets wrong: an empty test tree + exit 0 reports `success: true, total: 0` (can't tell "all passed" from "nothing ran"), and a runner that exits 0 despite failures gives `success: true` alongside `failed > 0`. Could we do `success = exitCode == 0 && failed == 0 && errors == 0`, and maybe flag the "no tests found" case?

**6. Process listener is attached after the process starts.** (`RunTestsTool.kt:166`) We await `processStarted`, then add the listener. `addProcessListener` doesn't replay, so if a run finishes in that gap we never get `processTerminated` and report a false timeout. Cheap to close:

```kotlin
handler.addProcessListener(listener)
if (handler.isProcessTerminated) exitCodeDeferred.complete(handler.exitCode ?: -1)
```

**7. The reflection-based filter injection** (`RunTestsTool.kt:261`+) — you already called this out, and I agree it's the shakiest part. Fixing #2 should remove most of the need for it; whatever remains I'd lean on the typed JUnit API (the plugin already depends on the Java plugin for FQN creation) and surface failures instead of only logging them.

## Nits

- `ide_list_tests` walks every PSI node of every test file calling `isTestMethod` per framework (`ListTestsTool.kt:107`). Fine for now and it's opt-in, but a stub-index lookup would scale a lot better on big repos.
- Root capture (`RunTestsTool.kt:143`) assumes ours is the first `onTestingStarted` — a concurrent manual run or a second `ide_run_tests` call could hand us the wrong tree. The comment notes it; maybe mention it in the tool description too.
- `parseTarget` treating a trailing lowercase segment as a method (`RunTestsTool.kt:310`) is ambiguous; worth making it `internal` and adding a small unit test since it's pure logic.
- `RunTestsResult` has no `skipped` count even though `TestStatus.SKIPPED` exists, so `passed + failed + errors` can be `< total` with nothing explaining the gap.
- The Gradle filter value is wrapped in literal quotes: `"\"$className.$methodName\""` (`RunTestsTool.kt:271`). Since these go in as individual task-name tokens, are the embedded quotes intentional? Want to make sure they don't end up in the `--tests` pattern.
- The added tests cover schema/registration/serialization only — none of `parseTarget`, the status mapping, or the method-element selection. A small platform test that lists + runs a fixture would be great insurance.

jevanlingen:
Thanks for the detailed review. I fixed the "reflection-based filter injection"; took me a good while. AI couldn't find the proper solution. Apparently, the whole fix consist of:

```diff
- .add(CommonDataKeys.PSI_ELEMENT, psiElement)
+ .add(Location.DATA_KEY, PsiLocation.fromPsiElement(psiElement))
```

(Thanks to good old googling and finding [this](https://github.com/JetBrains/intellij-community/blob/master/platform/testFramework/src/com/intellij/testFramework/PlatformTestUtil.java#L1231-L1235) gem).

---

I will pickup the other points later.

hechtcarmel:
ci issues..

hechtcarmel:
beyond that - 

## 1. `ide_run_tests` does not reliably observe the execution it launched

There are two related problems in `RunTestsTool.kt`.

First, the process listener is attached only after `processStarted`:

```kotlin
override fun processStarted(..., handler: ProcessHandler) {
    processHandlerDeferred.complete(handler)
}

// Later, after the coroutine resumes
handler.addProcessListener(...)
```

In the 2025.3 execution flow, the handler is started before `processStarted` is published, and adding a listener later does not replay output or termination. A fast/cached test can finish before this listener is attached. The test passes, but `exitCodeDeferred` never completes, so the tool waits until the user timeout and returns `timedOut=true`, `exitCode=-1`, with early output missing.

Second, the subscriptions are project-global but correlation is only by configuration name, while the first SM test root after subscription is accepted unconditionally:

```kotlin
if (environment.runnerAndConfigurationSettings?.name == configName) { ... }

testRoot.compareAndSet(null, testsRoot)
```

Two simultaneous calls for `All Tests`, or a manual test run starting at the same time, can make this request capture another execution's handler or test tree. That can return a plausible successful response containing the wrong tests.

The fix should bind the whole request to the exact `ExecutionEnvironment`/handler, attach the process listener from `processStarting(..., handler)` before `startNotify`, and obtain the test console from that handler. The public `RunContentManager.findContentDescriptor(executor, handler)` API gives the descriptor for the exact process, so the “first global test root” assumption is not needed.

The same ownership should cover cleanup: startup timeout currently does not cancel the scheduled run, coroutine cancellation only disconnects the message bus, and `destroyProcess()` is not awaited. A slow before-launch build can therefore start tests after the tool already returned an error, and a retry can launch a duplicate suite.

## 2. `ide_list_tests` is registered as universal, but its discovery API belongs to the optional Java plugin

`registerUniversalTools()` says those tools use platform APIs only, yet both new tools are registered there unconditionally. `ide_list_tests` calls:

```kotlin
TestFramework.EXTENSION_NAME.extensionList
```

The `com.intellij.testFramework` extension point is declared by the Java plugin, while this plugin keeps `com.intellij.modules.java` optional. When Java is absent, `extensionList` throws `Missing extension point: com.intellij.testFramework`; the following `frameworks.isEmpty()` check is never reached.

For example, calling `ide_list_tests` in WebStorm can fail before returning a tool result, even though README lists it under tools that work in every supported IDE. Python/JS/PHP/Go/Rust discovery also does not come from this Java extension point.

I would register/document `ide_list_tests` behind `PluginDetectors.java` for the JVM frameworks this implementation actually supports, then add other languages only when their native discovery APIs are implemented. Java-specific `PsiClass`/`PsiMethod` resolution should likewise stay behind the optional Java feature boundary so the existing-run-configuration path can remain platform-only.

## 3. Both PSI-based tools disable the repository's external-file synchronization

`ListTestsTool` and `RunTestsTool` both set:

```kotlin
override val requiresPsiSync: Boolean = false
```

That opt-out is intended for tools that do not use PSI/indexes. These tools do: listing walks PSI/test roots, and FQN execution resolves a class/method from PSI.

Concrete case: an agent creates `NewBehaviorTest.kt` outside IntelliJ and immediately calls `ide_list_tests` to check whether it was discovered. Even when the user enabled “Sync external file changes,” these overrides skip the VFS refresh/document commit and can omit the new test or fail to resolve it for execution.

Removing both overrides should be enough. If an existing named run configuration needs a cheaper path later, that can be separated without making the PSI paths stale.

## 4. Structured result mapping can report the wrong tests and statuses

`collectRunEntries` checks `isPassed` before skipped/ignored state. In the 2025.3 SDK, `isPassed` also returns true for a skipped-state proxy, so a skipped test can increment `passed` instead of being returned as `skipped`.

It also treats every non-root leaf as a test. An empty suite/configuration is a leaf too, so it falls through to `FAILED` and becomes a failed test that never actually ran.

Please map statuses from supported public state/events, handle skipped/ignored before the broad passed predicate, and exclude suite/configuration nodes. A small mapping test covering passed, skipped, failed, error, and an empty suite would catch this.

## Smaller correctness/documentation gaps

- `parseTarget` treats any lowercase final FQN segment as a method. A legal class such as `com.example.checkoutTest` is parsed as class `com.example` + method `checkoutTest`. The existing `Class#method` syntax is unambiguous; I would delete the dot heuristic.
- A misspelled `file` passed to `ide_list_tests` returns a successful empty list, indistinguishable from a real file containing no tests. It should return file-not-found.
- Exactly 500 discovered tests returns `truncated=true` even though nothing was omitted. Collect one extra candidate or track whether another match exists.
- README still says 40 tools while `ToolNames.ALL` now has 43, and the IDE matrix still says 16 universal tools. The bundled skill also adds `ide_run_tests` to its trigger while still routing all “run” work to the built-in JetBrains MCP only.
- The new tests cover schemas/registration but not discovery, fast termination, concurrent runs, cancellation, invalid files, the 500-result boundary, or result mapping. Those behavior paths are where the risks above live.

The overall direction looks good, and `ConfigurationContext` is a sensible way to create Java/Kotlin run configurations. The main thing is to make one MCP call own one exact IntelliJ execution from startup through cleanup, and to scope discovery to the IDEs/frameworks the implementation actually supports.

jevanlingen:
I think I’ve addressed all the points mentioned 🤞🏻. I hope it’s in a good enough state now, this PR ended up taking a lot more time than I expected.

P.S. I’ve already been using the tools this week, and for a standard Maven project they’re working great!

hechtcarmel:
nooo
I was gonne approve it after testing locally and it didn't pass verification 😢 
Looks like you are using an internal api

jevanlingen:
> It didn't pass verification 😢. Looks like you are using an internal api.

Yeah, that’s right, because of the 

> Structured result mapping can report the wrong tests and statuses. collectRunEntries checks isPassed before skipped/ignored state. In the 2025.3 SDK, isPassed also returns true for a skipped-state proxy, so a skipped test can increment passed instead of being returned as skipped.

comment, Claude came up with a [fix](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/239/changes/f4a2bfbe2522ee022210e99f534404bda2c6acce). Sadly this used an internal api. I did not know the pipeline would fail because of it, but makes sense you configured it that way!

So I [updated](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/239/changes/e54fb8b81fbdc947a7ca8f1148c4cb61a1e02f4f) the code to use an int instead of the actual enum, which does the same thing, but type unsafe (thus uglier code, but it bypasses the internal api...). I would suggest to move forward, and see if the API becomes stable in a newer version of the ide.

hechtcarmel:
Great, thank you for the contribution.
I'm merging and willsend it as part of v4.29.0  later today (jetbrains takes anywhere between 10 minutes to 5 days to approve)

PR #245, opened 2026-07-20
Add support for Scala 2 (do not use reflection)

ilx:
This is additional set of changes on top of https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/174

Not sure if Open-sourcing the LSP Client API has impact on this server (see https://blog.jetbrains.com/platform/2026/06/open-sourcing-the-lsp-client-api-in-intellij-idea-2026-2/)


hechtcarmel:
verification fails..

hechtcarmel:
The core problem — ScalaPluginDetector calls PluginManagerCore.getPlugin(), which is @ApiStatus.Internal in IntelliJ 2026.2, so the "Verify plugin" CI job fails with INTERNAL_API_USAGES.

hechtcarmel:
Other comments from claude:

Other blockers found:

All 10 new Scala tests skip via silent early-return instead of Assume.assumeTrue() — an explicit CONTRIBUTING.md violation, and I genuinely could not tell from the green CI Test log whether any Scala test actually executed. The repo's own ./scripts/check-pr.sh fails on this (I ran it; local unit tests pass once run with JDK 21).
No [Unreleased] CHANGELOG entry for a user-visible feature.
Notable non-CI findings (all detailed with code examples in Review.md):

The PR abandons the repo's reflection pattern for direct Scala PSI access, but keeps catch (Exception) safety nets — Scala-plugin API drift throws NoSuchMethodError (an Error), so version mismatch now crashes tool calls where reflection degraded softly.
The CLAUDE.md edit documents the Scala handler as "Reflection-based" — the exact opposite of what the PR does.
Likely bug: val/var names in ide_file_structure probably render as "unknown" (blind cast on a declaration holder instead of using declaredElements), and the tests are too loose to catch it.
Scope creep: .gitignore un-ignores .idea/ (risking the forbidden .idea/gradle.xml), and the new .ignore file is copy-pasted from a JS project (.eslintrc.js, .yarn, etc.).

hechtcarmel:
Hi, i see yo uare continuing to update this. great. Please write a comment so i know when to re-review it 😄

PR #249, opened 2026-07-21
fix: ide_search_text tool should also match substrings

jevanlingen:
## Summary

- Replace `PsiSearchHelper.processElementsWithWord` (word-index path) with `FindInProjectUtil` for **all** plain-text queries, not just regex ones.
- Plain-text searches now set `isRegularExpressions = false` on the `FindModel`, matching what IntelliJ's own Find in Files dialog does when "In Project" is selected.
- Remove the now-dead `searchText`, `extendSearchText` (word-index branch), and `createSearchScope` helpers.

## Why

`processElementsWithWord` searches IntelliJ's pre-built word index, which only stores **complete tokens**. Any query that is a prefix or substring of a token returns zero results. For example, searching for `aWord` would fail to find `aWordAndMore`, and searching for `a_word` would fail to find `a_word_and_another_word`, because in both cases the full identifier is stored as one token.

IntelliJ's own Find in Files dialog uses `FindInProjectUtil` with `isRegularExpressions = false`, which does a **substring scan** over file content and finds both cases correctly. The tool was using the word index — O(1) but whole-token only. Correctness wins over speed here.

The screenshot below shows IntelliJ's Find in Files returning matches in files:

image

where `ide_search_text` returned zero:

```
Tool: ide_search_text
Status: SUCCESS
Timestamp: 2026-07-21T11:21:58.349993Z
Duration: 1ms

Parameters:
{"query":"_jobs_stale"}

Result:
{"matches":[],"totalCount":0,"query":"_jobs_stale","nextCursor":null,"hasMore":false,"totalCollected":0,"offset":0,"pageSize":0,"stale":false}
```

## Test

Added `testSearchTextToolFindsSubstringOfUnderscoreSeparatedToken` in `ToolsTest`:

1. Creates a YAML file containing `a_word_and_another_word`.
2. Searches for `a_word`.
3. **Before fix**: `processElementsWithWord` returns 0 — assertion fails.
4. **After fix**: `FindInProjectUtil` finds the substring — assertion passes.


hechtcarmel:
Hi,
Thank you for contrinbuting. 
I'll look into it soon

hechtcarmel:
Thanks for this — the diagnosis is spot on.

Two things I need changed before I can merge, though.

### 1. `searchRegex` isn't thread-safe, and this PR puts it on every code path

```kotlin
val results = mutableListOf<TextMatch>()   // ArrayList
val seenLines = HashSet<String>()
```

`FindInProjectUtil` doesn't call the processor on one thread. `FindInProjectTask.unfoldAndProcessSearchItems` goes through `FilesScanExecutor.processOnAllThreadsInReadActionWithRetries`, so the callback fires concurrently from a worker pool (each inside its own read action, so the PSI access in `convertToTextMatch` is fine — it's only the collection that's a problem). You can see the platform's own signature further down leans entirely on concurrent primitives:

```java
private boolean processFindInFilesUsagesInFile(
    FindUsagesProcessPresentation, Processor<? super UsageInfo>,
    AtomicInteger, ConcurrentMap<? super VirtualFile, Set<UsageInfo>>,
    AtomicBoolean, StringSearcher, VirtualFile);
```

The code you deleted actually got this right — the old `searchText` used `ConcurrentLinkedQueue` + `ConcurrentHashMap.newKeySet()` + `AtomicInteger`. `searchRegex` never did, but that only ever ran on opt-in `regex: true` calls so nobody noticed. After this PR it's every single search.

Practically: dropped matches (silently — the same class of bug this PR is fixing), possible `ArrayIndexOutOfBoundsException` from a racing `ArrayList` grow, and a non-deterministic `limit` cutoff since `results.size` is read unsynchronized. The new test won't catch any of it — one small file means effectively one thread.

Can you put the concurrent collections back? Roughly:

```kotlin
private fun searchRegex(
    project: Project, findModel: FindModel, searchContext: Short,
    limit: Int, seenKeys: Set<String> = emptySet()
): List<TextMatch> {
    val results = ConcurrentLinkedQueue<TextMatch>()
    val seenLines = ConcurrentHashMap.newKeySet<String>().apply { addAll(seenKeys) }
    val count = AtomicInteger(0)
    val presentation = FindInProjectUtil.setupProcessPresentation(
        FindInProjectUtil.setupViewPresentation(findModel))

    FindInProjectUtil.findUsages(findModel, project, presentation, emptySet<VirtualFile>()) { usageInfo ->
        if (count.get() >= limit) return@findUsages false
        val match = convertToTextMatch(project, usageInfo, searchContext) ?: return@findUsages true
        if (seenLines.add("${match.file}:${match.line}")) {
            if (count.incrementAndGet() <= limit) results.add(match)
        }
        count.get() < limit
    }
    return results.toList()
}
```

That keeps the three imports you removed.

### 2. The `filePattern` validation went with the dead code

This block got dropped along with `createFileMaskCondition`:

```kotlin
val fileMaskCondition = try {
    createFileMaskCondition(filePattern)
} catch (e: PatternSyntaxException) {
    return createErrorResult("Invalid filePattern mask: ${e.message}")
}
```

It was running for both branches, not just the word-index one. The mask still gets parsed, just further down where nothing catches it — `FindInProjectUtil.createFileMaskCondition` declares `throws PatternSyntaxException` and it's called from the `FindInProjectTask` constructor. So a malformed mask (unbalanced `[`, say — masks go through `PatternUtil.convertToRegex`) now escapes `doExecute`, and `AbstractMcpTool.execute` only special-cases `"Outdated stub"` and rethrows everything else. The agent gets a raw stack trace instead of a readable message.

Wrapping the `findUsages` call is enough. Just remember `extendSearch` needs the same guard, since that one runs later from the pagination service.

---

Smaller stuff, none of it blocking:

**CHANGELOG.** Needs a `### Changed` entry under `[Unreleased]` — this is a semantics change to a tool that's on by default. `check-pr.sh` won't flag it because it only checks the section exists.

**Whole-word matching is gone entirely.** Everything used to be implicit whole-word, now `{"query": "id"}` matches `uuid`, `valid`, `idle`, `android`. I think substring is the right default (it's what the IDE does), but the escape hatch should be findable. Either add a `wholeWord` param — `FindModel.setWholeWordsOnly` exists, it's a two-line change — or just mention `regex: true` with `\bfoo\b` in the tool description. Right now there's nothing telling an agent either option exists.

**Test coverage.** The new test nails the reported bug, but `filePattern`, `context`, and `caseSensitive` all switched implementation on the plain-text path and none of them are covered. `filePattern` in particular went from a `DelegatingGlobalSearchScope` to `findModel.fileFilter`, which is a completely different mechanism. Three more tests in the same style would cover it. Also worth noting the old `caseSensitive: false` path had an explicit `contains(word, ignoreCase = true)` re-check that's now gone.

**Import churn in `ToolsTest.kt`.** Looks like an accidental Optimize Imports — about 40 of the PR's 99 added lines are explicit imports collapsed into wildcards. Could you revert that block? Same for the two `buildPaginatedResult(...)` calls that got reflowed from one line to six. Makes the actual change much easier to see.

**`isProjectScope = true` in `createFindModel`** is dead — `FindInProjectTask` reads `isCustomScope()`/`getCustomScope()`/`getModuleName()` and never touches `isProjectScope()`. Pre-existing, but since this is now the only search path it'd be nice to drop the misleading line.

jevanlingen:
Done

hechtcarmel:
Thank you for the contribution.
I'm merging and will send  a new version for jetbrains to review with it either today ot over the weekend

PR #251, opened 2026-07-22
fix: ide_search_text supports multi-line queries

mdproctor:
## Summary

- Non-regex queries containing newlines now auto-route to the FindInFiles path instead of the word index
- The word index (`PsiSearchHelper.processElementsWithWord`) splits text into individual words and fundamentally cannot match multi-line patterns
- Multi-line queries are escaped with `Pattern.quote()` to preserve literal matching, then routed through `FindModel` which already has `isMultiline = true`

## Example

Before: `{"query": "Map.class,\n            input"}` → no results (word index splits on whitespace/newlines)

After: same query → finds the multi-line match via FindInFiles

## Test plan

- [x] Unit tests pass
- [ ] Manual verification with multi-line query in runIde


hechtcarmel:
conflict with a pr i just merged 😅

mdproctor:
Closing — PR #249 (merged) removed the word-index path that couldn't handle multi-line queries. All queries now go through FindInProjectUtil with isMultiline=true, so multi-line works natively without this change.

PR #252, opened 2026-07-23
fix: block temp script hook bypass in claude-code-hooks

mdproctor:
## Summary

- Agents bypass Edit/Write hooks by writing scripts to temp directories that manipulate source files directly via file I/O
- The previous hook had an explicit exception allowing temp script execution — this was the exact escape hatch agents exploited
- Now temp script execution is blocked with an error message naming the correct IDE tools
- Also adds a pattern catching inline Python commands that write to source files

## Why this matters

The hooks exist to enforce IDE MCP tool usage (ide_replace_text_in_file, ide_edit_member, etc.) so changes go through IntelliJ's Document/PSI layer. Temp scripts bypass VFS, PSI, indexing, and undo — the IDE doesn't see the changes until a manual sync.

## Test plan

- [x] Hook blocks temp script execution with actionable error message
- [x] Normal bash commands still work
- [x] Unit tests pass

### DIFF
--- a/docs/claude-code-hooks.md
+++ b/docs/claude-code-hooks.md
@@ -100,15 +100,26 @@ if echo "$CMD" | grep -qE 'sed\s+-[a-zA-Z]*i' && \
   exit 2
 fi
 
-# python3 bulk replacement → ide_replace_text_in_file
-# Exception: scripts in /tmp/ are development tooling
-if echo "$CMD" | grep -qE 'python3\s' && \
-   echo "$CMD" | grep -qE '\.java|\.kt|\.ts|\.tsx|\.js|\.jsx|\.py' && \
-   echo "$CMD" | grep -qE 'replace|sub\(|re\.sub'; then
-  if echo "$CMD" | grep -qE 'python3\s+/tmp/'; then
-    exit 0
-  fi
-  echo "BLOCK: Use ide_replace_text_in_file instead of Python bulk replacement on source files." >&2
+# python3 /tmp/ scripts — common hook bypass pattern.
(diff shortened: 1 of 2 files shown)


hechtcarmel:
Fable flagged the following:

1. New false positives on everyday commands. The rewritten inline-python rule added open(/write( as keywords, and \.js substring-matches .json. Verified: python3 -c "print(open(\"package.json\").read())" — a plain read — is now blocked (allowed at the merge base). Same for reading tsconfig.json or a .ts file. Word-bounding the extensions (\b) and dropping bare open( fixes it.

2. .py was silently dropped from that rule, so python3 migrate.py --replace ... src/app.py is no longer blocked — while the doc's own "What it catches" table in the same file still claims it is. The PR description doesn't mention this removal. There's a defensible rationale (Hook 2 exempts Python), but it needs to be an explicit, documented decision.

mdproctor:
Fixed both issues, rebased on current main.

**1. False positives fixed:**
- Dropped bare `open(` from write-detection keywords — it catches reads. Pattern now: `write\(|replace|sub\(|re\.sub`
- Word-bounded extensions (`\.java\b`, `\.ts\b`, etc.) — prevents `.json`, `.jsonl` false positives
- Pattern 5b now only matches when `python3` is the actual command (start of line or after `&&`/`;`/`|`) — prevents false positives in `gh pr comment` bodies and similar commands that mention python/replace/.ts in their text arguments

**2. `.py` exclusion documented:**
Comment in the hook explains: "Python (.py) excluded — Hook 2 does not block Python file edits because there is no IDE structural editing support for Python yet."

57 tests, 0 failures (5 new false-positive regression tests).

PR #254, opened 2026-07-23
feat: add ide_create_module — generic module creation for any directory

mdproctor:
## Summary

- Adds `ide_create_module` — registers a directory as an IntelliJ module with a content root, enabling code intelligence for non-Maven projects (TypeScript, plain directories, git worktrees)
- Complements `ide_import_modules` (Maven-only) and `ide_open_workspace` (Maven workspace aggregator)
- Supports optional directory exclusions (e.g., `node_modules`, `dist`, `.next`)

## Why

When agents work in git worktrees or non-Maven projects, IntelliJ tools like `ide_find_references` and `ide_diagnostics` don't cover the files because the directory isn't registered as a content root. This tool fills that gap without requiring Maven.

## Implementation

- `CreateModuleTool.kt`: creates a `WEB_MODULE` with a content root and optional excluded directories
- Module creation runs inside `WriteAction` on EDT (required for `ModuleManager.newModule` and `rootModel.commit`)
- Guards: non-absolute path, non-existent directory, pom.xml present (directs to `ide_import_modules`), `..` in excludes, already-registered content root (idempotent)
- Disabled by default (schema version 6 migration)

## Checklist

- [x] All 6 doc locations (README, USAGE, CLAUDE.md, SKILL.md, tools-reference.md, CHANGELOG)
- [x] ToolNames constant + ALL list (alphabetical)
- [x] McpSettings: DEFAULT_DISABLED_TOOLS + schema v6 migration
- [x] ToolRegistry registration
- [x] Unit tests: schema, disabled-by-default, migration from v2
- [x] `check-pr.sh` passes all 13 checks
- [x] CONTRIBUTING.md code correctness checklist: threading, error handling, sibling consistency


hechtcarmel:
### 1. CHANGELOG entry is in the wrong section (required)

The entry was added under `## [4.28.0] - 2026-07-18`, which already shipped. That makes the release notes claim the tool existed in 4.28.0 (it doesn't), and since the next release collects the `[Unreleased]` section — which is empty — the tool would never show up in any release notes at all.

Per CONTRIBUTING.md, entries go under `## [Unreleased]` with an `### Added` heading. Just move the bullet there. (Not entirely on you — `check-pr.sh` should have caught this and doesn't; I'll fix the script separately.)

### 2. Module-name collision throws a raw platform exception (required)

The idempotency check compares content-root paths but never the module *name*, which defaults to the directory basename. Module names must be unique per project, so the tool's own headline scenario breaks:

```
ide_create_module {"path": "/work/feature-a/webapp"}   → module "webapp" created
ide_create_module {"path": "/work/feature-b/webapp"}   → ModuleWithNameAlreadyExists
```

`newModule` throws `ModuleWithNameAlreadyExists` when the name exists at a different path (see `ModifiableModuleModelBridgeImpl.newModule`). Nothing catches it: `AbstractMcpTool.execute` only maps `IndexNotReadyException` and stale-stub errors, so it propagates to `JsonRpcHandler`, which `LOG.error`s it — that fires the red "IDE internal error" notification for the user — and returns the bare `Module already exists: webapp` to the agent with no hint about the `name` parameter.

Please add a pre-check before the write action:

```kotlin
if (ModuleManager.getInstance(project).findModuleByName(moduleName) != null) {
    return createErrorResult(
        "A module named '$moduleName' already exists in this project. " +
        "Pass a different 'name' to create a module for this directory."
    )
}
```

### 3. `ModifiableRootModel` isn't disposed on failure (required)

Between `modifiableModel` and `rootModel.commit()` nothing is protected. If `addContentEntry` or an exclude throws, the model is neither committed nor disposed — the javadoc is explicit that `dispose()` "must be invoked for uncommitted models" — and the module created one line earlier is left in the project with no content root. Either wrap it:

```kotlin
val rootModel = ModuleRootManager.getInstance(module).modifiableModel
try {
    // addContentEntry + excludes
    rootModel.commit()
} catch (e: Exception) {
    rootModel.dispose()
    throw e
}
```

or use `ModuleRootModificationUtil.updateModel(module) { ... }`, which handles it for you.

### Strongly suggested (won't block on these, but they're cheap and worth doing here)

- **Use `addExcludeFolder(url)` instead of resolving a `VirtualFile` per exclude.** Right now excludes only apply to directories that already exist, so the main use case fails silently: fresh clone → `excludes: ["node_modules"]` does nothing → `npm install` → IntelliJ indexes all of it. The URL overload (`contentEntry.addExcludeFolder(VfsUtilCore.pathToUrl(...))`) is string-based, works for not-yet-existing dirs, and also removes the synchronous `refreshAndFindFileByPath` calls you're currently doing on the EDT inside the write action.
- **Surface the side effects.** The tool writes `<name>.iml` into the target directory (i.e. into the user's git working tree) and registers the module permanently in `.idea/modules.xml`, with no inverse tool. The success message should include the `.iml` path, and USAGE.md should mention both effects plus how to undo (Project Structure → Modules).
- **Add tests for the guards.** The new tests cover schema/registration/migration, but nothing executes `doExecute` — the relative-path, misasing-dir, `pom.xml`, and already-registered paths are all testable in the existing `ToolExecutionIntegrationTest` style without a heavy fixture.

### Nits

- `name` is unvalidated: `"name": "../foo"` writes the iml outside the target dir, `""` produces `.iml`. The excludes have a `..` guard; give `name` the same treatment.
- `excludes: [""]` resolves to the content root itself and excludes the whole module.
- The migration test message says "v2→v3" but `CREATE_MODULE` ships in the v6 migration; seeding `settingsSchemaVersion = 5` would test the actual step.
- The description tells agents to use "File → New → Module from Existing Sources" for Gradle — an agent can't click menus; pointing at `ide_open_project` is more useful.
- This is the only tool using bare `WriteAction` instead of `WriteCommandAction`. It's defensible for project-model changes (nothing to undo), just worth a line in the PR description.
- Maybe add a note in the description that indexing of the new root is async — an immediate `ide_find_references` can hit dumb mode, `ide_index_status` is the remedy.

FWIW I also checked the platform side: `newModule(String, String)` is non-deprecated public API, and `"WEB_MODULE"` degrades gracefully where it isn't registered (DataGrip shows "Unknown module type", indexing unaffected) — so no concerns there. `EmptyModuleType.EMPTY_MODULE` would avoid even the cosmetic case if you feel like it.

mdproctor:
Addressed everything. Force-pushed a squashed rewrite (14 files, 300 insertions).

**Required fixes:**

1. **CHANGELOG** — moved from `[4.28.0]` to `[Unreleased]` under `### Added`.

2. **Module-name collision** — pre-check via `ModuleManager.findModuleByName()` before the write action. Returns actionable error with hint about the `name` parameter.

3. **ModifiableRootModel dispose** — `rootModel.commit()` now inside try/catch; `rootModel.dispose()` in the catch block.

**Suggestions taken:**

4. **URL-based excludes** — switched to `contentEntry.addExcludeFolder(VfsUtilCore.pathToUrl(...))`. Works for not-yet-existing directories. Removed synchronous `refreshAndFindFileByPath` calls from the write action.

5. **Name validation** — rejects blank, `..`, `/`, `\` in module name (same `..` guard as excludes).

6. **Empty exclude guard** — `exclude.isBlank()` skip added.

7. **Surface side effects** — success message now includes the `.iml` path and an async-indexing note directing to `ide_index_status`.

8. **USAGE.md** — added side-effects paragraph mentioning `.iml` file, `.idea/modules.xml` registration, and how to undo (Project Structure → Modules).

9. **Migration test** — renamed to `testSchemaVersion5MigrationKeepsCreateModuleDisabled`, seeds `settingsSchemaVersion = 5` (the actual prior version).

10. **Description** — Gradle guidance changed from "File → New → Module" to `ide_open_project`.

11. **WriteAction vs WriteCommandAction** — kept bare `WriteAction` (project-model changes have no undo semantics). Noted in the commit message.

mdproctor:
Closing to re-create — fork main synced to upstream after SDK migration (#261), merge base is stale.

PR #255, opened 2026-07-23
add first-class ruby support when ruby plugin available

billdueber:
`RubyHandlers`/`LanguageHandlerRegistry`) for:

- `ide_call_hierarchy` (`RubyCallHierarchyHandler`)
- `ide_type_hierarchy` (`RubyTypeHierarchyHandler`)
- `ide_find_super_methods` (`RubySuperMethodsHandler`)
- `ide_file_structure` (`RubyStructureHandler`)
- `ide_resolve_symbol` (`RubySymbolReferenceHandler`)

It also extends hierarchy data with a `via` provenance field so super-method edges can distinguish `superclass` vs Ruby mixin origins (`include`, `prepend`, `extend`).

I was able to bring some knowledge of Ruby, the class/module hierarchy, and how symbols are resolved. I brought almost no knowlege of kotlin or the Intellij SDK and thus leaned on a variety of LLMs to do most of the real work.

In addition to adding files in `kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/handlers/ruby`, there is a small change adding a `via` attribute to `LanguageHandler.kt` to accomodate ruby's messy hierarchies, noting whether a path is derived via superclass, module inclusion, module extension, or module prepending.

Its default value of `superclass`
should allow existing language handlers to remain unchanged.

Ruby differs from other supported languages in that the call/superclass hierarchy is a mess

1. **No single-inheritance tree.** Ruby's method lookup is an MRO (method resolution order) built from `include` / `prepend` / `extend` mixins layered around the classic superclass chain. A "parent" in a hierarchy is not necessarily a superclass — it can be a module mixed in three different ways, each of which lands in a *different position* in the lookup order.

2. **FQN resolution is ambiguous.** Open classes, re-opened namespaces, and dynamic definitions mean no single API reliably returns a fully-qualified name. Resolution requires a layered strategy with fallbacks of decreasing fidelity.

3. **How the hierarchy is constructed is important.** The order in which mixins are included, prepended, or extended affects method resolution and must be accurately represented in the hierarchy. To accomodate this, I added a a new `via` attribute to `LanguageHandler.kt`. It defaults to `superclass` so no existing code has to change.

The Ruby plugin is not bundled and not open. It ships only in RubyMine / IntelliJ Ultimate. We cannot link against its PSI (`RClass`, `RModule`, `RMethod`, `RCallExpression`) at compile time, and it may be entirely absent at runtime. **All Ruby PSI access is therefore via reflection.**

- There are a handful of unit tests that run in any environment. They're not...worthless, but they're unable to tell much of a story given that everything is closed.
- The platforms tests will be skipped when run in the absense of the ruby plugin.

To acutally run the platform tests, you need an installed RubyMine or Intellij Ultimate with the ruby plugin installed. There also has to be a match in which version of kotlin was used to build each, but I don't understand all the ins-and-outs of that.

Modify `scripts/ruby-plugin-tests` to point at the location of the ruby plugin and the correct version numbering and run `ruby-plugin-tests enable` to modify `build.gradle.kts` so the platform tests can be run. `ruby-plugin-tests disable` backs out those changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,14 @@
 
 ## [Unreleased]
 
+### Added
+- **Support for ruby plugin, if available**. Adds full support for all `ide_` calls for ruby, 
+including attempting to fully deal with subclassing and module inclusion/extension/prepending.
+
+### Changed
+
+- added attribute `via` (default value: `superclass`) to LanguageHandler to indicate whether a path is derived via superclass, module inclusion, module extension, or module prepending.
+
 ## [4.30.0] - 2026-07-19
 ### Added
 - **`ide_edit_member`** — replace an entire class member declaration (signature + body) by structural name, not text match. Targets by file, class, and member name with optional overload disambiguation. Auto-reformats after editing. Supports Java and Kotlin. *(disabled by default)*
--- a/scripts/ruby-plugin-tests
+++ b/scripts/ruby-plugin-tests
@@ -1,0 +1,181 @@
+#!/usr/bin/env bash
+# ruby-plugin-tests — toggle the Ruby plugin on/off for the Gradle build
+#
+# Usage:  scripts/ruby-plugin-tests enable
+#         scripts/ruby-plugin-tests disable
+#
+# Both commands are idempotent.
+#
+# What it manages (all in the project directory):
+#   build.gradle.kts    — injects / removes the localPlugin() block
+#   ./gradle.properties — localRubyPluginPath + platformVersion
+#
+# IMPORTANT: localRubyPluginPath must point at the Ruby plugin ROOT DIRECTORY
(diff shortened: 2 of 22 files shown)


billdueber:
In addition to the platform tests (which will never run in CI unless/until there's a ruby plugin there), I tested the ruby code by going through the plugin via MCP using a set of test procedure run by agents. See https://github.com/billdueber/jetbrains-index-mcp-plugin/tree/feat/ruby/ruby if you're interested.

hechtcarmel:
Hi,
I hope to check it out over the weekend. 
Likely mostly with AI since Ruby is not my thing 😅 
Thank you for the contribution!

hechtcarmel:
also ci failed :) Please fix

billdueber:
I have no idea what the heck happened with this -- I rebased off the wrong branch? Witches? In any case going to resubmit.

PR #258, opened 2026-07-24
Add first-class ruby support when ruby plugin is installed

billdueber:
## Summary

Adds full Ruby support to the MCP navigation tools (`ide_type_hierarchy`, `ide_call_hierarchy`, `ide_find_super_methods`, `ide_file_structure`, `ide_find_symbol`) via a new `handlers/ruby/` handler family, plus the plumbing needed to run against a **closed-source, non-bundled** Ruby plugin and to report **method-resolution provenance** back to callers.

## Features

Added ruby handlers for all the language-specific calls.

- `ide_call_hierarchy` (`RubyCallHierarchyHandler`)
- `ide_type_hierarchy` (`RubyTypeHierarchyHandler`)
- `ide_find_super_methods` (`RubySuperMethodsHandler`)
- `ide_file_structure` (`RubyStructureHandler`)
- `ide_resolve_symbol` (`RubySymbolReferenceHandler`)

I was able to bring some knowledge of Ruby, the class/module hierarchy, and how symbols are resolved. I brought almost no knowlege of kotlin or the Intellij SDK and thus leaned on a variety of LLMs to do most of the real work.

## Model/contract updates

`SuperMethodData`  in  `LanguageHandler.kt` now includes `via` provenance (default `"superclass"`) so super-method edges can distinguish `superclass` vs Ruby `mixin` origins (`include`, `prepend`, `extend`). Using `superclass` as the default value ensures the change  doesn't affect existing code at all.

## Ruby's call / superclass hierarchy

Ruby differs from other supported languages in that the call/superclass hierarchy is a mess

1. **No single-inheritance tree.** Ruby's method lookup is an MRO (method resolution order) built from `include` / `prepend` / `extend` mixins layered around the classic superclass chain. A "parent" in a hierarchy is not necessarily a superclass — it can be a module mixed in three different ways, each of which lands in a *different position* in the lookup order.

2. **FQN resolution is ambiguous.** Open classes, re-opened namespaces, and dynamic definitions mean no single API reliably returns a fully-qualified name. Resolution requires a layered strategy with fallbacks of decreasing fidelity.

3. **How the hierarchy is constructed is important.** The order in which mixins are included, prepended, or extended affects method resolution and must be accurately represented in the hierarchy. To accomodate this, I added a a new `via` attribute to `LanguageHandler.kt`. It defaults to `superclass` so no existing code has to change.

The Ruby plugin is not bundled and not open. It ships only in RubyMine / IntelliJ Ultimate. We cannot link against its PSI (`RClass`, `RModule`, `RMethod`, `RCallExpression`) at compile time, and it may be entirely absent at runtime. **All Ruby PSI access is therefore via reflection.**

## Running the platform tests

- There are a handful of unit tests that run in any environment. They're not...worthless, but they're unable to tell much of a story given that everything is closed.
- The platforms tests will be skipped when run in the absence of the ruby plugin.
- To acutally run the platform tests, you need an installed RubyMine or Intellij Ultimate with the ruby plugin installed. There also has to be a match in which version of kotlin was used to build each, but I don't understand all the ins-and-outs of that.
- Modify `scripts/ruby-plugin-tests` to point at the location of the ruby plugin and the correct version numbering and run `ruby-plugin-tests enable` to modify `build.gradle.kts` so the platform tests can be run. `ruby-plugin-tests disable` backs out those changes.

### DIFF
--- a/scripts/ruby-plugin-tests
+++ b/scripts/ruby-plugin-tests
@@ -1,0 +1,181 @@
+#!/usr/bin/env bash
+# ruby-plugin-tests — toggle the Ruby plugin on/off for the Gradle build
+#
+# Usage:  scripts/ruby-plugin-tests enable
+#         scripts/ruby-plugin-tests disable
+#
+# Both commands are idempotent.
+#
+# What it manages (all in the project directory):
+#   build.gradle.kts    — injects / removes the localPlugin() block
+#   ./gradle.properties — localRubyPluginPath + platformVersion
+#
+# IMPORTANT: localRubyPluginPath must point at the Ruby plugin ROOT DIRECTORY
+# (the one containing lib/ and lib/modules/), NOT a single jar. The plugin's PSI
+# implementation classes (RClass, RubyOverrideImplementUtil, …) live in
+# lib/modules/intellij.ruby.*.jar — pointing at lib/ruby.jar loads only
+# META-INF/plugin.xml, so Class.forName() fails and the Ruby handlers report
+# isAvailable=false, causing the platform tests to silently SKIP (green, 0ms).
+set -euo pipefail
+
+# ─── Configuration ─────────────────────────────────────────────
+# Ruby plugin root directory (contains lib/, lib/modules/, rubystubs/, …).
+RUBY_PLUGIN_DIR="$HOME/Library/Application Support/JetBrains/IntelliJIdea2025.3/plugins/ruby"
+IDEA_VERSION="2025.3"
+# ───────────────────────────────────────────────────────────────
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+PROJECT_DIR="$(cd "$SCRIPT_DIR/.." && pwd)"
+
+LOCAL_PROPS="$PROJECT_DIR/gradle.properties"
+GRADLE_KTS="$PROJECT_DIR/build.gradle.kts"
+BAK_PROPS="$LOCAL_PROPS.ruby-plugin-tests-bak"
+
+MARKER_BEGIN="// BEGIN: ruby-plugin-tests"
+MARKER_END="// END: ruby-plugin-tests"
+
+# ─── build.gradle.kts: inject / remove the localPlugin block ──
+
+has_block() {
+    grep -qF "$MARKER_BEGIN" "$GRADLE_KTS" 2>/dev/null
+}
+
+inject_block() {
+    if has_block; then
+        echo "  ~  ruby block already present in build.gradle.kts"
+        return
+    fi
+    local tmp="$GRADLE_KTS.ruby-tmp"
+    awk -v pad="        " '
+        { print }
+        /^        plugins\(providers\.gradleProperty\("platformPlugins"\)/ {
+            print ""
+            print pad "// BEGIN: ruby-plugin-tests"
+            print pad "// Local Ruby plugin for Ruby handler tests"
+            print pad "providers.gradleProperty(\"localRubyPluginPath\").orNull?.let { path ->"
+            print pad "    localPlugin(file(path))"
+            print pad "    bundledModules("
+            print pad "        \"com.intellij.modules.ultimate\","
+            print pad "        \"com.intellij.modules.json\","
+            print pad "        \"org.jetbrains.plugins.yaml\""
+            print pad "    )"
+            print pad "}"
+            print pad "// END: ruby-plugin-tests"
+        }
+    ' "$GRADLE_KTS" > "$tmp"
+    mv "$tmp" "$GRADLE_KTS"
+    echo "  ✓  injected ruby block into build.gradle.kts"
+}
+
+remove_block() {
+    if ! has_block; then
+        echo "  ~  ruby block not present in build.gradle.kts"
+        return
+    fi
+    local tmp="$GRADLE_KTS.ruby-tmp"
+    awk -v begin="$MARKER_BEGIN" -v end="$MARKER_END" '
+        skip && /^$/ { next }
+        $0 ~ begin { skip=1; next }
+        $0 ~ end   { skip=0; next }
+        !skip { print }
+    ' "$GRADLE_KTS" > "$tmp"
+    mv "$tmp" "$GRADLE_KTS"
+    echo "  ✓  removed ruby block from build.gradle.kts"
+}
+
+# ─── gradle.properties: localRubyPluginPath ────────────────────
+
+has_prop() {
+    grep -q '^localRubyPluginPath=' "$LOCAL_PROPS" 2>/dev/null
+}
+
+set_prop() {
+    if has_prop; then
+        if grep -qF "localRubyPluginPath=$RUBY_PLUGIN_DIR" "$LOCAL_PROPS"; then
+            echo "  ~  localRubyPluginPath already set in $LOCAL_PROPS"
+            return
+        fi
+        local tmp="$LOCAL_PROPS.tmp"
+        grep -v '^localRubyPluginPath=' "$LOCAL_PROPS" > "$tmp" || true
+        printf 'localRubyPluginPath=%s\n' "$RUBY_PLUGIN_DIR" >> "$tmp"
+        mv "$tmp" "$LOCAL_PROPS"
+        echo "  ✓  updated localRubyPluginPath in $LOCAL_PROPS"
+    else
+        printf 'localRubyPluginPath=%s\n' "$RUBY_PLUGIN_DIR" >> "$LOCAL_PROPS"
+        echo "  ✓  added localRubyPluginPath to $LOCAL_PROPS"
+    fi
+}
+
+unset_prop() {
+    if ! has_prop; then
+        echo "  ~  localRubyPluginPath not present in $LOCAL_PROPS"
+        return
+    fi
+    local tmp="$LOCAL_PROPS.tmp"
+    grep -v '^localRubyPluginPath=' "$LOCAL_PROPS" > "$tmp" || true
+    sed -i '' '/^$/d' "$tmp"
+    mv "$tmp" "$LOCAL_PROPS"
+    echo "  ✓  removed localRubyPluginPath from $LOCAL_PROPS"
+}
+
+# ─── gradle.properties: platformVersion ────────────────────────
+
+# Back up gradle.properties once, before the first mutation.
+maybe_backup() {
+    if [ ! -f "$BAK_PROPS" ] && [ -f "$LOCAL_PROPS" ]; then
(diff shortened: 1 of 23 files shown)


hechtcarmel:
First off — thank you for this.  
I don't know Ruby, but multiple people wanted it in.

Since i don't know Ruby well, i delegated the review to Claude Fable: 

The handler architecture follows the existing reflection pattern correctly, threading is right everywhere I traced it, and the `via` (include/prepend/extend) model of Ruby's MRO is genuinely well thought out. I I checked out the branch and went through it in depth, and there are a few things that have to be fixed first — one of them makes the feature not work at all outside the test environment.

## The big one: the Ruby classes can't be loaded in a real IDE

The PR doesn't touch `plugin.xml`. In the IntelliJ Platform, a plugin's classloader can only see classes from plugins it *declares* as dependencies — that's why every other language here has an entry like:

```xml
<depends optional="true" config-file="rust-features.xml">com.jetbrains.rust</depends>
```

(`rust-features.xml` is an empty placeholder; the whole point of the pair is creating the classloader edge so `Class.forName` on that plugin's classes can resolve.)

There's no such entry for `org.jetbrains.plugins.ruby` and no `ruby-features.xml`. So in a real RubyMine / IU install: `PluginDetectors.ruby.isAvailable` returns true (it's just a registry lookup), then `Class.forName("...RClass")` in `RubyHandlers.register()` throws `ClassNotFoundException`, the catch logs "Ruby PSI classes not found, skipping registration", and none of the handlers ever register. Every tool answers "No ... handler available for language: ruby".

It works in your test setup because `scripts/ruby-plugin-tests` injects the Ruby plugin via Gradle `localPlugin(...)`, which puts the jars on the flat test classpath — so `Class.forName` succeeds there. Classic green-in-tests, dead-in-prod.

The fix is small: add the `<depends optional="true" config-file="ruby-features.xml">org.jetbrains.plugins.ruby</depends>` line plus an empty `ruby-features.xml`, mirroring the rust/go/php ones — and then please do one `runIde` smoke test against an IDE with the Ruby plugin to confirm registration actually happens. (The plugin verifier can't catch this — reflection is invisible to it — which is why CI is green.)

## check-pr.sh fails on this branch

`./scripts/check-pr.sh` is the mandatory pre-push gate (see CONTRIBUTING.md) and it exits 1 here:

```
✗ 3 new catch blocks return null — verify each distinguishes 'unavailable'
  from 'broken' (CONTRIBUTING.md § Error handling)
```

The flagged spots are `BaseRubyHandler.kt:705` and `RubySymbolReferenceHandler.kt:265` / `:283`. Related and more important: several of the broad `catch (_: Exception)` blocks (e.g. `RubyCallHierarchyHandler.kt:231-251`, `RubyStructureHandler.kt:98`) sit inside loops under a cancellable read action and swallow `ProcessCanceledException` (sometimes wrapped in `InvocationTargetException`). PCE must be rethrown — see how `ListTestsTool.kt:140` handles it.

## The platform tests report fake green in CI

All six `BasePlatformTestCase` classes gate with:

```kotlin
override fun shouldRunTest(): Boolean =
    PluginDetectors.ruby.isAvailable && super.shouldRunTest()
```

CI runs `./gradlew check` with no Ruby plugin, so all ~100 platform test methods execute there — and with `shouldRunTest() == false`, JUnit3's `runBare()` just returns early. Every method shows **PASSED (0ms)** having run nothing. That's the "silent skip" pattern CONTRIBUTING.md § Test honesty explicitly forbids, moved into a framework hook. (The `Assume.assumeTrue` calls in `requireRubyPlugin()` are unreachable — and wouldn't help anyway: in JUnit3-style tests an `AssumptionViolatedException` reports as a *failure*, not a skip.)

And it's not theoretical — the silent skip is already hiding a broken test. `RubySymbolReferenceHandlerTest.testResolveSymbolWithLeadingWhitespace` (line ~304) has a fixture of `class User; end` with **no methods**, then asserts:

```kotlin
val result = handler.resolveSymbol(project, "  User#find_by_email")
assertTrue("Should trim whitespace before resolution", result.isSuccess)
```

`resolveMethodInClass` returns failure for a missing method — and `testResolveNonExistentMethod` in the same file, with a structurally identical fixture, asserts `isFailure`. Both can't pass against the same implementation, so this test is red on any machine where it actually runs. There's also a literal `if (hierarchy.supertypes.isEmpty()) return` skip in `RubyTypeHierarchyPlatformTest.kt:404` whose assertion is unreachable everywhere (the light fixture never has a Ruby SDK).

Honest options: exclude the Ruby platform tests from the `test` task via a Gradle filter keyed on `localRubyPluginPath` being set (visibly "not run" instead of fake-passed), or move these classes to JUnit4 style where assumptions report as skipped. Either way, the two tests above need to be fixed or deleted.

## The `via` field never reaches MCP clients

`RubySuperMethodsHandler` carefully populates `SuperMethodData.via` — but `FindSuperMethodsTool` (lines ~90-102) maps every field *except* `via` into `SuperMethodInfo`, which doesn't have the field. So the headline feature of the PR — distinguishing superclass vs include vs prepend vs extend — is silently discarded before serialization; callers see a flat hierarchy. Needs `via` added to `SuperMethodInfo` + the mapping + a line in the tools reference (or, if you decide it's not worth surfacing, drop the field).

## Correctness issues in the handlers

- **`class << self` is misclassified.** `RubySuperMethodsHandler.kt:69` detects class methods with `rMethod.text.startsWith("def self.")` — so methods inside a `class << self` block are treated as instance methods and walked through the wrong chain with wrong `via` values. RubyMine has `RSingletonMethod` for this, reachable with the same reflection pattern you already use.
- **Instance/class methods cross-contaminate.** Ancestor matching (`findMethodByNameReflectively`) and symbol resolution (`resolveMethodInClass`) match by name only. `User#find` and `User.find` resolve to the same arbitrary element when a class defines both `def find` and `def self.find` — very common in Ruby. The resolved method's kind needs to be checked against the `#`/`.` separator.
- **The documented ancestor walk doesn't exist.** `RubySymbolReferenceHandler`'s KDoc promises "superclass + included modules" resolution and the error message says "Ensure the method is defined in the class or its ancestors" — but `resolveMethodInClass` only checks the class's own methods, so `User#save` inherited from a base class won't resolve. `BaseRubyHandler` already has everything needed (`rClassGetSuperClassFQN`, `getIncludedModuleFQNs`, `resolveByFQNRelative`).
- **"Ambiguous" early-return blocks the library fallback.** `RubySymbolReferenceHandler.kt:161-176`: if the project contains any same-short-name class (say the project has `Foo::Base` and you ask for `ActiveRecord::Base`), it hard-fails with "ambiguous" and never tries allScope where the gem class lives. Also, one non-matching candidate isn't "ambiguous" — report candidates only after all scopes fail.
- **Cycle guard doesn't seed the origin.** `collectSuperclassChain` (`RubySuperMethodsHandler.kt:219`) starts with an empty visited set, so a cyclic hierarchy (A→B→A in PSI) reports a method as its own super method. Your own `RubyTypeHierarchyHandler.getSupertypes` does this correctly — seed the start class before the loop. One-line fix.
- **The parameter/return-type extraction can never match.** `RubySuperMethodsHandler.kt:309-380` looks for `RCall` nodes with commands like `"restarg"`, `"kwsplat"`, `"forwarding_arg"` and calls `RMethod.getReturnType()` — those are whitequark-parser/Prism AST names, not RubyMine PSI API (params are `RArgumentList`/`RArgument` children; `RMethod` has no `getReturnType`). Parameters always come out empty and every signature degrades to the bare name (which is why the test only asserting `signature.isNotBlank()` stays green). Either wire it to the real argument API or delete the ~90 lines and return the name honestly.
- Smaller ones: the text-scan fallback regex `[A-Z][A-Za-z_:]*` can't match digit-bearing names (`include Base64`, `OAuth2` extract nothing — worth a failing test first, then `[A-Za-z0-9_:]*`); `class << self` blocks are dropped entirely from `ide_file_structure`; the structure signature regex can pick up `(...)` from the method body across newlines; `languageId = "Ruby"` should be the actual language id `"ruby"` (lookups currently only succeed through fallback paths); `PluginDetectors.ruby.fallbackClass` points at a class that doesn't exist (`org.jetbrains.plugins.ruby.psi.rubyFile` — should be `...ruby.lang.psi.RFile`); unused `rCallExpressionClass`/`rubyIncludedExtendedFQNIndexClass` and a couple of stale KDoc references to files that aren't in the repo (`ruby/research/...`, `ruby/agent_tests/...`).

## CHANGELOG + docs

There's no `CHANGELOG.md` entry under `[Unreleased]` (required for user-visible changes), and no doc mentions Ruby yet. The per-language lists live in: README.md (tool table + the `<!-- Plugin description -->` block, which becomes the Marketplace listing), USAGE.md (incl. the "Supported languages" line for `language`+`symbol`), CLAUDE.md, `src/main/resources/skill/ide-index-mcp/references/tools-reference.md`, and the hardcoded description strings in `TypeHierarchyTool` / `CallHierarchyTool` / `FindSuperMethodsTool` / `FileStructureTool` / `McpConstants.SERVER_DESCRIPTION`. Only the runtime error messages are registry-driven.

## The script

Appreciate the effort to make the closed-plugin tests runnable at all. As-is though it's macOS-only (`sed -i ''`, hardcoded `~/Library/.../IntelliJIdea2025.3/plugins/ruby`), the `\s` in the sed patterns is treated as a literal `s` by BSD sed so the platformVersion switching is a silent no-op, `sed '/^$/d'` strips every blank line from the tracked `gradle.properties`, and nothing references the script — while the test skip messages tell people to set `localRubyPluginPath`, which the committed build doesn't consume (two files also name it `localIdeaPath`). I'd rather we commit a permanent, conditional `localRubyPluginPath` block directly into `build.gradle.kts` (no-op when the property is unset), document it in CONTRIBUTING.md, and drop the mutating script.

## Summary

Must-fix before merge:

- [ ] `plugin.xml` optional depends on `org.jetbrains.plugins.ruby` + `ruby-features.xml`, verified once via `runIde`
- [ ] `./scripts/check-pr.sh` passes; PCE rethrown in the broad catches
- [ ] Honest platform-test skipping (Gradle exclusion or JUnit4), and fix/delete the two broken tests
- [ ] `via` surfaced end-to-end or removed
- [ ] CHANGELOG entry + the docs/description-string sweep

Strongly recommended in this PR (correctness): `RSingletonMethod`-based detection + kind-aware matching, the ancestor walk (or fixed docs/error text), the ambiguity-ordering fix, the visited-set seed, and removing the dead parameter-extraction code.

billdueber:
Wow! Your AI is much smarter than my AI! 😄 

The `via` thing is a red herring -- it's for internal use only when trying to trace the superclass.  The rest of it is, I'm assuming, legit. 

I'm actually off on vacation for a couple weeks, but I'll take a look when I can. 

Thanks for your patience!

Issue #31, opened 2025-12-15
Make port fixed value


alexheadliner:
Could you please make port fixed\configurable value, because It regenerates on every JetBrains restart, annoing to reconfigure it manually each time

hechtcarmel:
Hi, 
In short - yes. This is being looked into and I will update. Hopefully in about a week (+ jetbrains approval time).
(currently it uses the built-in server, so this requires a not-minor refactor).

Note that the ports should stay the same, even between restarts, unless you use multiple ides (e.g. intelliJ + pycharm), at which case they will race to grab the ports.

hechtcarmel:
Sent to jetbrain's approval.

PR #263, opened 2026-07-26
fix: stop tools from confidently reporting incomplete or false results

hechtcarmel:
Fixes **every code defect** from the independent 30-tool field test of the MCP server (the "blindspot report"), plus **eight adjacent defects** of the same class found by a follow-up code audit. Each fix was implemented from a root-cause investigation verified against the bundled 2025.3 platform bytecode — not from the symptom description alone.

The report's own summary named the common thread: *"confidently reporting an incomplete or empty result as complete."* That is what this PR eliminates.

## Report defects (all 7)

| # | Tool | Root cause | Fix |
|---|---|---|---|
| 1 | `ide_find_file` | Name enumeration capped at 1500 names over a **scope-blind** index stream — `FilenameIndex` enumerates library/JDK keys and never consults the scope during key enumeration, so broad queries (`Dao`, `*.java`, `*.properties`) filled the window with out-of-scope names in stable hash order and returned 4-of-26 / 0-of-~110 with `hasMore:false` | Uncapped enumeration + best-match-first resolution, mirroring the IDE's own Goto File; `ProcessCanceledException` no longer swallowed as silent truncation |
| 2 | `ide_diagnostics` | With the daemon inactive (Power Save Mode — which the plugin's own lifecycle service enables), the open-editor path polls an empty highlight set that is indistinguishable from "ran and found nothing", and stamped it `analysisFresh:true` | PSM routes open files to batch analysis; an empty daemon result not provably from a completed pass (daemon terminal event) falls back to batch, with "daemon did not run" kept distinct from "timed out" |
| 3 | `ide_refactor_rename` `targetType:"file"` | The raw `PsiFile` goes to `RenameProcessor`, and no platform processor couples a Java file to its class in the file→class direction | When the top-level class name matches the filename, the rename retargets to the class — which renames the file and updates all references. Non-matching and non-Java files keep plain file rename (Android resources unaffected) |
| 4 | `ide_find_symbol` | IntelliJ's Go to Symbol **suppresses a method when any super signature matches the same unqualified query** (`DefaultSymbolNavigationContributor.hasSuperMethod`) — that's why the interface method returned and the implementation vanished | Suppressed hierarchy methods complemented via `PsiShortNamesCache.processMethodsWithName`, deduped on the existing coordinate key |
| 5 | `ide_find_class` | `kind` inferred from the PSI implementation class *name* — `PsiClassImpl` for every Java declaration form, so `contains("class")` always won | Semantic probes (`isAnnotationType`/`isRecord`/`isEnum`/`isInterface`/`isTrait` via reflection — works for Java, Kotlin, PHP, JS/TS) with the name heuristic as fallback |
| 6 | `ide_find_references` | Position on a comment snaps unbounded up the PSI tree to the enclosing named element, with no echo; `totalCount` reported the 500 cap as the total | New `resolvedSymbol` field (name, kind, container, file, line) on every page; new `totalIsExact` flag; description documents the snapping |
| 7 | `ide_project_status` | Zeroed the managed list whenever `lifecycleEnabled` was off, while `ide_get_project_modes` read the persisted registry — same service, two definitions of "managed" | Reports persisted enrollment unconditionally + `lifecycle_enabled` flag + explanatory note when automation is off |

## Audit defects (8 more, same failure class)

- **`ide_find_implementations`** — every language handler capped results at 100 with `hasMore:false` presented as completeness (the cursor has no extender, so "cache = complete set" must actually hold). Caps raised to the pagination cache bound (5000).
- **`ide_refactor_safe_delete`** — two lies: a stale PSI element silently *skipped the delete* but reported "Successfully deleted"; and `catch (e: Exception) { /* assume there are none */ }` around the usage search let a failed search delete "safely". Both now explicit errors; dumb-mode/cancellation propagate to the standard retry handling.
- **`ide_change_signature`** — a silently aborted refactoring (read-only file; conflicts are auto-accepted headlessly) reported success. A conservative post-run check (SmartPointer re-resolve) errors only when *nothing* of the requested change applied.
- **`ide_move_file`** — `showConflicts` discarded detected conflicts; now surfaced via the existing `warnings` field, sanitized.
- **`ide_replace_text_in_file`** — now returns the `affectedLines` its description had always promised (tracked in new-text coordinates through both literal and regex paths).
- **`ide_insert_member`** — insertion offset re-validated inside the write action (validity + bounds), matching its sibling tools.
- **`ide_read_file`** — beyond-EOF `startLine` errors naming the real line count; `endLine` echoes the effective clamped value instead of parroting the request.
- **Conflict messages** — raw IDE-dialog HTML (`<b><code>`, XML entities) stripped via a shared sanitizer.
- **SKILL.md** — `ide_search_text` described as "exact word occurrences"; it is substring + regex.

## Not fixed here (deferred deliberately)

`ide_type_hierarchy` (100-subtype cap), `ide_call_hierarchy` (20-per-level cap), and `ide_find_super_methods` (10-super cap) silently truncate too — but the honest fix spans six language handler files and three result models each. Deferred to a follow-up; a task chip exists for it.

## Contract impact — reviewed line by line

`result-shapes.txt`: `FindUsagesResult` +`resolvedSymbol: object? (optional)`, +`totalIsExact: boolean (optional)`; `ReplaceTextResult` +`affectedLines: array? (optional)`. All additive, all optional — no breaking change for clients.
`tool-manifest.json`: one description extension (`ide_find_references`, documenting the new fields and the snapping behaviour). Every other tool byte-identical.

## Verification

- **918 tests (50 new), all green**; `check-pr.sh` 15/15.
- Key new tests are mutation-verified: the `find_file` completeness test fails with the old cap restored (revert false-pass odds ≈ 7e-5 by fixture design); the diagnostics tests fail with the exact false-clean inversion on revert; the change-signature abort test was verified to discriminate by temporarily forcing the pre-fix behaviour; SafeDelete/Move/Rename tests assert both directions on the saved files.
- Root causes verified by disassembling the bundled platform (`FileBasedIndexEx.processAllKeys` scope-blindness, `DefaultSymbolNavigationContributor.hasSuperMethod` suppression, `DaemonCodeAnalyzerImpl.runUpdate` PSM early-return, `RenamePsiFileProcessor` non-substitution) rather than inferred from symptoms.


hechtcarmel:
Added: `ide_create_file` no longer triggers the VCS "Add File to Git" confirmation (the app-modal dialog that froze the EDT and hung all in-flight MCP calls — and, in its silent setting, the stealth staging from report finding #31). Creation now goes disk-write + synchronous VFS refresh, which the VCS listener ignores by design. Mutation-verified test pins the mechanism.

PR #268, opened 2026-07-28
feat: add ide_create_module — generic module creation for any directory

mdproctor:
## Summary

- Adds `ide_create_module` — registers a directory as an IntelliJ module with a content root, enabling code intelligence for non-Maven projects (TypeScript, plain directories, git worktrees)
- Complements `ide_import_modules` (Maven-only) and `ide_open_workspace` (Maven workspace aggregator)
- Supports optional directory exclusions (e.g., `node_modules`, `dist`, `.next`)

## Why

When agents work in git worktrees or non-Maven projects, IntelliJ tools like `ide_find_references` and `ide_diagnostics` don't cover the files because the directory isn't registered as a content root. This tool fills that gap without requiring Maven.

## Implementation

- `CreateModuleTool.kt`: creates a `WEB_MODULE` with a content root and optional excluded directories
- Module creation runs inside `WriteAction` on EDT (required for `ModuleManager.newModule` and `rootModel.commit`)
- Guards: non-absolute path, non-existent directory, pom.xml present (directs to `ide_import_modules`), `..` in excludes, already-registered content root (idempotent)
- Disabled by default (schema version 6 migration)
- Updated for MCP Kotlin SDK migration (CallToolResult, ToolSchema)

Supersedes #254 (closed due to stale fork state).

## Checklist

- [x] All 6 doc locations (README, USAGE, CLAUDE.md, SKILL.md, tools-reference.md, CHANGELOG)
- [x] ToolNames constant + ALL list (alphabetical)
- [x] McpSettings: DEFAULT_DISABLED_TOOLS + schema v6 migration
- [x] ToolRegistry registration
- [x] Unit tests: schema, disabled-by-default, migration from v2
- [x] CONTRIBUTING.md code correctness checklist: threading, error handling, sibling consistency

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 # IDE Index MCP Server Changelog
 
 ## [Unreleased]
+### Added
+- **`ide_create_module`** — add a directory as an IntelliJ module with a content root, enabling code intelligence for non-Maven projects (TypeScript, plain directories, etc.). Supports optional directory exclusions (e.g., `node_modules`, `dist`). For Maven projects, use `ide_import_modules` instead. *(disabled by default)*
 
 ## [5.1.0] - 2026-07-30
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -365,7 +365,7 @@ Three tiers, selected by class-name suffix:
      schema into `src/test/resources/contract/tool-manifest.json`. This is the regression net
      for large refactors: one assertion covers every registered tool × every schema property, so
      a dropped `register(...)` call or a mutated parameter type fails here instead of shipping.
-     Scope: 47 of the 50 tools in `ToolNames.ALL` (the three needing the Kotlin or Maven plugin
+     Scope: 48 of the 51 tools in `ToolNames.ALL` (the three needing the Kotlin or Maven plugin
      are covered by set-equality instead), and **inputs only**.
    - `ResultShapeContractUnitTest` snapshots the other half of the client contract — the response
      side — into `src/test/resources/contract/result-shapes.txt`: the wire key set, JSON value
@@ -491,6 +491,7 @@ Tools are organized by IDE availability.
 - `ide_open_file` - Open a file in the editor with optional line/column navigation (disabled by default)
 - `ide_set_power_save_mode` - Enable/disable IDE Power Save Mode (IDE-wide). Suspends background inspections and code analysis while keeping the index and code intelligence operational (disabled by default)
 - `ide_close_project` - Close an open project window and free its memory. Non-blocking; refuses to close the last open project so the MCP server keeps a JSON-RPC context (disabled by default)
+- `ide_create_module` - Add a directory as an IntelliJ module with a content root, enabling code intelligence for non-Maven projects (TypeScript, plain directories, etc.). Supports optional directory exclusions. For Maven projects, use `ide_import_modules` instead. (disabled by default)
 - `ide_open_project` - Open a project by absolute path and wait until indexing completes (`timeoutSeconds`, default 600). Idempotent for already-open projects (disabled by default)
 - `ide_install_plugin` - Install a plugin zip into the IDE, replacing any existing version; auto-detects `build/distributions/*.zip` when no path is given (disabled by default)
 - `ide_restart` - Restart the IDE; terminates the MCP connection. Call after `ide_install_plugin` (disabled by default)
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ Each JetBrains IDE has a unique default port and server name to allow running mu
 
 ## Available Tools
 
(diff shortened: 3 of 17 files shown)


hechtcarmel:
### 1. `name` parsing bypasses the helper that exists for exactly this (required)

`arguments["name"]?.jsonPrimitive?.content` has two holes the blank-check doesn't cover:

- `"name": null` — agents emit explicit nulls for optional params all the time. `JsonNull` *is* a `JsonPrimitive`, so `.content` returns the string `"null"`, it passes validation, and you get a module literally named `null` with a `null.iml` in the user's working tree.
- `"name": " web "` — not blank, so it survives untrimmed: module ` web `, iml ` web .iml`.

`AbstractMcpTool.optionalStringArg` handles both (JsonNull → absent, trims, blank → absent) — it exists because a previous audit fixed this exact class of bug across the other tools. One line:

```kotlin
val moduleName = optionalStringArg(arguments, "name") ?: dir.name
```

The isBlank half of your validation then goes dead — keep the `..`/separator checks. Same story one line down: an `excludes` array containing a `null` element becomes an excluded directory named `"null"`; map through `contentOrNull` (or filter to `JsonPrimitive`) instead of `.content`.

### 2. Invalid excludes are silently swallowed (required)

`if (exclude.isBlank() || exclude.contains("..")) continue` — the agent that passed `excludes: ["../shared"]` gets a success result, and since the "Excluded N directories" line only prints when N > 0, there is no trace anywhere that the exclude was dropped. Same principle as the read-only error in #273: the response text is the only feedback channel an agent has, so a rejected input has to say so. `name` already errors on `..` — give excludes the same treatment and return an error naming the offending entry. (The skip shape came from my #254 nits, so that one's on me — erroring is right.)

### 3. The pom.xml guard is a dead end outside IDEA (required)

The guard redirects to `ide_import_modules`, but that tool only registers when the Maven plugin is present. In WebStorm — the IDE this tool is most for — pointing at a directory that happens to contain a pom.xml (a front-end living inside a Maven mono-repo, say) gets an error recommending a tool that doesn't exist there, with no way forward. Gate it:

```kotlin
if (File(dir, "pom.xml").exists() && PluginDetectors.maven.isAvailable) {
```

Where Maven tooling isn't installed, indexing a Maven-layout directory as a plain content root is exactly what WebStorm does natively — let it.

### 4. The tests from #254 didn't make the rewrite (required this time)

Round one I marked "add tests for the guards" as won't-block, and your "addressed everything" reply skipped it — fair, it was optional. But it's the only round-one item still open, and items 1 and 2 above are precisely the bugs a `doExecute` test would have caught, so this time it blocks: nothing in the suite executes the tool. All of it is cheap in the platform tier next to your `testAllToolsRegistered` change — happy path (module exists, content root registered, `contentEntry.excludeFolderUrls` contains the exclude), second call on the same path returns the idempotent message, name collision errors, pom guard redirects, invalid name errors. `McpPlatformTestCase` plus a temp directory is all it needs, and house rule applies: each test has to fail if you delete the guard it covers.

### 5. Docs drifted from the code during the rewrite (required)

Three accuracy fixes, all one-liners:

- The USAGE.md example response shows two lines, but the tool now always prints four — the `Module file: …` line and the async-indexing note are missing. Paste the real output.
- tools-reference.md's "Returns" has the same gap (no module-file path, no note).
- Adding tool 51 / manifest entry 48 stales two hard-coded counts: "47 of the 50" in CLAUDE.md's Testing section, and the same phrase in `ToolManifestContractUnitTest`'s KDoc. Both become 48 of 51.

mdproctor:
Addressed all 5 items. Force-pushed (rebased on 5.1.0).

**1. Name parsing** — switched to `optionalStringArg(arguments, "name")` which handles JsonNull, trims, blank-as-absent. Excludes now use `mapNotNull { (it as? JsonPrimitive)?.contentOrNull }` to filter null elements.

**2. Invalid excludes** — error instead of silent skip. Returns `createErrorResult` naming the offending entry for both blank and `..` cases. Validation moved to `doExecute` before the write action.

**3. pom.xml guard** — gated on `PluginDetectors.maven.isAvailable`. WebStorm users can now create modules for directories that happen to contain a pom.xml.

**4. Behavior tests** — 7 tests in `CreateModuleBehaviorTest`:
- Happy path (content root registered, exclude URLs present, all 4 response lines)
- Idempotent (second call returns "already registered")
- Name collision (same name, different path → error)
- Invalid name (`..` → error)
- Invalid exclude (`..` → error, blank → error)
- pom.xml guard (Maven available → error)

**5. Docs accuracy** — USAGE.md example response shows all 4 lines. tools-reference.md Returns updated. CLAUDE.md and ToolManifestContractUnitTest counts updated to 48/51. Golden manifest regenerated.

Unit tests pass, check-pr.sh 18/18.

PR #273, opened 2026-07-29
fix: add ensureWritable pre-flight check to all write tools

mdproctor:
## Summary

Adds `ensureWritable(virtualFile, filePath)` pre-flight check to all write tools. Returns an actionable error ("File is read-only: ... Check file permissions or VCS lock status") before the write attempt, preventing the platform from triggering the "Clear read-only status" dialog.

**Note:** You mentioned in #256 that you'd handle this yourself. Happy to close this if you'd prefer to do it your way — filing it in case having the code ready is useful.

## Tools covered

| Tool | Check location |
|------|---------------|
| `ReplaceTextInFileTool` | After `resolveFile`, before write action |
| `EditMemberTool` | After `resolveFile`, before prep |
| `InsertMemberTool` | After `resolveFile`, before prep |
| `ReplaceMemberTool` | After `resolveFile`, before prep |
| `MoveFileTool` | After `resolveFile`, before move execution |
| `ChangeSignatureTool` | After `resolveFile`, before processor |
| `RenameSymbolTool` | After `getPsiFile`, in validation |
| `SafeDeleteTool` | After `getPsiFile`, in preparation |

## Implementation

`ensureWritable` is a protected method on `AbstractMcpTool` returning `CallToolResult?` — null if writable, error result if not. Each tool calls it with `?.let { return it }` after resolving the file.

For `RenameSymbolTool` and `SafeDeleteTool`, the check is inlined because they use `PsiFile` and return sealed-class results rather than `CallToolResult` directly.

## Test plan

- [x] Unit tests pass
- [ ] Manual: attempt to edit a read-only file — should return error, not trigger dialog

Ref: #271

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Write tools now reject read-only files with a clear error** instead of failing deep inside the refactoring engine. Affected tools: `ide_replace_text_in_file`, `ide_edit_member`, `ide_insert_member`, `ide_replace_member`, `ide_move_file`, `ide_change_signature`, `ide_reformat_code`, `ide_optimize_imports`, `ide_refactor_rename` (both symbol and file modes), and `ide_refactor_safe_delete` (both symbol and file modes).
+
 ### Security
 
 - **`ide_install_plugin` rejects archives with zip-slip entries** (`../` traversal or absolute paths) that would write outside the IDE plugins directory. The whole archive is validated before the existing installation is removed, so a rejected archive leaves the current plugin intact, and extraction independently re-checks every normalized destination path.
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/AbstractMcpTool.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/AbstractMcpTool.kt
@@ -436,6 +436,14 @@ withContext(Dispatchers.EDT + ModalityState.nonModal().asContextElement()) { act
         return null
     }
 
+    /** Returns a "read-only" error when [file] is not writable, or `null` if the file is writable. */
+    protected fun ensureWritable(file: VirtualFile): CallToolResult? {
+        if (!file.isWritable) {
+            return createErrorResult("File is read-only and cannot be modified: ${file.path}")
+        }
+        return null
+    }
+
     /**
      * Gets the PSI file for a given path.
      *
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/refactoring/ChangeSignatureTool.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/refactoring/ChangeSignatureTool.kt
@@ -127,6 +127,7 @@ class ChangeSignatureTool : AbstractMcpTool() {
 
         val virtualFile = resolveFile(project, filePath)
             ?: return createErrorResult("File not found: $filePath")
+        ensureWritable(virtualFile)?.let { return it }
 
         val changeSignatureProcessorClass = try {
             Class.forName("com.intellij.refactoring.changeSignature.ChangeSignatureProcessor")
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/refactoring/EditMemberTool.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/refactoring/EditMemberTool.kt
@@ -61,6 +61,7 @@ class EditMemberTool : AbstractMcpTool() {
 
         val virtualFile = resolveFile(project, filePath)
             ?: return createErrorResult("File not found: $filePath")
+        ensureWritable(virtualFile)?.let { return it }
 
         val prep = suspendingReadAction {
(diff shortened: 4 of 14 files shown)


hechtcarmel:
# Review — PR #273: ensureWritable pre-flight check

Thanks for filing this — and no, don't close it. I did say in #256 I'd do it myself, but you've saved me the work and the shape is right: a small always-on helper on `AbstractMcpTool`, called right after file resolution, no setting, no toggle. That's exactly what I asked for. CI is green and the helper itself is fine.

I can't merge it as-is though. The short version: the check landed on the paths where it's merely a nicer error, and missed several of the paths where it's the actual fix. Details below.

### 1. Symbol rename didn't get the check (required)

`RenameSymbolTool` has two validation paths — `validateAndPrepare` for symbol renames and `validateAndPrepareFileRename` for file renames. The guard only went into the file-rename one. Symbol rename is the common case, and it's precisely the processor path (`BaseRefactoringProcessor.ensureElementsWritable`) that raises the "Clear Read-only Status" dialog — the reason this PR exists. Add the same guard to `validateAndPrepare` right after the `getPsiFile` null-check.

### 2. Safe delete: half covered, and the covered half reports the wrong error (required)

Same two-path situation: `prepareSymbolDelete` got the check, `prepareFileDelete` (target_type="file") didn't.

Bigger problem: in `prepareSymbolDelete` the read-only case is returned as `SymbolPreparationResult.FileNotFound("read-only: $file — …")`, which renders to the client as:

```
File not found: read-only: src/Foo.java — check file permissions or VCS lock status
```

The file exists. An agent seeing "File not found" will start second-guessing the path it passed, not the file's permissions — the error text is the whole point of a pre-flight check, so it has to be the right error. Add a `ReadOnly(file)` variant to both sealed classes (`SymbolPreparationResult` and `FilePreparationResult`) and render it with the same message `ensureWritable` produces.

### 3. "All write tools" is eight of twelve (required)

`ReformatCodeTool` and `OptimizeImportsTool` both resolve a single file and run `AbstractLayoutCodeProcessor` subclasses (`ReformatCodeProcessor`, `OptimizeImportsProcessor`) — that family calls `ensureFilesWritable` and raises exactly the dialog we're trying to prevent. They're one-line additions since both already have the `resolveFile` result in hand.

`StructuralSearchReplaceTool` (replace mode) and `ConvertJavaToKotlinTool` also write, but they're multi-file, so the single-file helper doesn't drop in cleanly. I'm fine deferring those two to a follow-up — but then say so in the PR body, and the title's "all write tools" needs to become true one way or the other. (`CreateFileTool` is exempt — the target must not exist, so there's nothing to check.)

### 4. The helper stole `resolveFile`'s KDoc (required)

`ensureWritable` was inserted between `resolveFile`'s doc comment and `resolveFile` itself, so "Resolves a file path to a VirtualFile… @return The VirtualFile, or null if not found" now documents the wrong function and `resolveFile` has none. Move the helper below `resolveFile` (or above the KDoc) and give it its own one-liner.

### 5. Tests (required)

House rule: a behavior change needs a test that fails if the code is deleted. The mechanism is straightforward in the platform tier — `writeProjectFile`, flip the flag with `ReadOnlyAttributeUtil.setReadOnlyAttribute(vf, true)` inside a write action, run the tool, assert the error mentions read-only *and* that the file content didn't change, then restore writability in cleanup so the fixture can be torn down. `RenameSymbolToolBehaviorTest` already has the post-hoc abort test for the same scenario, so there's a pattern to sit next to. One test per distinct code path you touched: the shared helper (any one of the direct-write tools), rename-symbol, rename-file, safe-delete-symbol, safe-delete-file.

### 6. CHANGELOG (required)

One line under `[Unreleased]` → Fixed. It's a user-visible `fix:`, so it owes the changelog an entry. `check-pr.sh` won't catch this one for you — it only checks that the section exists.

### Nit

There are now three phrasings of the same condition (the helper's two-sentence version, the inline rename one, the safe-delete one). Once #2 gives the sealed classes a proper variant, route all of them through one message — clients end up string-matching these, and one string is kinder than three.

---

None of this is structural — it's the same ten lines, applied to the right places, with the error type fixed and a handful of tests. With items 1–6 addressed I'll merge it. And genuinely: thanks for coming back after #256 with the focused version instead of walking away.

mdproctor:
Addressed all 6 points. Force-pushed a rewrite (14 files).

**1. Symbol rename** — added the check to `validateAndPrepare` (symbol path), not just `validateAndPrepareFileRename`.

**2. SafeDelete error type** — added `ReadOnly(val file: String)` to both `SymbolPreparationResult` and `FilePreparationResult` sealed classes. Handled in both `when` blocks. Applied to both `prepareSymbolDelete` and `prepareFileDelete`.

**3. ReformatCodeTool + OptimizeImportsTool** — added `ensureWritable` after `resolveFile` in both. Title now accurately reflects coverage.

**4. KDoc** — `ensureWritable` moved below `resolveFile` with its own one-liner doc.

**5. Tests** — new `ReadOnlyFileGuardBehaviorTest.kt` with 5 tests:
- `testReplaceTextRejectsReadOnlyFile`
- `testRenameSymbolRejectsReadOnlyFile`
- `testRenameFileRejectsReadOnlyFile`
- `testSafeDeleteSymbolRejectsReadOnlyFile`
- `testSafeDeleteFileRejectsReadOnlyFile`

Each: write file, mark read-only, call tool, assert error mentions "read-only", assert file unchanged, restore writability in tearDown.

**6. CHANGELOG** — added under `[Unreleased]` → Fixed.

**Error message** — unified to one phrasing: "File is read-only and cannot be modified: {path}" across all paths.

PR #286, opened 2026-08-03
fix: remove unescapeText that double-processes escape sequences in replaceText

### LINKED ISSUE #269: bug: ide_replace_text_in_file double-unescapes backslashes and newlines in replaceText
## Problem

`ide_replace_text_in_file` interprets escape sequences in `replaceText` via a custom `unescapeText()` function that converts `\n` → literal newline, `\t` → tab, `\\` → single backslash. This causes three recurring issues:

### 1. `\n` in Java/Kotlin string literals becomes a real newline

When replacing text that should contain the Java escape `\n` (e.g., `"header\n" + "body"`), the tool inserts a literal line break inside the string literal, producing a syntax error:

```
// Input replaceText: "header\\n" + "body"
// Expected in file:  "header\n" + "body"
// Actual in file:    "header
// " + "body"          ← broken string literal
```

### 2. Backslashes eaten in Kotlin/regex code

Kotlin strings with `\\` (e.g., `split("/", "\\")`) lose a backslash, producing `split("/", "\")` which is a syntax error. This was encountered multiple times during development of tools in this repo.

### 3. Agents can't reliably do find-and-replace on source code

The workaround is to use `ide_edit_member` instead, but that's a structural edit — overkill for a simple text replacement. Agents shouldn't need to reason about double-escaping.

## Root cause

`ReplaceTextInFileTool.unescapeText()` at line ~150:

```kotlin
private fun unescapeText(text: String): String {
    val sb = StringBuilder(text.length)
    var i = 0
    while (i < text.length) {
        if (text[i] == '\\' && i + 1 < text.length) {
            when (text[i + 1]) {
                'n' -> { sb.append('\n'); i += 2 }
                't' -> { sb.append('\t'); i += 2 }
                '\\' -> { sb.append('\\'); i += 2 }
                else -> { sb.append(text[i]); i++ }
```

This was originally added so agents could insert newlines by writing `\n` in the JSON `replaceText` parameter. But JSON itself already supports `\n` — a JSON string `"line1\nline2"` already contains a literal newline when parsed. The `unescapeText` layer double-processes escapes that JSON already handled.

## Proposed fix

Remove `unescapeText()` entirely. JSON string parsing already handles `\n`, `\t`, and `\\` — the MCP SDK delivers the parsed string to `doExecute()`. The extra unescape layer is redundant and destructive.

If there are agents that depend on the current behavior (sending `\\n` to mean newline), they'll need to switch to `\n` in their JSON — which is what JSON is designed for.

## Affected users

Any agent using `ide_replace_text_in_file` to modify source files containing backslash escapes (`\n`, `\t`, `\\`, regex patterns). The workaround is `ide_edit_member` for structural edits, but that's not always applicable.

mdproctor:
## Summary

Removes the `unescapeText()` function from `ide_replace_text_in_file`. JSON string parsing already handles `\n`, `\t`, and `\\` — the extra unescape layer double-processed these sequences, corrupting source code containing backslash escapes.

## The bug

When `replaceText` contains Java/Kotlin escape sequences (e.g., `"hello\nworld"`), `unescapeText` converts the literal `\n` into a real newline character, breaking string literals:

```java
// Before replacement: String msg = "hello";
// replaceText: "hello\nworld"
// Expected:    String msg = "hello\nworld";
// Actual:      String msg = "hello
//              world";  ← broken string literal
```

Same issue with backslashes: `"C:\\Users\\test"` loses backslashes.

## TDD

Two failing tests written first, verified they fail on unmodified main, then the fix applied:

- `testReplacePreservesBackslashEscapesInReplaceText` — `\n` in replacement stays as literal `\n`
- `testReplacePreservesBackslashesInReplaceText` — `\\` in replacement stays as `\\`

Three existing tests that relied on the old escape behavior updated to use JSON's native escape handling.

## Breaking change

Agents that sent `\\n` to insert a newline must now send an actual newline in the JSON string (`\n` in JSON). This matches how every other tool handles text parameters, and how JSON is designed to work.

## Test plan

- [x] Two new tests fail on main, pass with fix
- [x] All existing tests updated and pass
- [x] Golden manifest regenerated (description updated)
- [x] Unit tests pass

Closes #269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,11 @@
 
 ### Fixed
 
-- **`ide_create_file`** — `project_path` outside every known project root or content root now returns a clear error instead of silently falling through to `basePath` and creating the file in the wrong location. Traversal paths (`../..`) are also rejected.
+- **`ide_replace_text_in_file`** — removed the `unescapeText` layer that double-processed escape sequences in `replaceText`. JSON string parsing already handles `\n`, `\t`, and `\\`; the extra unescape converted Java/Kotlin string literals like `"\n"` into real newlines and ate backslashes in paths and regex patterns. Replacement text is now passed through as-is.
+
+### Breaking
+
+- **`ide_replace_text_in_file` `replaceText` no longer interprets `\n`/`\t`/`\\` escape sequences.** Agents that relied on sending `\\n` to insert a newline must now send an actual newline character in the JSON string (which JSON already supports as `\n`). This matches how every other tool handles text parameters.
 
 ## [5.2.0] - 2026-07-31
 
--- a/USAGE.md
+++ b/USAGE.md
@@ -2088,7 +2088,7 @@ Use this for mechanical text substitutions — e.g., replacing a method call wra
 |-----------|------|----------|-------------|
 | `file` | string | Yes | Path to the file relative to project root |
 | `searchText` | string | Yes | Text to find. Treated as literal unless `regex` is true |
-| `replaceText` | string | Yes | Replacement text. The escape sequences `\n`, `\t`, and `\\` are interpreted as newline, tab, and a single literal backslash — double each backslash that must stay literal (e.g. Windows paths). Supports regex group references (`$1`, `$2`) when `regex` is true |
+| `replaceText` | string | Yes | Replacement text. Passed through as-is (no escape processing). Supports regex group references (`$1`, `$2`) when `regex` is true |
 | `regex` | boolean | No | Treat `searchText` as a regular expression (default: false) |
 | `caseSensitive` | boolean | No | Case-sensitive matching (default: true) |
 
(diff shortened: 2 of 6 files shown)


hechtcarmel:
### 1. Fix changelog conflict

### 2. USAGE.md still teaches the old convention (required)

The `replaceText` row in the parameter table (USAGE.md, ~line 2091) still says the escape sequences `\n`/`\t`/`\\` are interpreted and to "double each backslash that must stay literal". After this PR that's instructions for corrupting your file. Bring it in line with the new schema text.

### 3. So does the bundled skill doc (required)

`src/main/resources/skill/ide-index-mcp/references/tools-reference.md:439` has the same row, word for word — and that file ships inside the plugin zip and gets installed as an agent skill, so agents would be reading the old contract straight out of the new build. Same one-line fix as USAGE.md. Grep for "double each backslash" once you're done to confirm those were the only two — the 5.1.0 changelog entry that documented the old convention stays as-is, that's history.

mdproctor:
Addressed all 3 items. Rebased on current main (5.2.0).

1. **CHANGELOG conflict** — resolved, our entries under `[Unreleased]`.
2. **USAGE.md** — `replaceText` row updated to "Passed through as-is (no escape processing)."
3. **tools-reference.md** — same fix. Confirmed with grep — no other "double each backslash" occurrences remain.

hechtcarmel:
Merged, thanks.
Will release in next version soon

PR #287, opened 2026-08-03
fix: edtAction uses ModalityState.any() so tools don't block behind modal dialogs

mdproctor:
## Summary

`edtAction` in `AbstractMcpTool` hardcodes `ModalityState.nonModal()`, which queues EDT actions behind any open modal dialog. While `McpToolDispatcher` wraps tool calls in `ModalityState.any()`, `edtAction` re-narrows inside the tool, defeating the dispatcher's intent. All write and refactoring tools that use `edtAction` (rename, move, safe-delete, change-signature, member editing, replace-text, reformat, optimize-imports, SSR replace) block indefinitely while any modal dialog is open.

## The bug — traced and proven

`McpToolDispatcher` line 156:
```kotlin
return withContext(ModalityState.any().asContextElement()) { block() }
```

`AbstractMcpTool.edtAction` line 166:
```kotlin
withContext(Dispatchers.EDT + ModalityState.nonModal().asContextElement()) { action() }
```

The `nonModal()` overrides the `any()` from the dispatcher's coroutine context. Any modal dialog ("Load filesystem changes", "Trust project?", update notifications, etc.) causes `nonModal()` to queue the EDT action behind the dialog — the tool hangs until a human dismisses it.

## Fix

Change `edtAction` to use `ModalityState.any()`, matching what the dispatcher already established.

`commitDocuments` intentionally retains `nonModal()` — `TransactionGuard` rejects `commitAllDocuments` under `ModalityState.any()` with "Write-unsafe context". That's a platform constraint we can't work around.

## TDD — 3 tests

| Test | Before fix | After fix | What it proves |
|------|-----------|-----------|----------------|
| `testEdtActionBlocksBehindModalDialog` | FAIL (timeout) | PASS | The bug: nonModal() blocks behind modal |
| `testModalityStateAnyCompleteDuringModal` | PASS | PASS | Control: any() works during modal |
| `testEdtActionWorksWithoutModal` | PASS | PASS | Control: edtAction works without modal |

Tests use `LaterInvocator.enterModal/leaveModal` to simulate a modal dialog, `runInDispatchThread() = false` to force the `withContext` branch, and `withTimeoutOrNull(2s)` to detect blocking.

## Test plan

- [x] 3 platform tests: bug proven (FAIL before, PASS after), controls pass
- [x] All unit tests pass
- [ ] Manual: open a modal dialog in IntelliJ, call an MCP write tool — should complete without blocking

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -158,7 +158,7 @@ jobs:
 
       # Run Qodana inspections
       - name: Qodana - Code Inspection
-        uses: JetBrains/qodana-action@v2025.2.2
+        uses: JetBrains/qodana-action@v2026.2.0
         with:
           cache-default-branch-only: true
 


mdproctor:
Closing — TransactionGuard rejects `ModalityState.any()` for all PSI-modifying operations (`WriteCommandAction`, `commitAllDocuments`), not just `commitDocuments`. The "Write-unsafe context" assertion fires in CI for all three member editing off-EDT tests.

The fundamental constraint: there is no modality state that both (a) doesn't queue behind modal dialogs AND (b) satisfies TransactionGuard for PSI writes. `nonModal()` satisfies (b) but blocks behind modals. `any()` satisfies (a) but TransactionGuard rejects it.

The `McpToolDispatcher` already uses `any()` at the coroutine context level, but `edtAction` re-dispatches to EDT with `nonModal()` — this is correct because TransactionGuard enforces it. The blocking-behind-modals behavior is a genuine IntelliJ platform limitation, not a plugin bug we can fix.

The real mitigation is what 5.0.x already ships: pre-trusting projects (#272), VCS add bypass (5.0.1), and the `ensureWritable` pre-flight (#273) which prevents the read-only dialog. These prevent the modals from appearing in the first place.

PR #296, opened 2026-08-06
feat: add cross-platform EDT freeze watchdog with install button

mdproctor:
## Summary

Adds an EDT freeze watchdog that detects frozen IntelliJ UI via jstack thread analysis and auto-restarts the IDE. Installable from Settings > Index MCP Server > EDT Freeze Watchdog.

## Why

When the IntelliJ EDT (Event Dispatch Thread) freezes — typically from macOS AWT/AppKit deadlocks, modal dialogs, or GC pressure — all MCP tool calls that need EDT access hang indefinitely. The MCP HTTP server stays responsive (runs on Ktor threads), so HTTP health checks cannot detect the freeze. Claude Code sessions time out waiting for tools that will never complete.

This was proven in production: `CMenu.delItem` -> `AWTThreading.executeWaitToolkit` deadlock froze the EDT while the MCP server kept responding to pings. Only jstack thread analysis could detect it.

## How it works

1. External scripts (bash for macOS/Linux, PowerShell for Windows) run via OS scheduler every 30 seconds
2. `jstack` (from the IDE's bundled JBR) inspects the `AWT-EventQueue` thread state
3. If the EDT is in WAITING/BLOCKED outside the normal idle loop (`EventQueue.getNextEvent` / `NonBlockingFlushQueue`), it's frozen
4. After 2 consecutive detections (~1 minute), kills the process and relaunches the IDE
5. Logs all detections and restarts to `~/.ide-index-mcp/watchdog/watchdog.log`

## Installation flow

1. User clicks "Install Watchdog" in Settings > Index MCP Server
2. `WatchdogInstaller` detects OS, extracts the correct script from bundled JAR resources
3. Configures the script with IDE-specific paths (JBR jstack, launch command, PID pattern)
4. Registers the OS scheduled task (crontab or Task Scheduler)
5. "Uninstall Watchdog" reverses everything

## Cross-platform support

| Component | macOS | Linux | Windows |
|-----------|-------|-------|---------|
| Script | `watchdog.sh` (bash) | `watchdog.sh` (bash) | `watchdog.ps1` (PowerShell) |
| Scheduler | crontab (2 entries for 30s) | crontab | Task Scheduler |
| jstack | `<IDE>/jbr/Contents/Home/bin/jstack` | `<IDE>/jbr/bin/jstack` | `<IDE>\jbr\bin\jstack.exe` |
| Kill | `kill -9` | `kill -9` | `taskkill /F` |
| Relaunch | `open -a "IntelliJ IDEA"` | `nohup <IDE>/bin/idea &` | `Start-Process idea64.exe` |

No external dependencies — uses the IDE's bundled JBR for jstack.

## Test plan

- [x] 7 unit tests (watchdog dir, resource bundling, placeholder verification, jstack resolution, launch command, PID pattern, log path)
- [x] All unit tests pass
- [x] check-pr.sh 16/16
- [x] Bash script syntax validated
- [x] Proven working on macOS (detected `CMenu.delItem` deadlock, killed and relaunched)
- [ ] Manual: Linux/Windows verification

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- **EDT Freeze Watchdog** — detects frozen IDE UI via jstack thread analysis and auto-restarts. Installable from Settings → Index MCP Server → EDT Freeze Watchdog. Runs externally via cron (macOS/Linux) or Task Scheduler (Windows), so it survives IDE crashes. Uses the IDE's bundled JBR for jstack — no external dependencies.
+
 ## [5.2.2] - 2026-08-04
 
 ### Fixed
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/settings/McpSettingsConfigurable.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/settings/McpSettingsConfigurable.kt
@@ -64,6 +64,10 @@ class McpSettingsConfigurable : Configurable {
     private var minimumOpenProjectsSpinner: JSpinner? = null
     private var managedProjectsContent: JPanel? = null
 
+    // Watchdog UI fields
+    private var watchdogButton: javax.swing.JButton? = null
+    private var watchdogStatusLabel: JBLabel? = null
+
     private var lastHostValidation: ValidationInfo? = null
     private var hostValidationErrorLabel: JBLabel? = null
     private var hostValidationIcon: AsyncProcessIcon? = null
@@ -153,6 +157,9 @@ class McpSettingsConfigurable : Configurable {
             .addSeparator(10)
             .addComponent(JBLabel(McpBundle.message("lifecycle.section.title")), 5)
             .addComponent(lifecyclePanel, 1)
+            .addSeparator(10)
+            .addComponent(JBLabel(McpBundle.message("watchdog.section.title")), 5)
+            .addComponent(createWatchdogPanel(), 1)
             .addSeparator(10)
             .addComponent(JBLabel(McpBundle.message("settings.tools.title")), 5)
             .addComponent(availableToolsPanel, 5)
@@ -311,6 +318,57 @@ class McpSettingsConfigurable : Configurable {
 
         content.revalidate()
         content.repaint()
+    }
+
+    private fun createWatchdogPanel(): JComponent {
+        val installed = com.github.hechtcarmel.jetbrainsindexmcpplugin.util.WatchdogInstaller.isInstalled()
+
+        watchdogStatusLabel = JBLabel(
+            if (installed) McpBundle.message("watchdog.status.installed")
+            else McpBundle.message("watchdog.status.notInstalled")
+        )
+
+        watchdogButton = javax.swing.JButton(
+            if (installed) McpBundle.message("watchdog.uninstall.label")
+            else McpBundle.message("watchdog.install.label")
+        ).apply {
+            addActionListener {
+                val installer = com.github.hechtcarmel.jetbrainsindexmcpplugin.util.WatchdogInstaller
+                val result = if (installer.isInstalled()) installer.uninstall() else installer.install()
+
+                val nowInstalled = installer.isInstalled()
+                watchdogStatusLabel?.text = if (nowInstalled)
+                    McpBundle.message("watchdog.status.installed")
+                else
+                    McpBundle.message("watchdog.status.notInstalled")
+                text = if (nowInstalled)
+                    McpBundle.message("watchdog.uninstall.label")
+                else
+                    McpBundle.message("watchdog.install.label")
+
+                NotificationGroupManager.getInstance()
+                    .getNotificationGroup(McpConstants.NOTIFICATION_GROUP_ID)
+                    .createNotification(
+                        McpConstants.PLUGIN_NAME,
+                        result.message,
+                        if (result.success) NotificationType.INFORMATION else NotificationType.ERROR
+                    )
+                    .notify(null)
+            }
+        }
+
+        val descLabel = JBLabel(McpBundle.message("watchdog.description")).apply {
+            foreground = JBColor.GRAY
+        }
+
+        return FormBuilder.createFormBuilder()
+            .addComponent(descLabel)
+            .addComponent(JPanel(FlowLayout(FlowLayout.LEFT, 0, 5)).apply {
+                add(watchdogButton)
+                add(javax.swing.Box.createHorizontalStrut(10))
+                add(watchdogStatusLabel)
+            })
+            .panel
     }
 
     private fun createToolsPanel(): JComponent {
@@ -634,6 +692,8 @@ class McpSettingsConfigurable : Configurable {
         lifecycleLogToFileCheckBox = null
         minimumOpenProjectsSpinner = null
         managedProjectsContent = null
+        watchdogButton = null
+        watchdogStatusLabel = null
         uiDisposable?.let { Disposer.dispose(it) }
         uiDisposable = null
     }
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/util/WatchdogInstaller.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/util/WatchdogInstaller.kt
@@ -1,0 +1,303 @@
+package com.github.hechtcarmel.jetbrainsindexmcpplugin.util
+
+import com.intellij.openapi.application.PathManager
+import com.intellij.openapi.diagnostic.logger
+import com.intellij.openapi.util.SystemInfo
+import java.io.File
+
+/**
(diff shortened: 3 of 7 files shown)


hechtcarmel:
Hi, i reviewed it carefully, with the help of AI, and asked him to summarize the long conversations i had with Fable and to write the comment itself:

# Review: PR #296 — EDT Freeze Watchdog

Hey @mdproctor — first, thank you for this. The problem you're describing is completely real: I've seen the `AWTThreading.executeWaitToolkit` / `CMenu.delItem` freeze myself (it's tracked as [JBR-8207](https://youtrack.jetbrains.com/issue/JBR-8207), still open), and you're right that when the EDT deadlocks, the Ktor server keeps answering pings while every EDT-dependent tool hangs until the client times out. That's the worst kind of failure for an agent session — indistinguishable from a slow tool. The idea of using the bundled JBR's jstack is genuinely clever (version-matched, zero dependencies), and the two-strike threshold plus the idle-state whitelist show you thought about false positives.

That said, I'm not going to merge this, and I want to be upfront that it's not fixable with a round of review comments — it's the design. There are two separate problems: this plugin shouldn't be installing an OS-level process killer, and the implementation as written doesn't actually work on any of the three platforms once it's running under the scheduler. Let me walk through both, then what I *would* merge, because the underlying pain is worth solving.

## Why the design doesn't fit this plugin

**A false positive here destroys work.** Since the 2025.2 write-action rework, the platform *intentionally* parks the EDT in WAITING while background write actions hold the write lock ("[Road to Responsive IDEs](https://blog.jetbrains.com/platform/2026/04/road-to-responsive-ides/)"). A long refactoring or VFS batch update on 2025.3 — our target platform — can legitimately hold the EDT waiting past 60s, which is exactly your kill window (two samples 30s apart). The watchdog would `kill -9` the IDE mid-write-action: unsaved buffers gone, indexes potentially corrupted, at the worst possible moment. And the sharpest version of this: JBR-8207's own thread dump shows the freeze lasting 57–65 seconds *and then recovering*. Your watchdog would have killed that IDE seconds before it came back.

**The persistence outlives the plugin.** Nothing removes the crontab entries or the script when the *plugin* is uninstalled — only the Settings button does, and after uninstall there is no button. Marketplace has no uninstall hook that reliably covers this, which means we'd be shipping a mechanism whose steady state, for some users, is an orphaned background job that force-kills processes matching a name pattern, forever, with nothing left on the machine explaining where it came from. I'm not comfortable owning that support burden, and I don't think it passes [Marketplace approval guidelines](https://plugins.jetbrains.com/docs/marketplace/jetbrains-marketplace-approval-guidelines.html) §2.2.c (no interfering with the product's intended behavior — bypassing shutdown hooks and VFS/index flush via SIGKILL is squarely that).

**The optics are rough.** An IDE process rewriting the user's crontab wholesale, and on Windows registering a scheduled task that runs `powershell.exe -ExecutionPolicy Bypass` every minute, is textbook scheduled-task persistence as far as corporate EDR is concerned. People will get security tickets because of us.

**One click is not enough consent** for "software that may kill your IDE on a heuristic." The button fires immediately from the Configurable (not on Apply), with no confirmation dialog, and rewrites the whole crontab.

## The implementation doesn't work as installed

I verified these rather than eyeballing them, because I wanted to be fair to the PR. Every one reproduces:

1. **Cron registration fails on any machine with no existing crontab** — which is the default. `registerCrontab` runs `crontab -l` with `redirectErrorStream(true)`, so on a clean machine the stderr line `crontab: no crontab for <user>` is captured as the "existing crontab" and written back through `crontab -`, which rejects the input. Registration must treat a non-zero `crontab -l` exit as "empty" instead of merging stderr into content.

2. **On macOS the script is a silent no-op under cron.** The jstack line depends on `timeout`, which stock macOS does not ship (only Homebrew coreutils provides it), and cron jobs run with `PATH=/usr/bin:/bin`. So `EDT_DUMP` is always empty and the script exits before ever checking anything — no log line, no detection. I suspect your successful macOS test ran the script from an interactive shell where Homebrew's `timeout` was on PATH; under cron it cannot have worked.

3. **The relaunch commands are destroyed by quoting on macOS and Windows.** `resolveIdeLaunchCmd()` returns *pre-quoted* strings which are then substituted into already-quoted template contexts. On macOS the result is that `eval` sees a mangled command for any app name containing a space (i.e., "IntelliJ IDEA") — the watchdog kills the IDE and never restarts it. On Windows it's worse: the substitution produces `$IdeLaunchCmd = ""C:\...\idea64.exe""`, which is a PowerShell **parse error**, so the entire .ps1 aborts on every run — the Windows watchdog never executes at all. (Also: the scheduled task pops a visible console window every 60 seconds, and `Start-Process -FilePath` doesn't accept embedded quotes even if the parse error were fixed.)

4. **On Linux the relaunch can't work either**: cron's environment has no `DISPLAY`/`WAYLAND_DISPLAY`, so `nohup .../bin/idea` dies immediately — and the script logs "IDE relaunched" regardless, because the launch result is discarded.

5. **The detection heuristic misses the deadlock class it was built for.** `head -1` of the extracted EDT block is the jstack *header* line ("`waiting on condition`", "`in Object.wait()`", "`waiting for monitor entry`"); the tokens `WAITING`/`BLOCKED`/`parking` only ever appear on the second line (`java.lang.Thread.State:`), which the check never sees. I reproduced this with a thread pinned in BLOCKED-on-monitor and in `Object.wait()`: both were classified as "doing work" and *reset* the failure counter. A monitor-deadlocked EDT — the classic case — is undetectable, while the string "waiting on condition" in the header is what makes the idle-adjacent states match. The check works by coincidence for some states and inverts for others.

6. **It only knows about IntelliJ IDEA.** `idea64` / `bin/idea` / the `MacOS/idea` pgrep pattern are hardcoded, but this plugin ships to 13 JetBrains IDEs — on PyCharm, WebStorm, GoLand etc. the watchdog is silently inert (or, via the `extractMacAppName` fallback to "IntelliJ IDEA", armed against a *different* IDE than the one that installed it). We already have per-product derivation in `IdeProductInfo`; anything like this would need to go through it. Relatedly, the single shared script path + marker + task name means installing from a second JetBrains IDE silently hijacks the first one's watchdog.

7. **The Settings panel lies after a failed install.** `install()` writes the script *before* registering the scheduler, and `isInstalled()` only checks file existence — so after the cron failure in (1), the panel shows "Installed — monitoring for EDT freezes" next to the error notification.

8. **The installer itself blocks the EDT.** `install()`/`uninstall()` run `crontab`/`schtasks` with unbounded `waitFor()` directly in the button's ActionListener. A freeze-prevention feature shouldn't be a SlowOperations violation; this needs `Task.Backgroundable` + `GeneralCommandLine` at minimum.

There are more (the "2 consecutive checks" counter never expires and survives inconclusive runs, so two unrelated 30s waits days apart can combine into a kill; the PowerShell log-trim pipeline truncates the log to zero instead of trimming it; schtasks runs at 60s, not the 30s the PR describes), but the above is enough to gate.

## Process items (would apply to any version)

- The branch conflicts with `main` — your CHANGELOG entry collides with the 5.3.0 section from #295. Needs a rebase.
- A new Settings section needs README.md / USAGE.md coverage.
- `testJstackPathResolution` wraps its assertions in `if (path != null)` — it passes even if the implementation is deleted. Our test rules (CLAUDE.md, "Assertion rules") don't allow conditional skips; use `Assume` or make it deterministic. And `testIdePidPatternIsNotEmpty` asserting the pattern contains "idea" would cement bug (6) as the contract.

## What I'd happily merge instead

The real, in-scope problem is: *agents can't tell a frozen IDE from a slow tool.* That's solvable without killing anything, in about 30 lines, at the right layer:

**An EDT heartbeat + fail-fast gate in `McpToolDispatcher`.** A background thread posts a trivial `invokeLater` every few seconds and records the last completion time. `McpToolDispatcher.call()` already returns error results before touching the EDT (that's how the disabled-tool gate works), and the Ktor/CIO path never needs the EDT — so when the heartbeat is stale, every tool call can return an immediate, explicit MCP error like *"IDE UI thread has been unresponsive for 90s — tool calls requiring the IDE are unavailable; consider restarting the IDE"* instead of hanging into the client timeout. That converts the undiagnosable hang into an actionable signal, which is the part of this that belongs in an MCP plugin. If you want to rework this PR in that direction I'll review it quickly and gladly.

If auto-restart genuinely proves necessary beyond that, the shape I could accept is an **opt-in child process** spawned by the plugin at startup that monitors its parent PID and exits with it — no crontab, no schtasks, nothing surviving uninstall, per-IDE paths resolved fresh each start, behind an explicit confirmation dialog. (Worth noting: the cron design's headline advantage, "survives IDE crashes", is inert in this PR anyway — the script's own first check is "no IDE process found → reset and exit"; it never restarts a *dead* IDE, only a frozen one.) Alternatively, the scripts could live in the repo as documented `contrib/` assets that a power user installs deliberately — that's an honest consent model for a tool this sharp.

Thanks again for the deep investigation — the jstack-of-the-bundled-JBR trick and the production thread dumps clearly took real work, and the diagnosis of *why* HTTP health checks can't see EDT freezes is exactly right. I'm closing this one, but I'd love to see the heartbeat gate as a follow-up.

PR #297, opened 2026-08-06
fix: resolve record component instead of backing field in symbol lookup

jevanlingen:
## Summary

`ide_find_references` (and other tools that resolve members by symbol name) returned 0 results for Java record component accessors. The symbol-mode lookup for e.g. `com.example.Point#x` resolved to the synthetic backing field rather than the `PsiRecordComponent`, and `ReferencesSearch` finds nothing on synthetic fields.

The fix: when resolving a member by name on a record class, check `recordComponents` first and return the matching component before falling through to `findFieldByName`.

## Why
IntelliJ did does find usages for a record:

image

where the MCP tool does not:

```
Tool: ide_find_references
Status: SUCCESS
Timestamp: 2026-08-06T07: 51:17.890806Z
Duration: 16ms

Parameters:
{
"language": "Java",
"symbol": "package.name.SomeProperties#workflowFlowId",
"project_path": "<project-path>"
}

Result:
{
  "usages": [],
  "totalCount": 0,
  "truncated": false,
  "nextCursor": null,
  "hasMore": false,
  "totalCollected": 0,
  "offset": 0,
  "pageSize": 0,
  "stale": false,
  "resolvedSymbol": {
    "name": "workflowFlowId",
    "kind": "field",
    "container": "SomeProperties",
    "file": "path/to/SomeProperties.java",
    "line": 16
  },
  "totalIsExact": true
}
```

## Test
A `FindUsagesRecordComponentBehaviorTest` is included to guard this behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **`ide_find_references` now finds usages of Java record component accessors.** Symbol-mode lookup (`language`+`symbol`) for a record member (e.g. `com.example.Point#x`) resolved to the backing field instead of the record component, causing `ReferencesSearch` to return 0 results. The resolver now prefers the `PsiRecordComponent` over the synthetic field for record classes.
+
 ## [5.3.0] - 2026-08-05
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


hechtcarmel:
Great, thanks! will release in next version later today

PR #301, opened 2026-08-06
feat: EDT heartbeat fail-fast gate for frozen UI thread

### LINKED ISSUE #300: EDT heartbeat: fail-fast gate for tool calls when UI thread is frozen
## Problem

When the IDE's Event Dispatch Thread (EDT) is frozen (e.g., due to native deadlocks like [JBR-8207](https://youtrack.jetbrains.com/issue/JBR-8207)), the Ktor HTTP server stays responsive — so MCP clients can still send tool calls. But those calls hang until the client timeout because they need the EDT for PSI operations, project resolution, or write actions. From the agent's perspective, this is indistinguishable from a slow tool.

## Solution

An EDT heartbeat monitor that lets `McpToolDispatcher` fail fast:

1. **`EdtHeartbeatService`** — a background thread posts `invokeLater` every 5 seconds and records when the callback executes. When the EDT is truly frozen, callbacks stop executing and the heartbeat goes stale.

2. **Fail-fast gate in `McpToolDispatcher.call()`** — before project resolution, checks the heartbeat. If stale for 90+ seconds, returns an immediate MCP error:
   > "IDE UI thread has been unresponsive for 120s — tool calls requiring the IDE are unavailable. Consider restarting the IDE."

This converts an undiagnosable hang into an actionable signal — no process killing, no external scripts, nothing that survives plugin uninstall.

## Context

Follow-up to the design discussion in #296, implementing the approach suggested by the maintainer.

mdproctor:
## Summary

Closes #300. Follow-up to the design discussion in #296.

When the EDT is frozen (e.g. [JBR-8207](https://youtrack.jetbrains.com/issue/JBR-8207) native deadlock), the Ktor HTTP server stays responsive — MCP clients can still send tool calls, but they hang until the client timeout because they need the EDT. This is indistinguishable from a slow tool.

- **`EdtHeartbeatService`** — app-level service that posts `invokeLater(ModalityState.any())` every 5 seconds from a pooled-thread `Alarm` and records the completion timestamp. When the EDT is truly frozen, callbacks stop executing.
- **Fail-fast gate in `McpToolDispatcher.call()`** — checks the heartbeat before project resolution. If stale for 90+ seconds, returns an immediate MCP error: *"IDE UI thread has been unresponsive for Ns — tool calls requiring the IDE are unavailable. Consider restarting the IDE."*
- Injectable `edtUnresponsiveDurationMs` constructor parameter follows the existing history-callback pattern for testability.

~40 lines of production code, 8 tests.

## Test plan

- [x] `EdtHeartbeatServiceTest` — 5 platform tests: responsive after init, null duration when responsive, unresponsive when stale, returns duration when unresponsive, recovers after fresh heartbeat
- [x] `McpToolDispatcherEdtGateTest` — 3 platform tests: fails fast with injected stale heartbeat, proceeds with responsive heartbeat, error message includes duration
- [x] Existing `McpToolDispatcherTest` and `McpToolDispatcherHistoryFailureTest` still pass (no regressions from new constructor param)
- [x] Full test suite passes (`./gradlew test`, `scripts/check-pr.sh` all green)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- **EDT heartbeat fail-fast gate** ([#300](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/issues/300)) — when the IDE's UI thread (EDT) is frozen, tool calls now return an immediate error instead of hanging until the client timeout. A background heartbeat monitors EDT responsiveness; if it has been unresponsive for 90+ seconds, `McpToolDispatcher` short-circuits with an actionable message telling the agent to restart the IDE.
+
 ## [5.3.1] - 2026-08-06
 
 ### Fixed
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/EdtHeartbeatService.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/EdtHeartbeatService.kt
@@ -1,0 +1,57 @@
+package com.github.hechtcarmel.jetbrainsindexmcpplugin.server
+
+import com.intellij.openapi.Disposable
+import com.intellij.openapi.application.ApplicationManager
+import com.intellij.openapi.application.ModalityState
+import com.intellij.openapi.components.Service
+import com.intellij.openapi.components.service
+import com.intellij.util.Alarm
+import org.jetbrains.annotations.TestOnly
+
+@Service(Service.Level.APP)
+class EdtHeartbeatService : Disposable {
+
+    @Volatile
+    private var lastHeartbeatNs: Long = System.nanoTime()
+
+    private val alarm = Alarm(Alarm.ThreadToUse.POOLED_THREAD, this)
+
+    companion object {
+        private const val HEARTBEAT_INTERVAL_MS = 5_000L
+        const val UNRESPONSIVE_THRESHOLD_MS = 90_000L
+
+        fun getInstance(): EdtHeartbeatService = service()
+    }
+
+    init {
+        val app = ApplicationManager.getApplication()
+        if (app != null && !app.isUnitTestMode) {
+            scheduleHeartbeat()
+        }
+    }
+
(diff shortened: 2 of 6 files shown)


hechtcarmel:
This is the approach I described in #296, so no design objections. Ran the branch locally, suite is green. A few things to fix before I merge:

**1. Use `System.nanoTime()` instead of `currentTimeMillis()`.**
Wall clock keeps ticking while the laptop is asleep. Close the lid for 2 hours, wake it, agent calls a tool -> "IDE UI thread has been unresponsive for 7200s, consider restarting the IDE" on a perfectly healthy IDE. nanoTime is monotonic and doesn't advance during suspend on mac/linux, so it's a straight swap.

**2. Remove the `<applicationService>` entry from plugin.xml.**
The class is already a light service via `@Service(Service.Level.APP)`. Registering it in XML as well is redundant (DevKit flags it), and the comment two lines below the one you added literally says services here are registered via annotation only.

Related: the service is only instantiated on the first tool call, so if the EDT is already frozen when that call arrives, the gate says "healthy" for 90s. Touch `EdtHeartbeatService.getInstance()` when the server starts so the baseline exists from startup.

**3. Conflicts** — CHANGELOG conflicts with the 5.3.1 section from #299.

Test notes:

- `EdtHeartbeatServiceTest` mutates an app-level service that survives across test methods (the heartbeat alarm never runs in unit test mode, so a stale timestamp stays stale). It currently passes only because JUnit 3 happens to run `testServiceReturnsDurationWhenUnresponsive` last. Reset the heartbeat in `setUp()`.
- New platform tests should extend `McpPlatformTestCase`, not `BasePlatformTestCase` (see CLAUDE.md).
- `testEdtGateMessageIncludesToolName` doesn't check the tool name, it just re-asserts the seconds formatting from the first test. Rename or drop it.
- `isEdtResponsive()` has no production caller; fine as a test helper but you could also just delete it.

The `ModalityState.any()` heartbeat and the gate placement (before `resolveOrOpen`, which can need the EDT) are both right. Fix the above and this goes in.

hechtcarmel:
Thanks, merging and wil lrelease today

Issue #310, opened 2026-08-13
ide_refactor_rename blocks on read-only file dialog instead of returning error


mdproctor:
## Problem

When `ide_refactor_rename` encounters read-only files in the refactoring scope (generated metamodel classes in `target/`, Flyway migration SQL, files from other content roots), the IntelliJ refactoring engine shows a modal "These files are read-only — proceed?" dialog instead of returning an error.

In headless/automated MCP sessions, this dialog blocks the EDT indefinitely. The agent never gets a response — just a client timeout with no explanation of what went wrong. The EDT heartbeat (#301) will eventually detect and report the hang, but the original tool call is lost.

## Expected behavior

The tool should detect read-only files in the rename scope *before* executing the refactoring and return an actionable error listing them:

```json
{
  "error": "Rename blocked by read-only files in scope",
  "readOnlyFiles": [
    "target/generated-sources/annotations/com/example/WorkItem_.java",
    "src/main/resources/db/migration/V1__create_work_item.sql"
  ],
  "suggestion": "Run 'mvn clean' + ide_sync_files to remove generated sources, or exclude Flyway migrations from the project's content roots."
}
```

## Scope

This affects any refactoring tool that can touch files outside the target (`ide_refactor_rename`, `ide_change_signature`, potentially `ide_refactor_safe_delete`). The `RenameProcessor.findUsages()` step already discovers affected files before modifying anything — the pre-check can run there.

The existing `ensureWritable` pre-flight (#273) only checks the *target* file, not the full refactoring scope.

PR #313, opened 2026-08-13
fix: auto-enable VFS sync on startup to prevent file-change dialog

mdproctor:
## Summary

When `GeneralSettings.isSyncOnFrameActivation` (aka `autoSyncFiles` in `ide.general.xml`) is `false`, every external file modification triggers a modal "Load File System Changes" dialog. In headless MCP sessions this blocks the EDT indefinitely — the agent never gets a response.

The fix: `McpServerService.initialize()` now enables auto-sync on startup. This is a one-line change that ensures the IDE automatically picks up external file changes without prompting, which is essential for MCP agents that modify files via `Write`, `Edit`, and other external tools.

The setting is only enabled, never disabled — if a user has it on already (the default), nothing changes. If they disabled it manually, the MCP plugin re-enables it because headless operation requires it.

## Test plan

- [x] Compiles clean
- [x] Verified manually: with `autoSyncFiles=false` in `ide.general.xml`, restarting with the plugin re-enables it and the dialog no longer appears

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **Auto-sync of external file changes is now enabled automatically.** The plugin sets `GeneralSettings.isSyncOnFrameActivation = true` on startup, preventing the "Load File System Changes" modal dialog that blocks headless MCP sessions when agents modify files externally.
+
 ## [5.5.1] - 2026-08-12
 
 ### Fixed
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/McpServerService.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/server/McpServerService.kt
@@ -103,6 +103,7 @@ class McpServerService(
         val startServer = shouldStartServer()
         if (startServer) {
             EdtHeartbeatService.getInstance()
+            ensureAutoSyncEnabled()
             val settings = McpSettings.getInstance()
             val port = settings.serverPort
             val host = settings.serverHost
@@ -110,6 +111,14 @@ class McpServerService(
             LOG.info("MCP Server Service initialized with Ktor CIO server")
         } else {
             LOG.info("Initialized MCP tool metadata without starting server in unit test mode")
+        }
+    }
+
+    private fun ensureAutoSyncEnabled() {
+        val gs = com.intellij.ide.GeneralSettings.getInstance()
+        if (!gs.isSyncOnFrameActivation) {
+            gs.isSyncOnFrameActivation = true
+            LOG.info("Enabled auto-sync of external file changes for MCP operation")
         }
     }
 


hechtcarmel:
# PR #313 review — auto-enable VFS sync on startup

Thanks for digging into this — modal dialogs blocking the EDT are a genuinely painful failure mode for this plugin, and we've been fighting them one by one (#310 pre-checked the read-only-files dialog for exactly this reason, and `EdtHeartbeatService` exists because of it). So the pain point is real and I appreciate the report. But I can't merge this, for two independent reasons: the diagnosis doesn't match how the platform works, and even if it did, the fix would still be the wrong kind of fix.

## The diagnosis

"Load File System Changes" isn't a dialog — it's a *button* on the **File Cache Conflict** dialog (`MemoryDiskConflictResolver`). That dialog fires when a file has **unsaved in-memory changes in an open editor** and an external process modifies the same file on disk. It is not gated on `GeneralSettings.isSyncOnFrameActivation`; it can appear with that setting in either state, whenever a VFS refresh discovers the memory/disk conflict. JetBrains documents the trigger conditions here: https://www.jetbrains.com/help/idea/file-cache-conflict.html

`isSyncOnFrameActivation` also can't do what the PR description says it does in a headless session. It triggers a refresh **when the IDE frame is activated** — i.e., when the user clicks back into the IDE window. In the scenario you describe (agent driving the IDE over MCP while it sits in the background), no frame activation ever happens, so the setting fires exactly zero times regardless of its value. The refreshes that matter in that scenario are the ones this plugin issues itself (`ensurePsiUpToDate` / `ide_sync_files` via `markDirtyAndRefresh`), and those run the same either way.

So the manual verification in the test plan — flip the setting, dialog stops appearing — is most likely correlation from a different mechanism: with auto-sync on, disk changes get loaded eagerly the moment you focus the IDE, before you start typing, so the document is less likely to be in the modified-plus-stale state that produces the conflict. That's a real effect for a user actively touching the IDE, but it's probabilistic, it isn't the headless case the PR claims to fix, and the conflict dialog can still appear with the setting on. If you have a reproducible case where a tool call blocks indefinitely with `autoSyncFiles=false`, I'd genuinely like to see the exact steps and which tool call hangs — that's worth an issue with the EDT stack trace attached, because it would tell us which of *our* code paths is pumping a refresh into a modal context.

## The fix

Even granting the diagnosis, a plugin silently and permanently overriding an IDE-wide user setting is not something I'm willing to ship:

- `isSyncOnFrameActivation` is an application-level setting users disable **deliberately** — huge monorepos, network mounts, battery life. With this PR installed, their choice silently un-makes itself on every IDE start, with nothing but a log line, forever, and no way to opt out short of uninstalling the plugin. That's the kind of behavior that earns plugins one-star reviews and Marketplace complaints, and rightly so.
- It also cuts against this plugin's own design. Sync behavior here is deliberately opt-in and scoped: the "Sync external file changes" setting is off by default *because* eager syncing has a real performance cost, and `ide_sync_files` exists so the agent can request a refresh exactly when it needs one. Flipping a global IDE setting to solve a per-session problem inverts that.
- Smaller stuff that would also need fixing if the approach were right: no test, and the mid-expression `com.intellij.ide.GeneralSettings` FQN should be an import (same nit as #307).

## What I'd take instead

If the underlying problem is "a modal dialog can wedge a headless session" — agreed, and the direction we've already settled on is a generic one: an EDT-heartbeat fail-fast gate in `McpToolDispatcher`, so any tool call that would wait on a blocked EDT returns an actionable error instead of hanging. That covers this dialog and every future one, without touching user settings.

If the underlying problem is specifically the File Cache Conflict dialog, the honest fix is on the conflict itself, not the sync setting — e.g. making sure our refresh paths save/commit documents in the right order, or detecting the conflicted state and reporting it to the agent. A repro would let us do that properly.

If after all that you still want the plugin to manage `isSyncOnFrameActivation`, it has to be an explicit **opt-in** setting in Settings → Tools → Index MCP Server, defaulting to off, with a notification when the plugin changes the value — never a silent override in `initialize()`.

Closing this one as-is, but happy to review a follow-up along any of those lines — and thanks again for the report; the dialog sighting itself is useful signal.

Issue #277, opened 2026-07-30
ide_run_tests times out on long running test, ignores timeoutSeconds


alekseytumbaev:
### IDE

IntelliJ IDEA

### IDE version

2026.1

### Plugin version

5.1.0

### MCP client

Claude Code

### Transport endpoint used

Streamable HTTP (/index-mcp/streamable-http)

### Steps to reproduce

1. Run `ide_run_tests` on the test that takes a long time, e.g. `Thread.sleep(300_000)`
2. Set `timeoutSeconds` to 7200

### Expected behavior

Test runs without timeout

### Actual behavior

Index MCP Server shows status `PENDING`:
```
Tool: ide_run_tests
Status: PENDING
Timestamp: 2026-07-30T08:57:49.285055600Z

Parameters:
{"target":"com.devexperts.dxcore.tools.ipfsynchronizer.futures.span.FuturesSpanMarginJobRequestListenerTest","timeoutSeconds":7200}
```

Claude Code says he receives this: ` {"error": "The operation timed out."}`

### Relevant idea.log snippet

```text
2026-07-30 12:57:49,285 [1616949]   INFO - FeatureRegistry[Tool] - Getting Tool: "ide_run_tests"
2026-07-30 12:57:49,285 [1616949]   INFO - FeatureRegistry[Tool] - Got Tool: "ide_run_tests"
2026-07-30 12:57:49,471 [1617135]   INFO - #c.i.w.i.i.j.s.JpsGlobalModelSynchronizerImpl - Saving global entities com.intellij.platform.workspace.jps.entities.SdkEntity to files
2026-07-30 12:57:49,471 [1617135]   INFO - #c.i.w.i.i.j.s.JpsGlobalModelSynchronizerImpl - Saving global entities com.intellij.platform.workspace.jps.entities.LibraryEntity to files
2026-07-30 12:57:49,539 [1617203]   INFO - #c.i.d.d.DataSourceStorageShared$App - Attempting to return empty stored state
2026-07-30 12:57:49,544 [1617208]   INFO - #c.i.c.i.CompileDriver - COMPILATION STARTED (BUILD PROCESS)
2026-07-30 12:57:49,963 [1617627]   INFO - #c.i.c.s.BuildManager - Using preloaded build process to compile C:/Work/xt/xt-1
[...]

PR #320, opened 2026-08-17
fix: ide_open_project auto-links Maven/Gradle when build file exists

### LINKED ISSUE #319: ide_open_project should auto-link Maven/Gradle when build file exists
## Problem

When a project is opened via `ide_open_project`, the directory is opened and indexed, but the Maven/Gradle build model is not linked — even when a `pom.xml` or `build.gradle` exists at the project root. This leaves the project in a state where:

- `ide_find_class` with `project_and_libraries` scope works against previously cached JARs
- `ide_reload_project` reports: "Maven: skipped — build file found on disk but project is not linked in IntelliJ"
- `ide_import_modules` reports: "already imported as a module" — but this is an IntelliJ directory module, not a Maven-linked module

There is no MCP tool that can perform the equivalent of "right-click pom.xml → Add as Maven Project" or clicking "+" in the Maven tool window.

## Impact

- SNAPSHOT dependency updates aren't picked up
- New types added in newer SNAPSHOTs are invisible to `ide_find_class`
- Agents can't self-heal — requires manual "link Maven in IntelliJ tool window" which defeats agentic workflow

## Suggested fix

`ide_open_project` should auto-link Maven when `pom.xml` is found at the project root (and Gradle when `build.gradle`/`build.gradle.kts` is found). This matches what a human would expect when opening a build-system project — no one opens a Maven project and then manually links it.

mdproctor:
## Summary

Closes #319.

When `ide_open_project` opens a directory that has a `pom.xml` or `build.gradle`/`build.gradle.kts` but no linked build model in IntelliJ, the tool now automatically imports the build system after indexing completes. Previously agents were stuck in a dead end:

- `ide_reload_project` → "Maven: skipped — build file found on disk but project is not linked"
- `ide_import_modules` → "already imported as a module" (directory module, not Maven-linked)
- No MCP tool could do the equivalent of "right-click pom.xml → Add as Maven Project"

The fix adds `autoLinkBuildSystem()` after `awaitSmartMode()`:
- Checks `ProjectDataManager.getExternalProjectsData()` to see if the build system is already linked
- If not linked but a build file exists on disk, imports via `ProjectUtils.importMavenModule` (Maven) or `ExternalSystemUtil.refreshProjects` (Gradle)
- Reports the outcome in the success message ("Maven project linked automatically.")

Build system detection reuses the same `ProjectSystemId` / build-file patterns as `ReloadProjectTool`.

## Test plan

- [x] Manual verification: `ide_open_project` on a Maven project that was never opened → Maven auto-linked, `ide_reload_project` succeeds on subsequent call


hechtcarmel:
# PR #320 review — not merging as-is

First off: #319 is a real problem and a good catch. The dead end is exactly as described — `ide_reload_project` says "not linked", `ide_import_modules` says "already imported as a module", and nothing on the MCP surface can do what "Add as Maven Project" does in the UI. I want this capability in the plugin. But this implementation doesn't deliver it, and in a few places it actively makes things worse, so I'm requesting changes rather than merging.

I verified everything below against the actual 2025.3 platform (decompiled `app.jar` / `maven.jar` from the IDE the plugin builds against, cross-checked with intellij-community sources), not just by reading the diff.

## The Gradle half doesn't do anything

`ExternalSystemUtil.refreshProjects(ImportSpecBuilder(project, GRADLE))` is a *refresh of already-linked projects*, not a link. The 2025.3 implementation literally does:

```java
var manager = ExternalSystemApiUtil.getManager(spec.getExternalSystemId());
if (manager == null) return;                       // Gradle plugin absent → silent no-op
var projectsSettings = settings.getLinkedProjectsSettings();
if (projectsSettings.isEmpty()) return;            // not linked → silent no-op
```

The branch only runs when `getExternalProjectsData(GRADLE)` came back empty — i.e. precisely the unlinked state — so it's a guaranteed no-op, after which the tool still returns **"Gradle project linked automatically."** That's a false success in the exact scenario the PR title claims to fix. It's also the one thing CONTRIBUTING.md calls out by name ("Detecting linked build systems": *if a build file exists but the project is not linked, tell the user how to link it rather than claiming success*).

## The Maven "already linked" check can never fire

Maven is not an External System. The bundled Maven plugin registers no `externalSystemManager` and contains zero references to `ExternalProjectsDataStorage` — the storage `getExternalProjectsData()` reads is populated only by external-system syncs (Gradle etc.). So `getExternalProjectsData(project, ProjectSystemId("MAVEN"))` is empty *by construction*, always.

(Yes, `ReloadProjectTool` has the same check. There it's been mostly harmless — worst case a wrong skip message. Copied here, it inverts into action: every fresh open of an already-imported Maven project now falls through the guard and re-runs a full import.)

Consequences:

- `MavenProjectAsyncBuilder.commitSync` → `addManagedFilesWithProfiles` forces a full incremental re-sync even when the pom is already managed. Every reopen of a healthy Maven repo pays a complete dependency re-resolution it didn't need, plus a bogus "Maven project linked automatically." message.
- Worse: because Maven is iterated first and its guard is dead, a **Gradle-linked repo that keeps a stray root pom.xml** (publishing pom, BOM, migration leftover) gets force-mavenized on open — managed pom registered, Maven modules layered over the Gradle model. That's project-model corruption the user has to undo by hand.

The check that actually works is `MavenProjectsManager`-based — which is exactly what the platform's own `MavenUnlinkedProjectAware.isLinkedProject` does (it calls `MavenProjectsManager.getProjects()`).

## It breaks the tool's blocking contract

`autoLinkBuildSystem` runs *after* the `withTimeoutOrNull(timeoutSeconds)` block has already returned READY. `commitSync` is fully blocking (`runBlockingMaybeCancellable` off-EDT) until the sync commits. So on a previously-imported multi-module repo the call can block for minutes, bounded by neither `timeoutSeconds` (whose docs say "maximum seconds to wait") nor the hard rule this repo established with the long-poll work in #277: no tool call may outlive the MCP client's ~60s timeout. The client times out, the result is lost, and the retry lands on the "already open" fast path — which brings me to:

## It doesn't actually fix #319's dead end

The auto-link only runs on the *fresh-open* path. The state the issue describes — project **already open**, build model unlinked — hits the early return at the top of `doExecute` ("Project 'X' is already open.") and never links. Same for the timeout branch ("open but still indexing"): no link, and the natural retry gets "already open". If the broken project is the only open one, `ide_close_project` refuses to close it, so close-and-reopen isn't even available. The reported scenario survives the fix verbatim.

## Smaller but real

- Returning "open and ready. Maven project linked automatically." right as the import re-enters dumb mode breaks the description's promise that subsequent calls succeed immediately. `OpenWorkspaceTool` re-awaits smart mode after importing — same pattern applies here. (Linking *before* the smart-mode wait would also avoid indexing everything twice.)
- Both branches report async work in the past tense. `refreshProjects` schedules; `commitSync` returns a preview module for newly-created projects while the real import continues in the background. `ReloadProjectTool` words the same call "reload scheduled" — that honesty matters double for agents.
- No try/catch around `autoLinkBuildSystem`: `AbstractMcpTool.execute` rethrows unknown exceptions, so any failure in this best-effort step (the reflective `getMethod` in `importMavenModule` sits outside its try/catch, and `MavenProjectAsyncBuilder` is wizard internals that can drift between IDE builds) turns a *successful* open into a failed tool call.
- `refreshAndFindFileByPath(basePath) ?: continue` silently falls through to the Gradle spec — a Maven project can get a "Gradle project linked automatically" answer. `ImportModulesTool` reports this exact failure ("could not resolve directory in VFS").
- Auto-link is unconditional. IntelliJ persists a per-project "user dismissed the link suggestion" choice; force-linking on open overrides that recorded intent and executes the repo's build scripts with no opt-out short of disabling the tool. Everything else in this plugin that changes behavior takes a flag.
- `BuildSystemSpec`/`BUILD_SYSTEMS` are copied verbatim from `ReloadProjectTool` (and the copy already diverged: per-system reporting there, whole-function abort here). The `if (label == "Maven")` dispatch on a display string is fragile on top.
- No tests, and no doc updates outside the changelog — README/USAGE/CLAUDE.md/tools-reference/SKILL.md and the in-code tool description all still describe the old behavior; USAGE's example response no longer matches. The changelog entry itself over-promises (it describes the Gradle behavior the code doesn't have) and sits under "Fixed" where new capability on an existing tool belongs under "Added".

## What I'd like instead

The platform already solved this uniformly, and it's public API: `ExternalSystemUnlinkedProjectAware` (`com.intellij.externalSystemUnlinkedProjectAware` EP). Both Maven and Gradle register implementations in 2025.3, and it gives you, per build system:

- `isBuildFile(project, file)` / `buildFileExtensions()` — detection, no hand-rolled file table
- `isLinkedProject(project, path)` — a linked-check that is *correct for Maven* (delegates to `MavenProjectsManager`) and correct for Gradle
- `suspend linkAndLoadProjectAsync(project, path)` — the exact code path the IDE's own "Load Gradle/Maven project?" notification runs when the user clicks Load

Iterating that EP handles plugin absence for free (extension simply not there — think WebStorm/PyCharm with a build.gradle in the repo), works for any build system that registers it, and deletes both private `BUILD_SYSTEMS` tables.

Shape-wise, I think this wants to be:

1. A dedicated **`ide_link_build_system` tool** (new tool checklist applies: `ToolNames`, `ToolRegistry`, `DEFAULT_DISABLED_TOOLS` + settings migration, the six doc locations, golden manifest) — so it also heals the *already-open* case, which is the case #319 actually reported. `ide_reload_project`'s "not linked" message should point at it instead of telling agents to click IDE UI they can't click.
2. Optionally, an explicit **`autoLink` parameter on `ide_open_project`** (default off) that calls the same code inside the timeout window and re-awaits smart mode afterwards.
3. Honest messages: "link scheduled/completed", "already linked", "build file found but <system> plugin not available", and never "linked" when nothing happened.
4. Tests for the guard logic and the message variants (the plugin-unavailable and already-linked paths are testable headlessly), per the testing rules in CONTRIBUTING.

Happy to review a follow-up along those lines — the capability is genuinely wanted, and thanks for the thorough issue writeup; it made the problem easy to reproduce and reason about.

hechtcarmel:
# Review: #320 — `ide_link_build_system` + `autoLink` on `ide_open_project` (round 2)

This is a much better shape than the first pass. Everything I raised last time has been dealt
with properly rather than papered over:

- Linking now goes through `ExternalSystemUnlinkedProjectAware.linkAndLoadProjectAsync`, so
  Gradle actually links instead of silently doing nothing.
- The `return null`/`continue` mix-up is gone.
- `ide_link_build_system` gives already-open projects a way out, and `ide_reload_project` now
  points at it instead of telling an agent to click something in a tool window.
- `autoLink` defaults to `false`, which is more conservative than I asked for and the right call.
- `BuildSystemLinker` is shared instead of copy-pasted.
- Registration is complete and correct — `ToolNames`, `ALL` in the right alphabetical slot,
  `DEFAULT_DISABLED_TOOLS`, schema bump to 8 with a migration entry, `ToolRegistry`, the golden
  manifest, and all six doc locations.

I still can't merge it. Four things, and the first two are the ones that matter.

## Blocking

### 1. The new test fails on its own, and covers nothing when it passes

```bash
./gradlew test --tests "*LinkBuildSystemToolBehaviorTest*"
```

```
2 tests completed, 1 failed

junit.framework.AssertionFailedError: Should mention no build file or plugin unavailable,
was: Path is not a directory:
/private/var/folders/.../T/unitTest_reportsErrorForNonExistentPath_3I8lif6kdrMzeKy64zsRpXgIvfN
```

Deterministic — two runs out of two on a clean checkout of this branch. The light fixture is
reused across test methods, so by the time `testReportsNoBuildFile...` runs, `project.basePath`
still points at the temp directory named for `testReportsErrorForNonExistentPath`, which was
deleted at that test's teardown. The tool's `File(path).isDirectory` guard trips and we never
reach the linker at all.

`./gradlew clean test --no-build-cache` does pass, so the test plan in the description isn't
wrong. But `./gradlew test --rerun-tasks` — what `scripts/check-pr.sh` runs, which CONTRIBUTING
says to run before every push — failed for me once with the same assertion (`1056 tests
completed, 1 failed`) before passing on a re-run. So it's flaky in the suite and reliably broken
in isolation.

The coverage problem is the bigger half. `platformBundledPlugins` is
`com.intellij.java,org.intellij.plugins.markdown,JavaScript` — no Maven, no Gradle. So
`ExternalSystemUnlinkedProjectAware.getInstance(MAVEN)` and `…(GRADLE)` both return null in
tests, `available` is empty, and `linkBuildSystem` returns `PluginUnavailable` before touching
any of the logic this PR exists to add. The `||` in the assertion hides which branch ran.

I checked rather than assumed: I replaced the entire body of `linkBuildSystem` with

```kotlin
return LinkResult.PluginUnavailable("Maven/Gradle")
```

and the test results were byte-for-byte identical — same one failure, same message. Per the
repo's own rule, a test that still passes when the code under it is deleted should be deleted.

What I'd like instead: register a fake `ExternalSystemUnlinkedProjectAware` on the EP for the
duration of the test (the platform has `MockUnlinkedProjectAware` in its own test sources for
exactly this) and assert the state machine — no build file → `NoBuildFile`; build file present
and unlinked → `linkAndLoadProjectAsync` called once → `Linked`; already linked → `AlreadyLinked`
and the link call *not* made. That version fails if the loop breaks. Please also stop leaning on
`project.basePath` existing on disk in a light fixture, per the fixture rules in CLAUDE.md.

### 2. `linkAndLoadProjectAsync` blocks for the entire import — this needs the long-poll pattern

The description says the link "triggers a full dependency resolution in the background. Allow
10-30 seconds on large projects." That isn't what happens. Following the Maven path:

```
MavenUnlinkedProjectAware.linkAndLoadProjectAsync
  → MavenOpenProjectProvider.linkToExistingProjectAsync
  → doLinkToExistingProjectAsync
  → MavenProjectAsyncBuilder.commit(project, projectFile, null, syncProject = true)
```

`commit` is a suspend function that returns `List<Module>` — it awaits the import. Gradle's
`linkAndSyncGradleProject` → `GradleOpenProjectProvider().linkToExistingProjectAsync` has the
same shape. On a cold `~/.m2` or a large Gradle build this is minutes, not 10-30 seconds, and
`ide_link_build_system` has no `waitSeconds` and no poll id.

CONTRIBUTING is explicit: a tool that can block longer than ~45s MUST use the long-poll pattern,
because the client kills the call at its own timeout and the transport can't stream progress.
That's the rule that shaped `ide_run_tests` and `ide_build_project`, and it applies here for the
same reason. `LongPollRegistry` + `waitSeconds` + a `linkId`, same as those two.

Related: `tryAutoLink` is still called outside `withTimeoutOrNull`, so `timeoutSeconds` doesn't
bound it. The plan doc in this PR says "inside the timeout block" — the code doesn't do that.

### 3. We report "Linked" without checking that anything got linked

`linkAndLoadProjectAsync` returns `Unit`, and the code maps "no exception thrown" to
`LinkResult.Linked`. But Maven's implementation is:

```kotlin
if (ExternalSystemTrustedProjectDialog.confirmLinkingUntrustedProjectAsync(project, systemId, projectRoot.toNioPath())) {
  // ... import ...
}
```

On an untrusted project that's a modal dialog. If nobody answers it, the MCP call hangs on a
dialog only a human can dismiss; if it's declined, the function returns normally and we tell the
agent "Maven project linked — dependency resolution scheduled." when nothing was linked.

Two cheap fixes, and I'd like both: re-check `aware.isLinkedProject(project, path)` after the
call and return `Failed` when it's still false, and have `ide_link_build_system` mark the path
trusted the way `OpenProjectTool` already does (or at minimum say in the description that an
untrusted project blocks on a dialog).

### 4. `hasBuildFile` drops the read action

`UnlinkedProjectStartupActivity` has the same three lines, and wraps them:

```kotlin
private suspend fun ExternalSystemUnlinkedProjectAware.hasBuildFiles(project: Project, externalProjectPath: String): Boolean {
  return readAction {
    val projectRoot = LocalFileSystem.getInstance().findFileByPath(externalProjectPath)
    projectRoot?.children?.firstOrNull { isBuildFile(project, it) } != null
  }
}
```

Ours is identical minus the `readAction`. It survives today because `MavenUtil.isPomFile` wraps
its own PSI access in `ReadAction.compute` when it hits a non-`pom.*` XML file — but that's one
implementation being defensive, not the interface's contract, and any other EP implementation
(Ant, Bazel, Cargo, third-party) is free not to be. `linkBuildSystem` is already `suspend`, so
this is a two-line change.

## Non-blocking

**Cancellation is swallowed.** `catch (e: Exception)` in both `BuildSystemLinker.linkBuildSystem`
and `OpenProjectTool.tryAutoLink` catches `CancellationException` (it's an `Exception`) and
`ProcessCanceledException`. `McpToolDispatcher` documents that cancellation propagates and
handles PCE itself. Rethrow those two, or catch something narrower.

**`tryAutoLink` swallows three outcomes.** `PluginUnavailable`, `NoBuildFile` and `AlreadyLinked`
all map to `null`. An agent that explicitly passed `autoLink: true` on a Gradle project in an IDE
without the Gradle plugin gets "Project 'x' is open and ready." and no indication its request was
ignored. At least surface `PluginUnavailable`.

**No re-await of smart mode after linking.** A successful link kicks off re-indexing, and we
return "open and ready" on top of it — which is the one promise `ide_open_project`'s description
makes. Your own plan doc lists this as a step; the implementation skips it.

**Drop `docs/plans/2026-08-18-link-build-system.md`.** `docs/plans/` doesn't exist in this repo
and there's no precedent for checking plans in. It's 380 lines, and it already contradicts the
code it planned (timeout placement, re-awaiting smart mode, a test stub with an empty body). It
belongs in the PR conversation, not the tree.

**The PR description is stale.** It still describes the first implementation —
`ProjectUtils.importMavenModule` for Maven and `ExternalSystemUtil.refreshProjects` for Gradle.
Worth rewriting before merge, since it's what ends up quoted in the release notes.

## Summary

Architecture is right now; what's left is the safety rails. In rough priority:

1. Replace the behavior test with one that fails when `BuildSystemLinker` breaks, and fix the
   isolation failure.
2. Put `ide_link_build_system` behind the long-poll pattern, and move `tryAutoLink` inside the
   timeout.
3. Verify the link took effect before reporting `Linked`; handle the trust dialog.
4. `readAction` around `hasBuildFile`; stop swallowing cancellation; drop the plan doc; refresh
   the description.

Verified locally on `147341c`: compiles clean, `./gradlew clean test --no-build-cache` green,
`./gradlew test --tests "*LinkBuildSystemToolBehaviorTest*"` red 2/2.

PR #329, opened 2026-08-20
feat: add path/directory scoping (`paths`) to the search tools (#328)

### LINKED ISSUE #328: Add path/directory scoping (`paths`) to the search tools
## Problem

No search tool can be restricted to a directory or path glob.

`ide_search_text` offers only `filePattern`, an IntelliJ file mask matched on the file **name**
(`"*.kt"`, `"*.java,!*Test.java"`). It cannot express a path segment, and the tool has no `scope`
parameter at all. `ide_find_references` and `ide_structural_search_replace` do have `scope`, but it is
an enum — `project_files`, `project_and_libraries`, `project_production_files`,
`project_test_files` — with no path granularity. So a query like "find this string under
`tools/refactoring/` but not under `tools/project/`" cannot be written as one call on any of them.

On a large project this means either a project-wide search whose results must be filtered client-side
— paying for every irrelevant hit in tokens — or several narrower searches that each guess at a
filename mask. `ide_find_file` has the same limitation from the other direction: its matcher runs on
the file name, so a path glob cannot be used to locate files either.

## Proposed tool or behavior

Add an optional `paths` parameter to `ide_search_text`, `ide_find_references` and
`ide_structural_search_replace`:

- Type: array of strings. Each entry is a project-relative glob. A leading `!` excludes.
- Example: `["src/main/kotlin/**/handlers/**", "!**/*Test.kt"]`.
- Semantics: includes are unioned, then excludes are subtracted. Omitting `paths` keeps today's
  behaviour exactly.
- Composes with, rather than replaces, the existing `scope` enum and `filePattern`.

Suggested implementation: resolve each glob to `VirtualFile` directories/files and build a
`GlobalSearchScope` (`GlobalSearchScopesCore.directoriesScope`, or a filter scope for file-level
globs), intersected with the scope the tool already computes in its `createFilteredScope` path. An
unresolvable glob should be reported in the response rather than silently ignored, so a typo cannot
look like "no matches".

Prior art: the `idea` MCP server built into the IDE takes exactly this shape — `search_text` and
`search_regex` accept a `paths` array of project-relative globs with `!` excludes, and
`search_structural` takes `directoryToSearch`. Adopting the same parameter name and semantics keeps
the two servers' contracts aligned.

## Alternatives considered

- `filePattern` — a filename mask; supports name-level excludes (`!*Test.java`) but cannot express a
  directory.
- `scope` — only production/test/library splits, no path granularity, and absent from
  `ide_search_text` entirely.
- Client-side filtering of project-wide results — works, but pays for every discarded hit, and
  interacts badly with pagination: a page can be entirely filtered away, which looks like an empty
  result.
- A single `directory` string instead of an array — simpler, but cannot express excludes, which is
  the common case (skip generated output, skip tests).

:robot:

hechtcarmel:
## Summary

Closes #328.

No search tool could be restricted to a directory. `filePattern` is an IntelliJ mask matched on the file **name**, and `scope` offers only production/test/library splits — so *"search under `tools/refactoring/` but not `tools/project/`"* could not be written as one call. The fallback, a project-wide search filtered client-side, pays tokens for every discarded hit and interacts badly with pagination: a whole page can be filtered away and look like an empty result.

This adds an optional **`paths`** parameter to `ide_search_text`, `ide_find_references` and `ide_structural_search_replace` — an array of project-relative globs where a leading `!` excludes:

```json
{"query": "createScope", "paths": ["src/main/kotlin/**/handlers/**", "!**/*Test.kt"]}
```

`*` matches within a path segment, `**` crosses directories (zero or more whole segments), `?` is one non-separator character, and a plain directory path includes everything beneath it. Includes are unioned, then excludes subtracted; with only excludes, everything else is searched. Omitting `paths` keeps today's behaviour exactly. The parameter name and semantics match the `paths` argument of the IDE's built-in MCP server, as the issue suggested, keeping the two servers' contracts aligned.

### Implementation

Follows the scope-filter pattern already in `handlers/` (`ExcludedPathScope`, `GeneratedSourcesExcludingScope`): `PathGlobScope` is a `DelegatingGlobalSearchScope` intersected with whatever scope each tool already computes. So `paths` **composes with** `scope`, `filePattern` and `includeGenerated` rather than replacing them, and IntelliJ never resolves PSI for filtered-out files.

Filtering in the scope rather than post-filtering results is what makes two behaviours fall out correctly:

- `ide_find_references` applies it to `totalCount` and to paginated follow-up pages, not just the first page's contents (the cursor's `searchExtender` carries the matcher).
- `ide_structural_search_replace` restricts what it **rewrites**, not merely what it reports.

Relative paths are computed the way the tools compute the paths they return — project base path first, then the longest matching module content root — so globs behave correctly in workspace/multi-module projects. Library files outside every project root are outside a project-relative glob's domain and never satisfy an include.

An include glob whose literal directory prefix does not resolve in the project returns an error naming the glob, so a typo cannot masquerade as *"no matches"* (called out in the issue). Excludes are not existence-checked — defensively excluding a directory that isn't there is legitimate.

One subtlety worth a reviewer's eye: `PathGlobScope` passes the filter's pattern set to `DelegatingGlobalSearchScope`'s `equality` parameter. Without it, two scopes over the same base scope with *different* globs compare equal (that constructor's `equals` looks only at the delegate and this discriminator), and `GlobalSearchScope.intersectWith`/`union` short-circuit on equality — so one filter could silently stand in for another.

### Tests

- `PathGlobMatcherUnitTest` (16 cases) pins the glob dialect: segment-local vs cross-directory wildcards, `**` matching zero segments, exclude subtraction, exclude-only filters, separator normalization, case sensitivity, regex metacharacters staying literal (`a+b`, `v1.2`), literal-prefix extraction, null-path handling, and equality-key distinctness.
- `SearchTextPathsBehaviorTest`, `FindUsagesPathsBehaviorTest`, `StructuralSearchReplacePathsBehaviorTest` drive the tools end to end with real files and source roots. The replace-mode test asserts **both directions on both files** — a filter that only trimmed reported matches while still rewriting everything would fail it.

Golden tool manifest updated for the three tools' new parameter and description lines.

## Checklist

- [x] `CHANGELOG.md` entry added under `[Unreleased]` only (no `## [x.y.z]` version sections)
- [x] `./scripts/check-pr.sh` passes — 15/15 static gates; its `./gradlew test` step could not run here, see below
- [ ] **`./gradlew test` could not be run in the authoring environment** — the sandbox's network policy returns 403 for `download.jetbrains.com` and `cache-redirector.jetbrains.com`, so the `idea:idea:2025.3` platform dependency cannot resolve and no Gradle compile/test task can start. Verified as far as the environment allowed instead: the real `PathGlobScope.kt` was compiled against minimal stubs of the four platform classes it touches, and the real `PathGlobMatcherUnitTest` was compiled and run against it — 16/16 pass. The platform-tier tests and the full compile are unverified locally and need CI (or a maintainer run) to confirm.
- [x] No new tools (a new parameter on three existing tools), so no `ToolNames`/`ToolRegistry`/`McpSettings` registration or settings migration applies; golden manifest regenerated and its diff reviewed
- [x] No forbidden files (`.idea/gradle.xml`, `scripts/build-install.sh`, `docs/pr-*.md`)

---
_Generated by [Claude Code](https://claude.ai/code/session_01HrKj6u2nDnUfBpbH2SW2q9)_


hechtcarmel:
Went through both Qodana findings. Fixed one, declining the other — reasoning below so it's easy to overrule.

**`Unnecessary non-capturing group` (Warning) — fixed in 271a453.**

Qodana is right. The group wrapped the generated fragment in `^(?:<fragment>)(?:/.*)?$`, guarding against the fragment containing a top-level alternation (which would make `^` bind to the first branch alone). That can't happen: `globToRegex` only emits concatenation, because a literal `|` in a glob is escaped via `REGEX_METACHARS`.

Rather than keep a group that reads as redundant, I removed it and left a comment naming the invariant, so a future edit that makes `globToRegex` emit alternation has a visible tripwire instead of an implicit one. Behavior is unchanged — the 16 `PathGlobMatcherUnitTest` cases still pass. The remaining `(?:/.*)?` group carries the `?` quantifier and stays.

**`Redundant escaped dollar characters in string literals` (Notice) — not changing.**

This is the trailing `\$` in SSR patterns like `"SsrpSearchApi.call(\$arg\$)"`. Technically correct — `$)` isn't a template start, so that second escape is unnecessary — but `\$arg\$` mirrors the `$var$` delimiter pair SSR actually uses, and it's the established convention here: 10 pre-existing instances repo-wide, 6 of them in `StructuralSearchReplaceBehaviorTest`, the file directly alongside the one I added. Since Qodana runs in PR mode it only sees my new file, so "fixing" it would leave the new test inconsistent with its sibling for a notice-level style point. Happy to change it if you'd rather have it lint-clean, but it'd be worth doing across all 11 together.

---

PR #339, opened 2026-08-25
ide_run_tests : use caller's timeoutSeconds as process-start deadline instead of hardcoded 15s

jevanlingen:
## Summary
Replace the hardcoded `PROCESS_START_TIMEOUT = 15.seconds` process-start guard in `ide_run_tests` with the caller-supplied `timeoutSeconds` value.

## Why
The 15-second constant was an arbitrary safety guard to prevent `processHandlerDeferred.await()` from blocking indefinitely if a process silently failed to start. That intent is fine, but the fixed value breaks large/slow Maven projects where the JVM and test classpath simply take longer than 15 seconds to spin up.

More importantly, the resulting error is misleading when the caller passes a larger timeout:

**Tool call:**
```
Tool: ide_run_tests
Status: ERROR
Duration: 15038ms

Parameters:
{
  "target": "com.example.service.SomeServiceImplTest#someMethod",
  "project_path": "/Users/dev/workspace/some-project",
  "timeoutSeconds": 120
}

Error:
Test process did not start within 15 seconds for 'SomeServiceImplTest.someMethod'.
```

The caller declared a 2-minute budget; the plugin aborted after 15 seconds. An LLM receiving this error has no reason to suspect the process-start window — it looks like the timeout the caller set was hit, which it wasn't.

Using `timeoutSeconds.seconds` as the deadline preserves the safety guard (the call still fails if the process never starts) while honouring the caller's actual budget and producing an accurate error message.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **`ide_run_tests` no longer aborts with a misleading "did not start within 15 seconds" error on slow Maven/Gradle projects** — the hardcoded 15s process-start guard is replaced by the remaining `waitSeconds` budget (~45–55s by default), so large projects that take longer than 15s to compile and launch the test JVM no longer fail spuriously. The error message now also tells the agent to retry or raise `waitSeconds`.
+
 ## [5.8.2] - 2026-08-23
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Fixed
 
-- **`ide_run_tests` no longer aborts with a misleading "did not start within 15 seconds" error on slow Maven/Gradle projects** — the hardcoded 15s process-start guard is replaced by the remaining `waitSeconds` budget (~45–55s by default), so large projects that take longer than 15s to compile and launch the test JVM no longer fail spuriously. The error message now also tells the agent to retry or raise `waitSeconds`.
+- **`ide_run_tests` no longer aborts with a misleading "did not start within 15 seconds" error on slow projects** — the hardcoded 15s process-start guard is replaced by the remaining `waitSeconds` budget (~45–55s by default). The error message now also tells the agent to retry or raise `waitSeconds`.
 
 ## [5.8.2] - 2026-08-23
 
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/project/RunTestsTool.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/project/RunTestsTool.kt
@@ -58,7 +58,6 @@ class RunTestsTool : AbstractMcpTool() {
     companion object {
         private val LOG = logger<RunTestsTool>()
         private const val DEFAULT_TIMEOUT_SECONDS = 120
-        private val PROCESS_START_TIMEOUT = 15.seconds
 
         /** Grace period to let the IDE's test tree finalize after the process exits. Normally instant. */
         private val TEST_TREE_FINALIZE_TIMEOUT = 10.seconds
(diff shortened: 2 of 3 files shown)


hechtcarmel:
Hi,
Thank you for submitting this PR.
Claude found some issues with it  that I agree with and will be posting soon

hechtcarmel:
> _Review by Claude Code (Fable), requested by the maintainer._

Thanks for this PR — the diagnosis is spot-on, but the fix as written trades a misleading error for a protocol-breaking one, so we can't merge it in this form. Details below.

## What's right

The problem is real and worth fixing. The hardcoded 15s guard is genuinely too aggressive: in IntelliJ, "before the process starts" includes the pre-run build step (JPS/Gradle/Maven compile before the test JVM launches), which routinely exceeds 15 seconds on large projects. And the current error message — "did not start within 15 seconds" when the caller asked for 120 — is actively misleading to an agent, exactly as the PR description argues.

## Why the implementation can't be merged as-is

The process-start wait (`withTimeoutOrNull(...) { processHandlerDeferred.await() }` in `startRun`) runs **synchronously inside the tool call**, *before* the run is registered in `ActiveTestRunRegistry` and before the `waitSeconds` budget applies. The entire `LongPoll` architecture (issue #277, see `tools/LongPoll.kt`) exists because no single call may block past ~45–55s: MCP clients time out at 60s by default, and the stateless Streamable HTTP transport can't send keep-alives.

With this change, a slow start blocks the HTTP call for up to `timeoutSeconds` — 120s by default, and the tool's own docstring shows an example with `timeoutSeconds: 7200`. Whenever a start takes longer than ~60s:

1. **The MCP client aborts the request.** The agent gets no response at all — strictly worse than today's misleading-but-delivered error.
2. **The run is orphaned.** It was launched via `restartRunProfile` but never registered, so the agent never receives a `runId`, no watchdog will ever kill the process at `timeoutSeconds`, and the results are never collectible.

There's also a semantic problem: `timeoutSeconds` is documented as "maximum seconds the whole test **run** may take," and its clock deliberately starts at process start (see the comment in `startRun`: "config resolution and process spawn-up must not be billed to the run"). Reusing the same number as the start deadline double-spends that budget and quietly changes the parameter's meaning — worst case, a call could now block `timeoutSeconds` on start *plus* `timeoutSeconds` of run time.

Note that CI won't catch any of this: nothing in the suite covers the start-timeout path, which is why the build is green.

## Suggested direction

Either of these would preserve your intent while keeping the 45–55s contract intact:

- **Minimal fix:** bound the start wait by `min(timeoutSeconds, remaining wait budget)` — the same arithmetic `finalizeWaitMillis` already does with `waitSeconds` and `callStartMs`. That raises the effective window from 15s to ~45–55s (covering most slow starts), never breaches the client timeout, and lets the error message be honest: "did not start within Ns — the IDE may still be compiling; retry, or raise waitSeconds (max 55)."
- **Complete fix (more work):** let `ActiveTestRunRegistry` register a run in a "starting" state (handler as a deferred), so a slow start returns `{"status": "running", "runId": ...}` like any other long wait and polls attach across the start phase. This is the only way to survive starts longer than 55s, which the motivating scenario (large Maven project, 120s budget) plausibly includes.

Two checklist items from CONTRIBUTING.md either way: a `CHANGELOG.md` entry under `[Unreleased]` (this is a user-visible behavior change), and a unit test on the deadline computation (extracting it into a small `internal fun` alongside `finalizeWaitMillis` would make it testable in the unit tier).

Happy to see this land with the minimal fix — the writeup in the PR description was excellent and made the review easy.

---

jevanlingen:
Ah, nice Claude Fable review! I intentionally went with the simpler fix because it directly solves the problem I was running into 😅 . My build startup time was typically around 20 to 30 seconds, so increasing the process-start timeout already removes the main pain point.

I also think with a more accurate timeout error, LLMs are more likely to retry `ide_run_tests` instead of falling back to terminal commands. In my case, the agent started running `./mvnw test -Dtest=TestCircle#xyz` immediately, even though simply retrying the tool would have been preferable since compilation had already completed.

The more complete solution is probably cleaner, but it also adds complexity to the code base. For this particular use case, I'm wondering if the additional complexity is actually justified.

hechtcarmel:
Thanks. 
I'll be merging it and it'll be part of the next version sent to jetbrains review later today. 
Anywhere between 10 minutes to 5 days til lthey approve and its available on the marketplace

PR #343, opened 2026-08-27
Suggest ide_diagnostics on tool execution failures

jevanlingen:
## Summary

When a tool execution fails due to an unexpected IDE-side error, agents currently receive a message like:

```
Tool: ide_run_tests
Status: ERROR

Parameters:
{"target":"com.some.project.ExampleTest","project_path":"<path>"}

Error:
Test process failed to start for 'ExampleTest'.
```

There is no next step. The agent stops, retries blindly or even starts using a different tool.

## Why

- Reduces dead ends: agent gets an actionable next step instead of stopping
- No false claims: "possibly more information" rather than "compilation errors" (we genuinely don't know the cause)
- Consistent: same hint across all similar failure points

## Implementation note

After investigation, `processNotStarted` (the IntelliJ `ExecutionListener` callback that triggers this error) fires in three scenarios: before-run tasks fail (e.g. a "Make" compilation step), the runner throws an `ExecutionException`, or the process creation itself fails. The most common real-world trigger is compilation errors — and `ide_diagnostics` is exactly the right tool to surface those.

Crucially, the public `ExecutionListener.processNotStarted` signature carries **no throwable** — we cannot tell the agent *why* it failed. So we don't claim to know: we just suggest the tool that has the best chance of giving more information.

The same pattern applies to other tool execution failures where the cause is opaque and code state may be involved.

## Changes

Added mention of using the ide_diagnostics tool in the execution-failure error messages in:

- `ide_run_tests` — `processNotStarted` callback and its fallback
- `ide_refactor_rename` — rename execution failure
- `ide_move_file` — move execution failure
- `ide_change_signature` — preparation failure and execution failure
- `ide_convert_java_to_kotlin` — conversion failure
- `ide_refactor_safe_delete` — safe delete and file deletion failures
- `ide_optimize_imports` — optimize imports failure
- `ide_reformat_code` — reformat failure

Input-validation errors (file not found, missing parameters, invalid values) are intentionally left unchanged — those messages are already clear and `ide_diagnostics` would not help.

## Test
I built the plugin locally and deliberately removed an import. I then asked Claude to run a test. After the test failed, Claude automatically ran `ide_diagnostics`, identified and fixed the issue, and reran the test successfully. This is exactly the workflow that this MR is intended to enable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@
 
 ### Fixed
 
+- **Tool execution failures now suggest running `ide_diagnostics`** — when `ide_run_tests`, `ide_refactor_rename`, `ide_move_file`, `ide_change_signature`, `ide_convert_java_to_kotlin`, `ide_refactor_safe_delete`, `ide_optimize_imports`, or `ide_reformat_code` fail during execution (not due to bad input), the error message now appends "Run ide_diagnostics for possibly more information." so the agent has an immediate next step rather than stopping.
+
+## [5.8.3] - 2026-08-26
+
+### Fixed
+
 - **`ide_run_tests` no longer aborts with a misleading "did not start within 15 seconds" error on slow projects** ([#339](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/339)) — the hardcoded 15s process-start guard is replaced by the remaining `waitSeconds` budget (~45–55s by default). The error message now also tells the agent to retry or raise `waitSeconds`.
 
 ## [5.8.2] - 2026-08-23
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Changed
 
-- **Tool execution failures now suggest running `ide_diagnostics`** — when `ide_run_tests`, `ide_refactor_rename`, `ide_move_file`, `ide_change_signature`, `ide_convert_java_to_kotlin`, `ide_refactor_safe_delete`, `ide_optimize_imports`, or `ide_reformat_code` fail during execution (not due to bad input), the error message now appends "Run ide_diagnostics for possibly more information." so the agent has an immediate next step rather than stopping.
+- **Tool execution failures now suggest running `ide_diagnostics`** — when `ide_run_tests`, `ide_refactor_rename`, `ide_move_file`, `ide_change_signature`, `ide_convert_java_to_kotlin`, `ide_refactor_safe_delete`, `ide_optimize_imports`, or `ide_reformat_code` fail during execution (not due to bad input), the error message now appends "Run ide_diagnostics for more details" so the agent has an immediate next step rather than stopping.
 
 ## [5.8.3] - 2026-08-25
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,11 @@
 
 ## [Unreleased]
 
-## [5.8.3] - 2026-08-25
-
-### Fixed
+### Changed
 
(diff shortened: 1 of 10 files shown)


hechtcarmel (changes requested):
> 🤖 **This is a Claude review** — this comment was written by Claude Code and posted at the maintainer's request.

Thanks for this PR — the idea is genuinely good and we'd like to merge it. It matches a pattern this codebase already uses (`AbstractMcpTool` and `RefactoringScopeGuard` steer agents to `ide_sync_files`; `ReloadProjectTool` steers to `ide_link_build_system`), `ide_diagnostics` is enabled by default with an optional `file` param so the hint is actionable, and limiting the hint to opaque execution failures — with the honest "possibly" instead of a false claim about the cause — is exactly the right scoping. The manual end-to-end validation (break an import → test fails → agent runs `ide_diagnostics` → fixes → green) is appreciated.

Three changes requested before merge:

### 1. The `ChangeSignatureTool` prep hunk breaks CI and contradicts the PR's own scoping

CI's Test job failed with exactly one failure out of 1113: `ChangeSignatureBehaviorTest.testChangeSignatureOnNonMethodFails` ([log](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/actions/runs/33087656412/job/98573926680)):

```
expected:<...or on a method name.[]> but was:<...or on a method name.[. Run ide_diagnostics for possibly more information.]>
```

That's the `prep.isFailure` branch (`ChangeSignatureTool.kt:155`). The failures `prepareChange` returns are input-validation errors — "Line N is out of range (file has M lines)", "No element found at line N, column M", "No method found at line N, column M. Position the cursor on a method name." — precisely the category the PR description says was intentionally left unchanged, and where the hint misleads: the agent should fix its coordinates, not run diagnostics. The assertion output also shows the double period this append produces ("…method name.**.** Run …").

**Fix:** drop the `prep.isFailure` hunk entirely. The `applyChange` catch hunk (`ChangeSignatureTool.kt:371`) is a genuine opaque execution failure — keep that one. This resolves the CI failure and the scoping issue at once, with no test edits needed.

### 2. The hint fires twice on the PR's flagship path in `RunTestsTool`

The hint is baked into the `processNotStarted` exception (`RunTestsTool.kt:491`), and that exception is rethrown by `processHandlerDeferred.await()` into the `catch (e: Exception)` at `RunTestsTool.kt:327-330` — which appends the hint again. So the exact scenario this PR was written for produces:

> Test process failed to start for 'X'. Run ide_diagnostics for possibly more information. Run ide_diagnostics for possibly more information.

**Fix:** keep the hint in exactly one of the two places. The catch block is the better home, since it also covers exceptions thrown by `restartRunProfile` itself.

### 3. Changelog: drop the manually added `## [5.8.3]` section

`./scripts/check-pr.sh` flags this directly: *"New versioned release entry added — only the maintainer creates ## [x.y.z] sections"* (see CONTRIBUTING.md). The release automation already opened #341 for exactly this move — with the correct date (the 5.8.3 tag was created Aug 25, not Aug 26) — so as-is this PR is guaranteed to conflict with it.

**Fix:** keep only your `[Unreleased]` entry; leave moving released entries to the release flow. (Minor: `### Changed` fits the entry slightly better than `### Fixed`, since this is a behavior improvement rather than a bug fix.)

### Minor, non-blocking

- The `". Run …"` suffix produces a double period whenever the underlying message already ends with one — IntelliJ's refactoring-hint messages often do. A small shared helper (e.g. an `appendDiagnosticsHint(message)` that trims a trailing period before appending) would normalize punctuation and replace the 11 copies of the literal.
- Phrasing: "for possibly more information" reads awkwardly; consider "Run ide_diagnostics for more details." or "…which may reveal the cause (e.g. compilation errors)."

With 1–3 addressed and CI green, this looks good to merge. Running `./scripts/check-pr.sh` before pushing will catch the changelog rule and run the test suite in one go.

---

jevanlingen:
I addressed all review comments. In the end I moved the hint to the `createErrorResult` function. I think that's both cleaner and more developer friendly.

hechtcarmel:
Thank, merging and sending to jetbrains for review

PR #354, opened 2026-08-31
Move tool enable/disable to a new Exposed Tools settings page

jevanlingen:
## Summary

Moves the "Available Tools (uncheck to disable)" checkbox list off the main settings page onto a new child page at Settings → Tools → Index MCP Server → **Exposed Tools**.

- New `McpToolsConfigurable` child page registered in `plugin.xml` under `mcp.settings`, holding the per-tool checkboxes, tooltips, and the "Server is initializing..." guard exactly as before
- `McpSettingsConfigurable` keeps the server host/port, history, projects mode, response format, sync, and lifecycle sections unchanged
- No stored state changes — existing per-tool enable/disable settings carry over untouched
- New platform tests cover the initialized and uninitialized-server paths
- Docs updated: CHANGELOG, README, `docs/mcp-kotlin-sdk-migration.md`, `smoke-tests/mcp-protocol.md`

## Why

This plugin now exposes many tools, making the main interface increasingly cluttered. I’ve moved the tools out of the main plugin screen and into a separate page, mirroring the IDE’s own MCP Server settings layout. Possible next steps could be adding categories and displaying the description as a secondary column, similar to the official MCP Server interface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- **Tool enable/disable moved to a new "Exposed Tools" settings page** — the "Available Tools (uncheck to disable)" checkbox list left the main settings page and now lives on a child page at Settings → Tools → Index MCP Server → Exposed Tools, mirroring the IDE's own MCP Server settings layout. The main page keeps the server host/port, history, projects mode, response format, sync, and lifecycle sections unchanged; the per-tool checkboxes, tooltips, and "Server is initializing..." guard behave exactly as before, and no stored state changes — existing per-tool enable/disable settings carry over untouched.
+
 ## [5.9.2] - 2026-08-29
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ enable the write tools.
 - [Quick Start](#quick-start)
 - [Community Integrations](#community-integrations)
 - [Client Configuration](#client-configuration)
-- [Available Tools](#available-tools)
+- [Exposed Tools](#exposed-tools)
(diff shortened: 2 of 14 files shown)


hechtcarmel:
Thanks for this — the move is well executed: the checkbox behavior carries over verbatim, no stored state changes, and the new tests drive the real `Configurable` lifecycle including the `plugin.xml` registration. One change is needed before this can merge:

**Please update `McpToolDispatcher.kt` to the new settings path.** The runtime error a client receives when calling a disabled tool still points at the old location:

```kotlin
// src/main/kotlin/.../server/mcp/McpToolDispatcher.kt:90
"Tool '$toolName' is disabled. Enable it in Settings → Index MCP Server → Available Tools."
```

That string is exactly what coding agents relay to users when a tool is disabled, and after this PR it directs them to a page where the checkboxes no longer exist. The PR already updated the snippet quoting this code in `docs/mcp-kotlin-sdk-migration.md`, so the doc currently shows the new text while the actual source still carries the old one. Concretely:

- `McpToolDispatcher.kt:90` — error message → `Settings → Tools → Index MCP Server → Exposed Tools`
- `McpToolDispatcher.kt:36` — the KDoc line with the same old path

Two nice-to-haves, fine here or as a follow-up:

- `USAGE.md` repeats "enable in Settings > Tools > Index MCP Server" for each disabled-by-default tool (~30 occurrences) — those could take the `> Exposed Tools` suffix too.
- The new page is titled "Exposed Tools" but its internal header still reads "Available Tools (uncheck to disable)" — unifying on one name would be tidy.

With the dispatcher fix in, this looks good to merge.

---

PR #365, opened 2026-09-03
fix: restore wildcard imports removed by cross-module same-package moves

### LINKED ISSUE #360: ide_move_file removes valid wildcard imports on cross-module same-package moves
## Problem

When moving files between Maven modules that share the same Java package name (e.g., extracting `io.casehub.blocks.summarisation` types from `blocks/` to `summarisation-api/`), `ide_move_file` incorrectly removes `import io.casehub.blocks.summarisation.*` from consuming files.

IntelliJ's `MoveFilesOrDirectoriesProcessor` treats the moved types as "leaving" the package even though the package name is unchanged — only the Maven artifact changed. The tool reports success (`affectedFiles` includes the consuming files) but the import removal causes `cannot find symbol` at the next build. No warning is emitted.

## Root cause

This is an IntelliJ platform behavior, not a bug in the plugin's code. `MoveFilesOrDirectoriesProcessor` processes import updates based on the file's location in the project structure (content roots), not the Java package declaration. When a file moves to a different content root that happens to have the same package structure, the processor sees it as a package change and removes/updates imports accordingly.

## Reproduction

1. Have two Maven modules with the same package: `module-a/src/main/java/com/example/pkg/` and `module-b/src/main/java/com/example/pkg/`
2. `module-a` has `Foo.java` in `com.example.pkg`
3. `module-c` uses `import com.example.pkg.*` and references `Foo`
4. Call `ide_move_file(file="src/main/java/com/example/pkg/Foo.java", destination="../../module-b/src/main/java/com/example/pkg/")`
5. The wildcard import in `module-c` is removed even though `Foo` is still in `com.example.pkg`

## Suggested fix

After a same-package cross-module move, the tool should detect removed wildcard imports for the unchanged package and either:
1. Restore them (active fix)
2. Warn about them in the result (passive fix)

The passive fix is safer — add to the result's `warnings` list:
> "Wildcard imports for package 'com.example.pkg' were removed from N file(s) during the move. The package is unchanged — you may need to restore these imports."

mdproctor:
## Summary

Closes #360. Workaround for a platform bug — upfront about that.

When moving a file between Maven/Gradle modules that share the same Java package name (e.g., extracting `io.casehub.blocks.summarisation` types from `blocks/` to `summarisation-api/`), IntelliJ's `MoveFilesOrDirectoriesProcessor` incorrectly removes `import io.casehub.blocks.summarisation.*` from consuming files. It treats the moved types as "leaving" the package because the content root changed, even though the `package` declaration is identical. The tool reports success but the next build fails with `cannot find symbol`.

### How it works

1. **Before the move**: if the source file is a `PsiJavaFile`, records its package name and snapshots which project files have wildcard (`.*`) imports for that package
2. **After the move**: if the destination file has the same package declaration, checks whether any of the snapshotted wildcard imports were removed
3. **Restores**: re-adds removed wildcard imports via `PsiElementFactory.createImportStatementOnDemand` and reports each restoration as a warning in the result

### Limitations

- This is a workaround for `MoveFilesOrDirectoriesProcessor` behavior, not a fix for the root cause. The platform removes the import inside `performRefactoring` where the plugin cannot intercept.
- The snapshot scans all Java source files in the project via `ProjectRootManager.contentSourceRoots`. On very large monorepos this could be slow.
- The light test fixture cannot reproduce the bug because it has no real Maven module boundaries — the test verifies the import survives (the happy path) rather than proving the restoration logic fires. A proper integration test would need a multi-module Maven setup.

We accept this may not meet the bar for inclusion — it's a pragmatic workaround for a real pain point in agentic workflows, not a clean fix. Happy to iterate on refinements.

## Test plan

- [x] `testSamePackageMovePreservesWildcardImports` — verifies wildcard import survives a same-package cross-root move (passes in test fixture where the platform doesn't remove the import; would catch regressions if the restoration logic breaks)
- [x] Manual reproduction: moved types between casehub modules sharing `io.casehub.blocks.summarisation` — wildcard imports preserved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **`ide_move_file` now restores wildcard imports removed by cross-module same-package moves** ([#360](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/issues/360)) — when moving a file between Maven/Gradle modules that share the same Java package name, IntelliJ's move processor incorrectly removes `import pkg.*` from consuming files. The tool now snapshots wildcard imports before the move, detects when the package is unchanged, and restores any that were removed, reporting each restoration as a warning.
+
 ## [5.9.4] - 2026-09-01
 
 ### Fixed
--- a/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/refactoring/MoveFileTool.kt
+++ b/src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/refactoring/MoveFileTool.kt
@@ -250,6 +250,13 @@ open class MoveFileTool : AbstractRefactoringTool() {
                 val filePointer = SmartPointerManager.createPointer(preparation.psiFile)
                 val modifiedFilesBeforeMove = collectUnsavedProjectFiles(project)
 
+                val sourcePackage = detectJavaPackage(preparation.psiFile)
+                val wildcardImportSnapshot = if (sourcePackage != null) {
+                    snapshotWildcardImports(project, sourcePackage)
+                } else {
+                    emptyMap()
+                }
+
                 when (preparation.backend) {
                     MoveBackend.GENERIC_FILE_MOVE -> conflictWarnings = executeGenericFileMove(preparation)
                     MoveBackend.PHP_SEMANTIC_MOVE -> executePhpSemanticMove(project, preparation)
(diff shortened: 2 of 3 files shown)


hechtcarmel:
Thanks for the PR, and for the detailed issue that made the reproduction possible. This landed independently in #362 (released in 5.9.5) a few hours before you opened this, via a `MoveUsageGuard` hook that snapshots and restores imports in the usage files the move processor rewrites. It covers the same case, also restores single-class imports, stays gated on the Java plugin so non-Java IDEs are unaffected, and runs inside the move's write action rather than as a second pass. I'm going to close this as superseded.

One favor: #362 could not reproduce the import removal on 2025.3, so the guard there is a safety net. Could you retry your casehub move on 5.9.5 and confirm whether the wildcard import survives, and whether the result now shows a "restored" warning? If it still breaks, that would point to a different root cause (likely an optimize-imports pass while the moved class was unresolvable), and I'd want to chase that.

---

PR #372, opened 2026-09-09
Add semantic handles and refactoring previews

RomanPozdeev:
## Summary

  Allow MCP clients to reuse semantic targets across navigation and refactoring calls, preview refactorings before applying them, and consume large results
  incrementally.

  - Add project- and session-scoped `symbolId` handles backed by smart PSI pointers, with bounded retention and explicit expiry.
  - Introduce unified `target` selectors: symbol handle, file position, or qualified name and language.
  - Add non-mutating `dryRun` previews for rename, safe delete, and change signature, including applicability, affected files, usage/conflict counts, and
  warnings.
  - Add bounded, deterministic hierarchy pagination with continuation cursors and preserved declaration identity.
  - Extend `ide_file_structure` with structured nodes, source ranges, and optional symbol handles.
  - Support multi-file diagnostics with a shared deadline, per-file freshness and timeout metadata, and explicit truncation reporting.

  Also validate exact symbol rebinding after member replacement, synchronize deleted paths within the selected project root, improve Kotlin declaration
  metadata, and fix the Ktor linkage error that prevented MCP server startup.

  Update documentation, smoke-test instructions, golden contracts, and behavior tests for the new functionality.

  ## Compatibility

  Existing selectors and legacy response fields remain supported. Symbol handles and pagination cursors belong to the active project/server session.

  Reconnect MCP clients after updating the plugin to refresh cached tool schemas.

  ## Checklist

  - [x] Changelog entries added under `[Unreleased]`
  - [x] `./scripts/check-pr.sh` passes on this branch
  - [x]  Full `./gradlew test` suite passes on this branch
  - [x] No new MCP tool names; existing schemas, documentation, and golden contracts updated
  - [x] No forbidden files included

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,30 @@
 
 ## [Unreleased]
 
+### Added
+
+- **Opaque, server-owned `symbolId` handles for semantic navigation and refactoring** — symbol/class search, definition, references, symbol info, implementations, and hierarchy results now expose reusable IDs backed by IntelliJ `SmartPsiElementPointer`s. Symbol-targeting semantic and refactoring tools accept the ID instead of coordinates, preserve or replace it with current `updatedSymbol` metadata after a refactoring, and explicitly invalidate it after safe delete. Handles are scoped to the exact open project and MCP server generation, bounded by a 4,096-entry access-order LRU and a one-hour inactivity TTL; deleted/unrestorable PSI returns `SYMBOL_ID_EXPIRED` without selecting a nearby declaration. The handle is intentionally non-canonical: one PSI declaration may have several unequal, simultaneously valid IDs, while reusing one concrete ID remains stable across line shifts and rename.
+- **Additive structured targets** — target-aware semantic and refactoring tools accept exactly one nested `target` variant: `{symbolId}`, `{position:{file,line,column}}`, or `{qualifiedName,language}`. Runtime validation rejects mixed/ambiguous selectors, the dispatcher routes by nested `target.symbolId`, and all legacy top-level selectors remain supported.
+- **Non-mutating refactoring previews** — `ide_refactor_rename`, `ide_refactor_safe_delete`, and `ide_change_signature` accept `dryRun: true` and return a common contract with applicability, current target metadata, the planned change, estimated affected files, usage/conflict counts, warnings, and elapsed time. Preview performs resolution and conflict discovery without entering the refactoring/source-write phase, saving documents, mutating source, or creating an undo command.
+- **Bounded hierarchy pagination** — type and call hierarchy tools enforce `maxNodes` during deterministic breadth-first traversal and return `returnedNodes`, `truncated`, `elapsedMs`, `hasMore`, and an opaque `cursor`. Continuation frontiers are smart-pointer backed, scoped to the exact session/project, and kept in an independent cache of at most 128 snapshots with a ten-minute inactivity TTL; pages are the progress contract for stateless HTTP. Type hierarchy keeps the legacy `supertypes`/`subtypes` arrays and adds `traversal: [{direction, element}]` so the combined BFS order is observable without breaking existing clients.
+- **Structured file outlines** — `ide_file_structure` preserves the formatted `structure` string and adds `nodes` with name, semantic kind, signature, modifiers, source range, children, and an optional `symbolId` bound to the exact extracted PSI element.
+- **Multi-file `ide_diagnostics`** — a non-empty `files` array (up to 100 supplied paths, with lexical aliases analyzed only once) is available as an alternative to `file`. All requested files share one analysis deadline, problems are aggregated, and `fileAnalyses` reports mode/freshness/timeout/message metadata for every analyzed path; location filters and intentions remain single-file only.
+
+### Changed
+
+- **Bounded cache retention and maintenance** — symbol lookup avoids linear scans and expiry is amortized; symbol/hierarchy caches clean idle state on a schedule and drop closed projects promptly. Hierarchy history shares persistent storage, keeps at most ten snapshots per traversal, reuses retained successor cursors on retries, and enforces global pointer/text budgets with eviction counters. Existing one-hour symbol and ten-minute cursor idle TTLs remain unchanged.
+- **Fresh MCP schemas publish the new contracts** — semantic/refactoring schemas expose `symbolId` and structured `target`, and all three previewable refactorings expose `dryRun`. The protocol contract covers a fresh `initialize` → `tools/list`. The server continues to advertise `tools.listChanged: false`; after a plugin update, reconnect/restart MCP clients such as Codex because their `ALL_TOOLS` registry may be cached for the connection.
+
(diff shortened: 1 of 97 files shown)


hechtcarmel:
Hi @RomanPozdeev, thanks for putting this together. The direction is one this project wants: stable symbol handles, dry-run previews, and bounded hierarchy pages all address real pain points for agents.

**This comment was written by a Claude Code automation run by the maintainer.** It checked out the PR head (`fe63b2d`), merged current `main` locally, ran the static gates in `scripts/check-pr.sh`, and traced the diff against the platform sources. It could not execute the Gradle test suite in its sandbox, so the "full suite passes" checkbox is unverified below. If any finding here is wrong, say so and it will be re-checked.

**Bottom line: not mergeable as one PR.** It is 97 files and about 14.5k lines in a single commit, it changes behavior for existing clients in ways the description says it does not, it has several confirmed bugs, and no CI run exists for it yet. Splitting it into a series would let each piece be reviewed and shipped on its own.

Paths below are under `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/`; line numbers are at `fe63b2d`.

### Compatibility breaks (the PR says legacy fields remain supported)

- **`ide_call_hierarchy` no longer returns a tree.** Every node is emitted with `children = null`, and `CallElement` has no parent or depth field, so a depth-3 result is a flat mix of levels with no way to tell who calls whom. Fresh calls are also capped at 20 nodes *total*, where the old handler returned up to 20 per level per node. `tools/navigation/CallHierarchyTool.kt:479-500` and `:101-108`.
- **`ide_type_hierarchy`** loses the nested supertype chain (`TypeElement.supertypes` is now always null) and gains a default cap of 100 nodes. `tools/navigation/TypeHierarchyTool.kt:540-561`.
- **Pointer-bearing pagination cursors** (`ide_find_references`, `ide_find_class`, `ide_find_symbol`, `ide_find_implementations`) now fail with `SEARCH_INVALIDATED` on page 2 after any PSI edit, replacing the documented `stale: true` behavior. `tools/AbstractMcpTool.kt:878-931`, `server/PaginationService.kt:553`.
- **Kotlin abstract/sealed classes** now report `CLASS` instead of `ABSTRACT_CLASS`, because the reflective kind check runs before the abstract-modifier check. `handlers/java/JavaHandlers.kt:136`.
- **Smaller ones:** `ide_diagnostics` now rejects `line`/`column` without `file` (`tools/intelligence/GetDiagnosticsTool.kt:174-177`); a misspelled path in `ide_sync_files` walks up to the project root and triggers a full recursive refresh instead of returning "Not found" (`tools/project/SyncFilesTool.kt:149-168`); six tools dropped their `required` arrays.

### Confirmed bugs

- **Handle deleted on cancellation.** `SymbolIdRegistry.resolve` wraps `pointer.element` in `runCatching`, which swallows `ProcessCanceledException`, then removes the entry. A pending write action during a lookup permanently expires a valid `symbolId`. `server/SymbolIdRegistry.kt:183-190`.
- **Parameter safe-delete regressed.** The new path uses `SafeDeleteProcessor.createInstance(..., searchNonJava = true)`, so a parameter named `name` is blocked by the word appearing in a README or `.properties` file, and lambda / `catch` / for-each parameters hit `JavaSafeDeleteProcessor`'s `PsiMethod` assertion. `tools/refactoring/SafeDeleteTool.kt:1328-1340`.
- **Preview disagrees with apply.** Rename preview reports `canApply: false` whenever an `AutomaticRenamer` has candidates (for example `UserServiceTest` for class `UserService`), while the non-dry-run path accepts exactly those renames headlessly. Forced deletion of a file with no declarations has the same mismatch. `tools/refactoring/RenameSymbolTool.kt:634-640` vs `HeadlessRenameProcessor.kt:150-157`; `SafeDeleteTool.kt:579-580` vs `:684-699`.
- **One global lock.** `McpServerEpoch.ifCurrent` holds a single static monitor while `SymbolIdRegistry.bind` dereferences PSI inside it, so every bind (up to 500 per `ide_find_class` page), page lookup, and hierarchy registration across all requests serializes on it, and the EDT waits behind it on server stop and project close. `server/McpServerEpoch.kt:36-38`, `server/SymbolIdRegistry.kt:121-127`.
- **Lost hierarchy pages.** A page that crosses the 8,192-pointer continuation budget returns an error and discards its results, so a wide traversal can never complete. `tools/navigation/CallHierarchyTool.kt:379-381`, `server/HierarchyContinuationRegistry.kt:185-191`.
- **Swallowed cancellation in handlers.** Several rewritten reflection paths (Python `getSubtypes`, PHP, Go, Rust, JS) still catch `Exception` and return empty, and the tool now treats empty as "level complete", so a dumb-mode flip mid-search produces a definitive empty answer.

### Other things to resolve

- **CI never ran.** No workflow run exists for this PR, so nothing on GitHub has verified the test claim. A run will need to be approved once the PR is in reviewable shape.
- **The Ktor `IllegalAccessError` changelog entry.** Ktor is bundled plugin-first here, and Kotlin compiles `HttpMethod.Get` to the public `Companion.getGet()`, never to the private backing field; the same file still uses `HttpStatusCode.Forbidden` the same way. The string comparison is harmless, but please attach the stack trace and IDE build number, or drop the entry.
- **Unrelated change.** The hostname/IPv6 validator rewrite in `settings/McpSettingsConfigurable.kt` is not part of this feature, has no test, is not in the changelog, and will show a validation error to anyone whose saved host contains an underscore. Separate PR, please.
- **Schema growth.** `tools/list` grows from about 109 KB to 142 KB. Every client pays that in context on every session, so the new descriptions are worth trimming.
- **Dead and duplicated code.** The old recursive traversal survives in every handler; `MAX_VISITED_POINTERS_PER_CONTINUATION` and `UnifiedTargetArguments.optionalNonBlankString` are unused; the three refactoring tools each carry their own copy of the preview assembly.
- **Untestable surface.** About 1,200 lines of the diff are in the Python/Go/PHP/Rust/JS reflection handlers, which this repo's suite cannot exercise. Please state which IDEs you verified them in by hand.
- `CHANGELOG.md` conflicts with `main` (trivial).

### Suggested split

1. `SymbolIdRegistry` plus `symbolId` on `ide_symbol_info` and `ide_find_definition` only.
2. The nested `target` selector.
3. One `dryRun` PR per refactoring, with preview and apply agreeing on `canApply`.
4. Hierarchy pagination that keeps `children` (or adds a `depth`/`parent` field) and preserves the current first-page defaults.
5. Multi-file diagnostics.
6. `ide_sync_files` changes.
7. `ide_file_structure` nodes.

Each with its own changelog entry under `[Unreleased]`.

Credit where it is due: the new behavior tests are mostly real. `RefactoringDryRunBehaviorTest` snapshots disk bytes and hooks the apply phase, `SymbolIdBehaviorTest` asserts both directions on rename, and the hierarchy tests would fail on inverted logic. Two of them currently pin the flattened hierarchy and the `canApply: false` mismatch as intended behavior, so they would change along with the fixes.

---

PR #374, opened 2026-09-10
Validate and normalize MCP bind hosts before applying settings

RomanPozdeev:
## Summary

Validate bind-host syntax before DNS lookup so wildcard DNS cannot make malformed labels or `host:port` input valid. Normalize IDN hostnames before persistence, binding, and restart while preserving unchanged legacy hostnames and temporarily offline scoped IPv6 settings.

Keep DNS-rebinding protection enabled for every configured bind host that resolves to loopback. The Host allow-list accepts the standard loopback aliases plus the configured bind host's normalized spelling; incoming `Host` values are never DNS-resolved. Applying an unchanged legacy value also clears stale validation UI state.

## Validation

Final HEAD: `1a40facf88e97fef7b0e618537536da188549035`, one commit on current `upstream/main` `d01ebf520f4b6f78937a422f812458ba49ef2d7a`.

- `KtorMcpServerHttpTest`: 16 tests, 0 failures, including a configured non-standard loopback bind/Host regression.
- Platform: local IntelliJ IDEA 2025.3 / JBR 21 fallback; temporary build wiring was removed and was not committed.

## Context


hechtcarmel:
Thanks for splitting this out of #372 as requested. Review of PR #374 at HEAD `831248d`.

**This comment was written by Claude Code and drafted together with the repository owner (@hechtcarmel).** It checked out the PR head, diffed it against `main` (`f06fb23`), read the surrounding settings and server code, and ran the static checks from `scripts/check-pr.sh` (all pass). It could not run the Gradle suite in its sandbox and no CI run exists for this PR yet, so the test counts in the description are unverified here. Line numbers below are at `831248d` in `src/main/kotlin/.../settings/McpSettingsConfigurable.kt` unless stated otherwise.

## Bottom line

Good change: small, contained, and the normalize-then-persist half is a genuine fix (JVM sockets do not IDN-convert, so a Unicode host could never bind reliably before). Injecting the restart callback is a welcome test seam. **Merge after fixes 1 and 2 below, preferably 3 as well.**

## What it does

`apply()` now runs the host text through `normalizeBindHost` (:659-691): IPv4-looking strings keep the strict 4-octet rule, strings containing `:` are accepted iff `InetAddress.getByName` accepts them and are persisted verbatim, everything else goes through `IDN.toASCII(..., USE_STD3_ASCII_RULES)` plus a 63-char LDH label check and 253-char total, with a trailing dot preserved. The normalized string is what gets persisted, bind-checked and passed to the restart; `isModified()` (:318) compares normalized text against the saved value. `normalizeBindHostForSettings` (:649-656) skips normalization when the text equals the saved value **and** contains `:`, so an unchanged scoped IPv6 literal whose interface is down is not re-parsed.

## Findings

**1. Required: the underscore case from the #372 review is still there, and now blocks Apply.** `normalizeBindHostForSettings` only short-circuits an unchanged value that contains `:`. An unchanged hostname is re-run through `IDN.toASCII` with STD3 rules (:676), which rejects `_`. So for a persisted `my_host` (accepted on `main` whenever the resolver knew it, e.g. `/etc/hosts`, Docker/WSL DNS), `isModified()` is already true when Settings opens (`null != "my_host"`), and `apply()` throws "Invalid host" at :355 for any unrelated change, e.g. toggling a tool. The same applies to labels longer than 63 chars and leading/trailing hyphens. Suggested fix: never reject an unchanged persisted value, e.g. `if (trimmedHost == savedHost) return normalizeBindHost(trimmedHost) ?: savedHost`, and consider allowing `_` in `HOSTNAME_LABEL_PATTERN` (underscored names are common in practice). Please add a test: persist `my_host`, change `maxHistorySize`, `apply()` succeeds, host unchanged, no restart. It fails today.

**2. Required: a blocking resolver call on the EDT from `isModified()`.** The platform re-evaluates `isModified()` after every key/mouse event in the Settings editor. For text containing `:` that differs from the saved value, `normalizeBindHost` calls `InetAddress.getByName` (:672-674). When the first character is a hex digit, `:` or `[`, the JDK parses locally and returns immediately. When it is anything else (`localhost:29170`, the exact text of the "server running on" notification, or `host:8080`) the JDK skips literal parsing and calls the system resolver, so each keystroke blocks the UI for a DNS round trip. The old `isModified()` was a pure string compare, and this file's own focus-loss design (:498-507, pooled thread) exists to keep resolution off the EDT.

The same branch also undercuts the PR's headline claim: on a network with a wildcard search suffix, `host:8080` resolves, is persisted as the bind host, and `isValidHost` (:640) then short-circuits on `contains(':')`; the next start fails. Fix for both: syntax-precheck the `:` branch (optional `[...]`, `[0-9A-Fa-f:.]+`, optional `%zone`) before calling `InetAddress.getByName`, and add a `localhost:29170` rejection test through the injectable resolver asserting the resolver is never called.

**3. Recommended: two newly first-class forms bypass the DNS-rebinding guard.** `localhost.` (trailing dot deliberately preserved and tested) and `::ffff:127.0.0.1` (the JVM binds it as `127.0.0.1`) are loopback binds, but `LocalOriginGuard` only enforces the `Host` allow-list when the bind host is literally one of `127.0.0.1`, `localhost`, `::1` (`LocalOriginGuard.kt:39`, `:111`). Either canonicalize (`localhost.` to `localhost`, `::ffff:a.b.c.d` to `a.b.c.d`) or have the guard resolve the bind host and use `isLoopbackAddress()`. Not a regression, but this PR is what makes these forms first-class.

**4. Nits.** The "unchanged scoped IPv6 stays usable when its interface is offline" claim holds only while the field is never focused: tabbing through it runs the real validation on focus loss (:498-507) and `apply()` then fails, same as `main`. `java.net.IDN` is IDNA2003, so `straße.example` becomes `strasse.example` rather than the punycode modern resolvers use; harmless for a bind address. The CHANGELOG overstates the two claims above and should list the tightened rules (`_`, long labels, hyphens) under Changed rather than Fixed.

## Tests

`BindHostValidationUnitTest` drives the real validator through the injectable resolver with no live DNS; good. The platform apply tests assert persisted state and the injected restart. `BindHostOfflineInterfaceBehaviorTest` picks a nonexistent interface name deterministically. Gaps: no `:` input starting with a non-hex character (the exact hole in finding 2), and no unchanged-underscore test (finding 1). Note the IPv6 literal tests pass on `main` too, so they are characterization rather than regression tests.

## Cross-PR note

---

RomanPozdeev:
Thanks for the detailed review. Updated in `4218f78`, rebased onto current `main` (`357e47e`).

Addressed the findings:

- unchanged persisted legacy hosts are grandfathered, so unrelated Settings changes apply without changing the host or restarting the server; applying an equivalent spelling also clears stale pending validation state;
- colon/bracket input now passes a syntax-only IPv6 gate before any resolver call, so inputs such as `localhost:29170` are rejected without blocking the EDT on DNS;
- `localhost.` and IPv4-mapped loopback binds are covered by Host-header protection, with consistent trailing-dot normalization for Origin and Host values;
- added regression coverage for legacy values, malformed colon inputs, IPv6 forms, loopback aliases, equivalent Apply, and restart behavior.

hechtcarmel:
Re-review of the update at `4218f78` (rebased onto `357e47e`).

**This comment was written by Claude Code and drafted together with the repository owner (@hechtcarmel).** It re-checked out the head, diffed it against the previously reviewed `831248d`, traced the settings and origin-guard code, and ran the static part of `scripts/check-pr.sh` (passes). The Gradle suite still cannot run in its sandbox and no CI run exists yet, so the "full suite passes" claim remains unverified here. Line numbers are at `4218f78`.

## Verified fixed

- **Legacy hosts grandfathered.** `normalizeBindHostForSettings` falls back to the saved value when normalization fails (`McpSettingsConfigurable.kt:665`), `apply()` gates both validation paths and the restart on `hostChanged` (`:361`, `:374`, `:385`, `:421`), and `isModified()` is false on open for a persisted `my_host`. Editing to another invalid value is still rejected; edit-and-revert applies. `testLegacyPersistedHostsDoNotBlockUnrelatedSettings` covers exactly the requested case.
- **No resolver from `isModified()`.** Colon and bracket input goes through the pure `isValidIpv6Literal` (`:681-683`); `InetAddress` is reached only from `isValidHost` on focus loss (pooled thread) and from `apply()` for a changed host, which is pre-existing behavior. `localhost:29170`, `host:8080` and `1.2.3.4:80` normalize to null with the resolver never called, and the unit test asserts that. The IPv6 literal validator checks out against all the forms tried (`::1`, `::`, `[::1]`, scoped, IPv4-mapped, over-long, double `::`, `:::1`, non-hex group).
- CHANGELOG is one bullet under `[Unreleased]`; the `## [5.9.6]` section in the diff is main's rebase, not this PR.

## Remaining before merge

**1. Required: the origin guard now locks out every loopback bind that is not one of four literal spellings.** `installMcpOriginGuard` turns enforcement on for any bind host that resolves to loopback (`LocalOriginGuard.kt:118`, `isLoopbackBindHost`), but the request's `Host` is still compared against the four literals (`:148`, `hostnameOf(hostHeader) !in LOOPBACK_HOSTS`). So a server bound to `127.0.0.2`, `127.000.000.001`, `0:0:0:0:0:0:0:1`, `::ffff:7f00:1`, or a hostname that resolves to loopback such as `localhost.localdomain` rejects every request with "Host not allowed", because the client necessarily sends that same spelling. On `main` those binds skipped enforcement and worked. Fix: whenever enforcement is on, also allow the bind host's own normalized spelling, e.g. `LOOPBACK_HOSTS + normalizeHost(bindHost.trim().lowercase())`; do not resolve the `Host` header itself, since that is the rebinding vector. Please add a regression test in `KtorMcpServerHttpTest` that binds to `127.0.0.2` and sends `Host: 127.0.0.2:<port>` expecting 200 (it binds fine on `ubuntu-latest`). Related: `testLoopbackAliasesEnableHostGuard("::ffff:127.0.0.1")` cannot fail if the new set entry is deleted, because the resolver fallback still returns loopback; the `Host`-header side of that alias is untested.

**2. CHANGELOG.** Once item 1 is in, make the last sentence about Host-header protection describe the final behavior. The bullet is also long enough to split.

**3. Optional.** After a successful Apply of an unchanged legacy host, the flags are cleared (`:363-367`) but the red border and "invalid host" label painted on focus loss (`:531-538`) are not; cosmetic.

Nothing else regressed that I could find: restart fires exactly when the host or port changed, `[::1]` is persisted bracketed as before, and `isLoopbackBindHost` resolving at server start happens on background threads for a name that is about to be bound anyway.

---

RomanPozdeev:
Thanks — addressed the re-review in `1a40fac`, rebased onto current `main` (`d01ebf5`) after #375.

- Loopback Host enforcement now allows the normalized configured bind-host spelling in addition to the standard aliases, while the incoming `Host` header is still never resolved.
- Added end-to-end coverage that the default bind rejects `Host: 127.0.0.2` and a configured non-standard loopback bind accepts its own Host spelling. The acceptance test uses `127.0.0.2` where the platform can bind it (including `ubuntu-latest`) and the equivalent `127.000.000.001` spelling on macOS.
- Split the CHANGELOG wording into the bind-validation change and the final Host-header protection behavior.
- Also applied the optional UI cleanup: a successful Apply of an unchanged legacy host clears the stale label, icons, and `ComponentValidator` border.

Local validation on the final committed HEAD: `gw ./scripts/check-pr.sh` reported `Passed: 16  Failed: 0` and reran the complete suite successfully. The HTTP transport class passed 16/16. Temporary local-SDK wiring was removed before push.

PR #375, opened 2026-09-10
Add exact symbol handles to definition and metadata lookups

RomanPozdeev:
## Summary

`ide_find_definition` and `ide_symbol_info` return and accept opaque `symbolId` handles, allowing clients to reuse the exact declaration after line shifts or renames. Handles identify their owning open project and are bounded by the project instance, server session, cache capacity, and idle lifetime. Expired handles return rediscovery guidance without selecting a nearby declaration.

Lookups preserve the stored PSI target, including distinct JVM accessors and declarations outside the usual named-element interface. Deleting and recreating a file at the same path expires its old handles. Full previews of generated declarations without their own text use source context instead of failing.

Validation: published HEAD `a08710839684` is tree-identical to validated HEAD `f38f1b70458c`. The branch-specific `check-pr.sh` gate passed with 1,220 tests; 0 failures, 0 errors, and 0 skipped. Platform: IU-253.33813.55 / JBR 21. Local build overrides were not committed.

## Stack

Independent change based on the `upstream/main` 5.9.6 snapshot (`f06fb23`).

## Context

This focused pull request was split from upstream PR [#372](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/372) after review. The branch contains one commit relative to its stated base.


hechtcarmel:
Thanks for splitting this out of #372 as requested. Review of PR #375 at HEAD `a087108`.

**This comment was written by Claude Code and drafted together with the repository owner (@hechtcarmel).** It checked out the PR head, diffed it against `main` (`f06fb23`), read every production file in the diff plus the tests, and ran the static checks from `scripts/check-pr.sh` (all pass). It could not run the Gradle suite in its sandbox and no CI run exists for this PR yet, so the test counts in the description are unverified here. Paths below are under `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/`; line numbers are at `a087108`.

## Bottom line

This is the most valuable of the four split PRs and the design is sound: smart PSI pointers plus a VFS identity, bounded by a 4,096-entry LRU, a one-hour idle TTL, project-close cleanup and a server epoch that resets on stop; ids are unguessable; forged or expired handles fail cleanly with `SYMBOL_ID_EXPIRED` and no nearest-element fallback. **Both bugs from the #372 review are fixed, with tests that would catch regressions. Merge after the small fixes below.**

## Confirmed fixed from #372

- **Handle deleted on cancellation:** `resolve` dereferences the pointer at `server/SymbolIdRegistry.kt:212` with no `runCatching`; the entry is removed only on null / invalid / identity mismatch, and only if entry and pointer are still the same objects. `ProcessCanceledException` propagates and the enclosing read action retries against the intact entry. Pinned by the three `assertPointerFailureKeepsHandle` tests.
- **One global lock around PSI:** in `bind` the pointer is dereferenced (:145-146) before `serverEpoch.ifCurrent` (:148); critical sections are now map operations only. Lock order is consistent everywhere (service monitor, then epoch, then registry) and the sweep/project-close paths take only the registry monitor. Pinned by `testSessionResetDoesNotWaitForPointerRestorationDuringBind`.

Also checked and sound: cross-project isolation (`owner !== project` identity check, weak project refs), thread safety (single `LinkedHashMap` mutated only under the registry monitor), `CacheMaintenance` disposal, both `bind` call sites inside `suspendingReadAction`, negligible hot-path cost for other tools, golden diffs scoped exactly to the two tools (manifest +594 bytes).

## Findings

**1. Should-fix: `"symbolId": null` is now a hard error for every tool.** `server/mcp/McpToolDispatcher.kt:134-139`: `JsonNull` is a `JsonPrimitive` with `isString == false`, so the call returns "Parameter 'symbolId' must be a non-blank string." before the tool runs, for any tool. A few lines above, `project_path: null` is tolerated explicitly, and `optionalStringArg` normalizes `JsonNull` to absent, so null-for-unset clients are a known case in this codebase. Fix: treat `JsonNull` as absent.

**2. Should-fix (scope): tools beyond the two requested inherit `symbolId`, and their error text advertises it.** `resolveLookupMode` (`tools/AbstractMcpTool.kt:572-574`) and `resolveElementFromArguments` (:705-706) are shared, so every tool that resolves its target through them (at least `ide_find_references` and `ide_type_hierarchy`) accepts a handle end to end, dispatcher routing included, although their schemas do not declare it. Their missing-target error is now the globally edited `SYMBOL_OR_POSITION_REQUIRED`, "Must specify symbolId, file+line+column, or language+symbol" (`constants/ErrorMessages.kt:26-27`). The requested split was the two tools only. Either gate the `SYMBOL_ID` mode per tool and keep the old message for the rest, or declare, document and test the extra tools deliberately.

**3. Should-fix: the post-execution epoch check applies to all tools and is effectively unreachable.** `McpToolDispatcher.kt:159-163` turns any completed result into `SESSION_CHANGED_MESSAGE` and records the call as ERROR in history if the epoch advanced during the call. The epoch only advances after the Ktor server's stop has drained in-flight calls (`server/McpServerService.kt:236-242`), so a client can essentially never see it; when it does fire, it reports a completed mutation such as a rename as failed. It also buys nothing for the registry: `bind` already refuses a stale generation (:148-151), and a handle bound just before a reset simply expires next time. Recommend deleting it.

**4. Should-fix: documentation placement.** The per-tool sections still say "Target (mutually exclusive): file + line + column OR language + symbol", the parameter tables have no `symbolId` row and the Returns lines do not mention it (USAGE.md and tools-reference.md), and the `CLAUDE.md` inventory lines for the two tools (:498-499) are unchanged. Instead an identical paragraph was appended after the footer of README.md, USAGE.md, CLAUDE.md, SKILL.md and tools-reference.md. Please move the substance into the per-tool sections. This is also the only place where the four split PRs conflict with each other.

**5. Nits.** Leftover #372 scaffolding that nothing uses: `invalidate()` (`SymbolIdRegistry.kt:265-274`), `bindSymbolId` (`AbstractMcpTool.kt:740-750`), `responseHandleBudget` (:78-79), `isCurrentGeneration` (:132-133), `CacheEvictionReason.WEIGHT` / `REPLAY_LIMIT` and the `stored*` fields on `CacheStats` (`server/CacheMaintenance.kt:37-53`), the nullable `sweep` parameter, and the pagination-cursor carve-out in the dispatcher. `bind` failures (`require` at :145-146, `getOrThrow` at :107) surface through the dispatcher's `LOG.error` path rather than as clean tool errors; rare, but a `Result`-returning path would be cleaner. Bound the echoed id in `symbolIdExpired` (e.g. `take(64)`) since it is reflected into responses and command history. The registry class doc should say plainly that smart pointers hold the `Project` and, for physical elements, the `VirtualFile` strongly, and that the `projectClosed` listener, TTL and LRU cap are what bound retention (CONTRIBUTING's "no PSI in long-lived services" rule).

**6. Untested.** The `SYMBOL_ID_AND_OTHER_TARGET_EXCLUSIVE` branch, `JsonNull` handling, routing with two projects open, and the post-execution check. One unconfirmed risk worth a test: handles to light elements (the implicit enum methods your test covers) are hard pointers, and `LightElement.isValid()` does not track the declaring class, so after the enum is edited out of a file that still exists the handle may keep resolving. Bind `Choice.values()`, remove the enum, `ide_sync_files`, resolve, expect expired.

## Tests

Real and well targeted: the registry, exact-target, replacement and end-to-end lookup tests would fail if the corresponding logic were deleted or inverted, and the rename test asserts the same id comes back. `SymbolHandleReplacementBehaviorTest` relies on VFS record ids not being reused within a session; please make sure it runs on CI on this platform build.

## Cross-PR note

---

RomanPozdeev:
Thanks for the detailed review. Updated in `b1664de`, rebased onto current `main` (`357e47e`).

Addressed the findings:

- `symbolId: null` is treated as absent, and dispatcher routing considers a handle only for tools whose schema declares `symbolId`;
- handle lookup is explicitly opt-in at the shared resolver boundary, so tools outside `ide_find_definition` and `ide_symbol_info` neither inherit undocumented handle support nor advertise it in errors;
- removed the post-execution epoch check that could misreport an already completed mutation, while retaining the bind/reset generation guards;
- moved the contract details into the per-tool documentation and removed unused cache/handle scaffolding; echoed invalid IDs are bounded;
- added coverage for null/exclusive selectors, cross-project routing, and stale synthetic hard-pointer PSI. A conservative backing-file stamp expires self-navigating synthetic elements after source modification, while real coroutine cancellation/control-flow failures still propagate.

PR #376, opened 2026-09-10
Add bounded multi-file requests to ide_diagnostics

RomanPozdeev:
## Summary

`ide_diagnostics` accepts a `files` batch of up to 100 relative or in-project absolute paths under one shared analysis deadline. Results combine one bounded problem list with fail-closed per-file coverage using the project-diagnostics vocabulary (`analyzed`, `timed_out`, `failed`, `skipped`, `not_analyzed`) plus `not_found`, and explicit aggregate/per-file truncation metadata.

Path resolution is part of the shared budget and each resolved file is analyzed once. Isolated analyzer failures remain per-file, while real request-coroutine cancellation propagates. Ineligible targets are `skipped`; files not started before the deadline are `not_analyzed`. The shared single-file service now bounds refresh, PSI setup, lock waits, and analysis, and a slow open-editor daemon cannot receive a second batch-fallback budget.

Existing single-file and build/test-only requests remain compatible. A path equal to a project/content root is still not treated as a file target.

## Validation

Final HEAD: `1d233c0da8f2923e4321fe91c9881f1729643126`, one commit on current `upstream/main` `d01ebf520f4b6f78937a422f812458ba49ef2d7a`.

- Tool-manifest and result-shape contract tests passed.
- `GetDiagnosticsToolBehaviorTest`: 38 tests, 0 failures.
- Platform: local IntelliJ IDEA 2025.3 / JBR 21 fallback with deterministic hosts; temporary build wiring was removed and was not committed.

## Context


hechtcarmel:
Thanks for splitting this out of #372 as requested. Review of PR #376 at HEAD `ce28be0`.

**This comment was written by Claude Code and drafted together with the repository owner (@hechtcarmel).** It checked out the PR head, diffed it against `main` (`f06fb23`), read the diff plus the unchanged parts of the diagnostics engine and `ProjectDiagnosticsTool`, and ran the static checks from `scripts/check-pr.sh` (all pass). It could not run the Gradle suite in its sandbox and no CI run exists for this PR yet, so the test counts in the description are unverified here. Paths below are under `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/intelligence/`; line numbers are at `ce28be0`.

## Bottom line

The feature works and stays inside the repo's long-poll rule, and the #372 compatibility break (rejecting `line`/`column` without `file`) is gone. The problem is overlap: `ide_project_diagnostics` already accepts explicit file paths and has a fail-closed per-file coverage vocabulary, and this PR introduces a second, incompatible vocabulary over the same engine. Result models are wire contract here, so **this needs a decision before merge, then a handful of small fixes.**

## What it does

`ide_diagnostics` gains a `files` array (1 to 100 project-relative paths, mutually exclusive with `file`; location filters are single-file only). The batch path (`GetDiagnosticsTool.kt:249-340`) captures one 30-second deadline (`DiagnosticsAnalysisService.kt:47`, captured at :253), analyzes files sequentially with the remaining budget, aggregates problems into one list capped at 100, and emits a `FileDiagnosticsAnalysis` entry per file; files reached after the deadline are marked `timedOut` without work. Partial results come back in-band, never as an error. New wire fields are additive: `problemsTruncated` and `fileAnalyses` on `DiagnosticsResult`, plus the new `FileDiagnosticsAnalysis` model.

The good part that the description undersells: `analyzeFile` is now wrapped as a whole in `withTimeoutOrNull` (`DiagnosticsAnalysisService.kt:115`), so the single-file call is bounded at 30 s including disk refresh and the main-pass lock wait. On `main` the lock wait was unbounded and a running project analysis could push a single-file call past the 55 s limit. That is a real fix and it belongs in the CHANGELOG.

## Overlap with `ide_project_diagnostics`

| | `ide_diagnostics` + `files` (this PR) | `ide_project_diagnostics` (existing) |
|---|---|---|
| Enabled by default | yes | no |
| Scope input | explicit list, at most 100 files | `paths`: files and/or directories, or whole project |
| Unresolvable path | per-entry "File not found", call succeeds | whole call errors |
| Time bound | fixed 30 s, synchronous | `timeoutSeconds` up to 3600, long-poll via `analysisId` |
| Per-file coverage | `mode`, `fresh`, `timedOut`, free-text `message` | `state` enum (`analyzed` / `timed_out` / `failed` / `skipped` / `not_analyzed`) + `reason`, per-state counts, `complete` |
| Cancellation mid-run | whole call fails, results lost | per-file `failed`, loop continues |
| Absolute paths | rejected for `files`, accepted for `file` | accepted |
| Build/test merge, intentions | yes / single-file only | no / no |

There is a genuine gap: the default-enabled tool cannot answer "check the eight files I just edited" in one call, and the project tool is opt-in, exclusive, and fails the whole call on one bad path. So this is not pure duplication. But the second vocabulary is a permanent cost. Recommendation if the feature is kept: make `fileAnalyses` entries carry the same `state` / `reason` enum as `FileCoverageInfo` (adding `not_found`) before merging; renaming later is a breaking change for clients. If two batch surfaces are not wanted, the alternative is closing this in favor of `ide_project_diagnostics` with `paths`.

## Findings

**1. Should-fix: one cancellation loses the whole batch.** A `ProcessCanceledException` from one file's analysis is rethrown (:296-301) and surfaces as a JSON-RPC error, discarding the files already analyzed. `ProjectDiagnosticsTool.kt:380-391` deliberately treats the same exception as per-file `failed`. The test at `GetDiagnosticsToolBehaviorTest.kt:725` pins the divergent choice. Treat it as a per-file failure unless the coroutine itself is cancelled.

**2. Should-fix: absolute paths rejected only here.** `normalizeProjectRelativePath` (:411-434) rejects absolute paths for `files`, while `file` and the project tool accept them through `resolveFile`, which already confines absolute paths to project roots. Agents send absolute paths habitually. Also, the function's return value is discarded at :143 while its KDoc and the tool description ("lexical aliases are analyzed only once") claim it is the dedup key; the actual key is the resolved VFS URL (:144-145), which is better (symlink aliases collapse, distinct files through `..` across a symlink stay distinct, as your test at :745 proves), so fix the wording rather than the code.

**3. Should-fix: the deadline starts late.** The parse phase calls `resolveFile` for every path (:127-154), which for files not yet in VFS is a synchronous refresh each, before the deadline is captured at :253. With 100 just-written files, the headline use case, that work is outside the budget. Capture the deadline first and resolve each path once (the loop resolves again at :265).

**4. Should-fix: the 100-problem aggregate cap defeats the batch.** With `severity: all`, a few files' ordinary warnings exhaust the cap (:283-293) and the rest come back as `problemCount: 0, problemsTruncated: true` with an instruction to re-query per file, which is the N-round-trip workflow the feature exists to remove. Add a `maxProblems` parameter (default 100, capped) like the project tool.

**5. Tests.** Remove the stray comment at `DiagnosticsWhitespacePathBehaviorTest.kt:16` ("Standalone review probe: ... do not edit the snapshot"). The symlink test (:745) calls `Files.createSymbolicLink` (:753) with no `Assume` guard and leaves `src/link`, `other/` and an added source root behind for later tests in the class. `testMultiFileDiagnosticsUseOneSharedTimeoutBudget` (:862) leaves 100 ms for refresh, EDT hops, read action and lock on a loaded runner; widen the margins. No batch test runs the real engine (all inject problems through the override); one with plain syntax errors needs no JDK and would close that.

**6. Nits.** Every closed file repeats the closed-batch boilerplate message (`DiagnosticsAnalysisService.kt:358-362`, up to roughly 12 KB per response); hoist it to one top-level note. Late files say "timed out after 37ms" and nothing says when lock contention with another diagnostics run was the cause. `File(basePath, path).canonicalPath` at :145 can throw `IOException`. `uniqueItems: true` in the schema versus tolerant dedup in the tool. The USAGE example response omits `analysisMode`, `problemsTruncated` and `fileAnalyses`; tools-reference says one entry per requested path, but aliases collapse to one entry under the first spelling.

## Cross-PR note

The four split PRs merge cleanly against each other in code and conflict only in `CHANGELOG.md`, `CLAUDE.md` and `SKILL.md`. Suggested order once fixed: #374, #377, #375, then #376 if kept. CI needs a workflow approval to run for this contributor.

---

RomanPozdeev:
Thanks for the detailed review. Updated in `ed212f9`, rebased onto current `main` (`357e47e`). The explicit small-file batch remains because it covers the default-enabled, bounded "check these edited files" workflow; its coverage vocabulary is now aligned with `ide_project_diagnostics`.

Addressed the findings:

- per-file results now use shared `state` / `reason` semantics, including explicit `not_found`;
- an analyzer `ProcessCanceledException` is isolated as a per-file failure only while the request coroutine remains active; real coroutine cancellation still propagates;
- relative and in-project absolute paths are supported consistently, aliases are described according to the actual resolved identity, and the shared deadline begins before path resolution with each target resolved once;
- added configurable `maxProblems` and explicit aggregate/per-file truncation metadata;
- tightened symlink/content-root cleanup and expanded coverage for deadlines, paths, failures, cancellation, and response limits.

hechtcarmel:
Re-review of the update at `ed212f9` (rebased onto `357e47e`).

**This comment was written by Claude Code and drafted together with the repository owner (@hechtcarmel).** It re-checked out the head, diffed it against the previously reviewed `ce28be0`, traced the batch path against `DiagnosticsAnalysisService` and `ProjectDiagnosticsTool`, and ran the static part of `scripts/check-pr.sh` (passes). The Gradle suite still cannot run in its sandbox and no CI run exists yet. Paths are under `src/main/kotlin/.../tools/`; line numbers are at `ed212f9`.

## Verified fixed

- **Shape aligned.** `FileDiagnosticsAnalysis` is now `{file, state, reason?, mode?, problemCount, problemsTruncated}`, reusing the project tool's state constants with `not_found` added there rather than in a parallel enum. Wire changes vs `main` are additive only.
- **Cancellation isolated.** `ProcessCanceledException` is caught first, `ensureActive()` rethrows only when the request coroutine itself is cancelled, other `CancellationException`s propagate, and `ensureActive()` also runs at each loop top (`intelligence/GetDiagnosticsTool.kt:253`, `:277`, `:321-345`).
- **Paths and deadline.** `files` entries go through `resolveFile`, so in-project absolute paths work; the deadline is captured at `:250` before resolution; each path is resolved once into `BatchFileTarget`; the dedup key is the resolved VFS URL and the docs now say so.
- **`maxProblems`** with lookahead on both paths, and the zero-slot probe so hidden errors are never presented as clean.
- **Tests.** Probe comment gone; symlink test uses `Assume.assumeNoException` and cleans up in `finally`; timing margins widened; a real-engine batch test exists. CHANGELOG has the single-file timing fix under Fixed.

## Remaining before merge

**1. Contract: ineligible files are labeled `not_analyzed`.** `GetDiagnosticsTool.kt:356-360` maps a result with `analysisFresh = false` and `analysisTimedOut = false` to `not_analyzed`. `ProjectDiagnosticsTool.coverageStateFor` (`:76-80`) maps the identical engine result to `skipped`, and that tool's contract defines `skipped` as "not eligible, with reason" and `not_analyzed` as "cut off by maxFiles/timeout". As written, `ide_diagnostics` never emits `skipped`, its docs list five states without it, and an agent seeing `not_analyzed` on a binary or PSI-less file will retry it. Fix: use `coverageStateFor` directly (it is `internal`), which also returns a null reason for `analyzed` files and so removes the closed-batch boilerplate repeated per file; add `skipped` to the state lists in USAGE.md, tools-reference.md and CLAUDE.md.

**2. Decide: `resolveFile` now accepts a path equal to a root.** `AbstractMcpTool.kt:425-426` (`isWithinRoot`) adds `canonicalPath == rootPath`; on `main` only `startsWith(root + separator)` matched. This is a shared helper, so it changes every tool: `ide_project_diagnostics` with `paths: ["."]` now walks the whole project instead of returning "Path not found", `ide_diagnostics` with `file: "."` returns "not eligible" instead of "File not found", and the other callers degrade to their null-PSI paths. Either revert to strict containment or document and test it; it is currently neither. Related: `File(path).isAbsolute` replaces the `startsWith("/")` check, so on Windows `/src/X.java` is now resolved relative to the base path rather than rejected.

## Non-blocking

- "Real cancellation still propagates" is asserted only in a comment. A test that cancels the request coroutine mid-file and asserts `execute` throws instead of returning a per-file `failed` entry would pin it.
- `resolveFile` no longer throws `IOException`, so the `resolutionFailure` state (`:198-199`, `:264-266`, `:283-290`) can never be emitted.
- With the outer `withTimeoutOrNull` at `DiagnosticsAnalysisService.kt:115`, the open-editor "daemon timed out, fall back to batch" branch (`:206-228`, `:240`, `:250`) can no longer execute: the outer timeout always fires first. The net effect is desirable (`main` could spend 30 s waiting plus 30 s in batch), but a slow-daemon open file now returns `analysisTimedOut` with no problems rather than batch results, and neither the CHANGELOG nor a test says so.
- The CHANGELOG Fixed entry names only `ide_diagnostics`, but the change lives in the shared service, so `ide_project_diagnostics` per-file timeouts now include refresh and lock wait too.
- The comment at `:335-337` duplicates the one at `:322-324`.

---

RomanPozdeev:
Thanks — addressed the re-review in `1d233c0`, rebased onto current `main` (`d01ebf5`) after #375.

- Multi-file results now use `ProjectDiagnosticsTool.coverageStateFor`, so ineligible targets are `skipped`, analyzed entries do not repeat the closed-batch boilerplate, and all state lists document `skipped` versus `not_analyzed` explicitly.
- Restored strict shared `resolveFile` containment: a root itself is not a file target. Leading `/` and `\\` absolute syntax is still routed through absolute-path validation on every platform.
- Removed the unreachable path-resolution failure model/handling and the duplicated cancellation comment; added a real request-coroutine cancellation regression and a PSI-less `skipped` regression.
- Removed the unreachable inner open-daemon timeout outcome. The outer complete-operation timeout owns the budget; only an early `DaemonDidNotRun` can fall back to batch, and a slow daemon cannot receive a second budget. Added coverage for both analyzer overrides and documented the shared-service behavior for both diagnostics tools.

Local validation on the final committed HEAD: `gw ./scripts/check-pr.sh` reported `Passed: 16  Failed: 0` and reran the complete suite successfully. Tool-manifest/result-shape contracts passed; `GetDiagnosticsToolBehaviorTest` passed 38/38. Temporary local-SDK wiring was removed before push.

PR #377, opened 2026-09-10
Make targeted file synchronization precise for new and deleted paths

RomanPozdeev:
## Summary

`ide_sync_files` validates every explicit target before touching VFS and returns one whole-batch error listing all invalid paths instead of partial success. Absolute targets may match any safe project/content root. Relative targets try the project base and module content roots when `project_path` is omitted or selects the base; selecting a specific content root confines relative resolution there.

Known deleted targets refresh through their nearest existing parent, new targets are discovered through shallow ancestor refreshes without repeating the same ancestor work, and registered symlink-root spellings are preserved. The response reports normalized `syncedPaths`, actual absolute `refreshedRoots`, and `deletedPaths`. Full synchronization fails explicitly when the project base cannot be resolved safely rather than silently choosing another root.

## Validation

Final HEAD: `fe4dcc7bca430a8179f27477559fb6c36039b50d`, one commit on current `upstream/main` `d01ebf520f4b6f78937a422f812458ba49ef2d7a`.

- Tool-manifest and result-shape contract tests passed.
- `SyncFilesToolBehaviorTest`: 15 tests, 0 failures; targeted new-path scope test passed.
- Platform: local IntelliJ IDEA 2025.3 / JBR 21 fallback with deterministic hosts; temporary build wiring was removed and was not committed.

## Context


hechtcarmel:
Thanks for splitting this out of #372 as requested. Review of PR #377 at HEAD `433eb9b`.

**This comment was written by Claude Code and drafted together with the repository owner (@hechtcarmel).** It checked out the PR head, diffed it against `main` (`f06fb23`), read the new `SyncFilesTool` in full next to `main`'s version and `AbstractMcpTool.resolveFile`, and ran the static checks from `scripts/check-pr.sh` (all pass). It could not run the Gradle suite in its sandbox and no CI run exists for this PR yet, so the test counts in the description are unverified here. Line numbers below are at `433eb9b` in `src/main/kotlin/.../tools/project/SyncFilesTool.kt` unless stated otherwise.

## Bottom line

The #372 concern is properly fixed, the containment and symlink logic is sound, and refreshing the symlink-alias tree the project actually registered (instead of the canonical tree `resolveFile` finds) is a real correctness win that the alias tests genuinely prove. **Merge after the fixes below; two of them are regressions for agents.**

## What it does, and what was confirmed

The tool grows its own validation pipeline instead of `resolveFile`. Every entry in `paths` is validated before any VFS call: blank, `..`, leading separator, drive letter and `Path.isAbsolute` are rejected (:201-225); the lexical target must stay under the root; the nearest existing on-disk ancestor must canonicalize under the root's real path; and a path missing on disk **and** unknown to `LocalFileSystem.findFileByPath` is rejected as "Not found" (:190-192), thrown from the validation `map` at :75 before any refresh. So a typo triggers no refresh at all, not even a shallow one. Known-deleted paths get one shallow refresh of the nearest existing parent (:100-102); new paths are discovered by shallow hops from the deepest cached ancestor (:89-99) and then refreshed recursively. Containment uses component-wise `Path.startsWith` throughout, so there is no `/proj/foo` vs `/proj/foobar` prefix bug. Threading is unchanged from `main` (synchronous refresh off the EDT under the dispatcher's modality). Wire: three additive result fields (`requestedPaths`, `refreshedRoots`, `deletedPaths`), nothing removed or renamed; the "Synchronized N of M, Not found: ..." success message can no longer occur.

## Findings

**1. Should-fix: absolute in-project paths are now rejected** (:208-214, :221-223). `main` accepted them through `AbstractMcpTool.resolveFile` ("Absolute paths are validated against project roots"), and every other tool still does. Claude Code and similar clients emit absolute paths habitually, so `{"paths": ["/home/u/proj/src/A.java"]}` goes from refreshed to `isError`. The containment machinery already exists: accept an absolute path whose real path lies under the selected root's real path, relativize it, and proceed.

**2. Should-fix: no cross-content-root fallback when `project_path` is omitted** (:151, :170). Paths now resolve against `basePath` only, unlike `resolveFile` and the workspace convention in CLAUDE.md (basePath, then all module content roots). In a workspace with a content root outside basePath, `src/Foo.java` becomes "Not found", or worse, a same-named file under basePath is refreshed instead. Either mirror the root loop when `project_path` is omitted, or document the divergence in CLAUDE.md's workspace section and the CHANGELOG.

**3. Should-fix: only the first invalid path is reported** (:74-78, `map` throws on the first). All-or-nothing validation is fine, since refresh is idempotent and a retry is cheap, but an agent with three typos in a twenty-path batch pays three round-trips. Collect every validation error into one response.

**4. Should-fix (tests): content roots added to the shared light module are never removed.** `PsiTestUtil.addContentRoot` at `SyncFilesToolBehaviorTest.kt:213`, `:275-276`, `:332`, `:367` and `SyncFilesRoutingBehaviorTest.kt:65`, and `removeContentEntry` appears nowhere in the test tree; two tests register the same `linked-module` URL and a `/tmp` canonical root outlives its directory. `distinctBy { it.path }` (:156) masks this for this tool, but every later test in the fork sees dead roots in `getModuleContentRoots`. Remove them in `finally`. The symlink tests (`Files.createSymbolicLink` at `:184`, `:208`, `:269`, `:327-328`) are the first in the repo and need an `Assume.assumeTrue` capability probe per CONTRIBUTING's visible-skip rule.

**5. Docs and CHANGELOG.** The new CLAUDE.md section is appended after the "Template Source" footer (`CLAUDE.md:716-721`) while the `ide_sync_files` inventory bullet at :508 is untouched; same placement in SKILL.md. Please move the substance into the bullet. The CHANGELOG lists two behavior changes (absolute paths rejected, partial success removed) under Fixed; they belong under Changed.

**6. Nits.** `refreshedRoots` mixes an absolute root with relative entries (:267-270), so a batch touching the root yields `["/abs/root", "a/b/New.java"]`. N new files under one new directory re-walk and re-refresh the same ancestors N times (:80-85 vs :89-99): correct, just 2N synchronous round-trips. Whole-project entries use `Path.toString()` (:123-124, :268), which gives backslashes on Windows where `main` returned `basePath` with forward slashes. `requestedPaths` always equals `syncedPaths` in targeted mode and is `[]` otherwise, so consider dropping it before it ships as a contract field. `catch (_: Exception)` at :182 and :284 erase the failure kind. The comment at `SyncFilesToolBehaviorTest.kt:66-73` still describes the EDT limitation this PR removed. Dangling symlink targets known to VFS cannot be synced (:180-186) although a parent refresh would work.

## Tests

Mostly good and honest: fixtures use `writeProjectFile` and real filesystem paths, no bare-return skips. `testMisspelledPathDoesNotRefreshTheWholeProject` fails on `main`. The deleted-path test fails if :102 is inverted. The alias tests are the best in the PR; on `main` the alias document stays stale. `SyncFilesNewTargetScopeBehaviorTest` passes on `main` too (its `refreshAndFindFileByPath` is also shallow) but does fail if :93 is flipped to recursive, so it guards the #372 regression. `testSyncResolvesPathsAgainstSelectedWorkspaceContentRoot` passes only because the light project's temp dir is already canonical and never asserts the deleted file's VirtualFile was evicted.

## Cross-PR note

---

RomanPozdeev:
Thanks for the detailed review. Updated in `6acfbb0`, rebased onto current `main` (`357e47e`).

Addressed the findings:

- absolute paths inside the selected project/content roots are accepted while traversal and symlink escapes remain rejected;
- when `project_path` is omitted, relative targets fall back across module content roots, and routing handles null or externally deleted roots without widening refresh scope;
- all invalid paths are collected and reported before any refresh begins;
- refresh roots are normalized and minimized while preserving the registered symlink path that VFS actually owns;
- tests now use visible symlink capability skips and remove temporary content roots in `finally`; the per-tool documentation was integrated into the existing sections.

hechtcarmel:
Re-review of the update at `6acfbb0` (rebased onto `357e47e`).

**This comment was written by Claude Code and drafted together with the repository owner (@hechtcarmel).** It re-checked out the head, diffed it against the previously reviewed `433eb9b`, traced the new resolution pipeline against `main`'s `SyncFilesTool` and `AbstractMcpTool.resolveFile`, and ran the static part of `scripts/check-pr.sh` (passes). The Gradle suite still cannot run in its sandbox and no CI run exists yet. Line numbers are at `6acfbb0` in `tools/project/SyncFilesTool.kt` unless stated otherwise.

## Verified fixed

- **Absolute in-project paths** are accepted by lexical or real-path containment per root (`:186-197`); only absolute syntax that is not absolute on the platform is rejected (`:269-273`). Tests cover inside base, outside all roots, and canonical-path-into-alias-root.
- **Cross-root fallback** with `project_path` omitted: `candidateRoots` is all allowed roots (`:68`), first root that resolves wins, base path first, matching `resolveFile`'s order.
- **Every invalid path reported**: parse, containment and VFS-parent failures are collected and returned together before the first VFS mutation at `:110`.
- **Tests**: every `addContentRoot` is undone in `finally`; all four symlink tests go through `createSymbolicLinkOrSkip` with `Assume.assumeNoException`, inside `try` so cleanup runs on skip.
- **Docs**: CLAUDE.md inventory bullet and SKILL.md rule updated in place, nothing after the footers. `requestedPaths` dropped; `refreshedRoots` are absolute, system-independent. Golden diff vs `main` is additive.

No correctness regression found in the new code paths; threading is unchanged.

## Remaining before merge

**1. CHANGELOG (`:7-9`).** Still under `Fixed`, and "instead of falling back to a full-project refresh" is not what `main` did: `main` skipped unresolved paths and reported them in a success message ("Synchronized N of M requested path(s). Not found: ..."), and refreshed VFS-known deleted files directly. Please move the entry under `Changed` and state the actual behavior changes: partial success replaced by a whole-batch error listing every invalid path; absolute in-project paths accepted; relative paths fall back across module content roots when `project_path` is omitted; new `refreshedRoots` and `deletedPaths` fields.

**2. Design gap: the fallback never applies when `project_path` is given.** With `project_path` set, `candidateRoots` is the selected root only (`:68`), for absolute paths too (`:188` iterates only `candidateRoots`). In any multi-project session the dispatcher requires `project_path`, and SKILL.md tells agents to send the project base path, so a path under a content root outside the base path (the `ide_import_modules` workspace case the fallback was added for) returns "Not found. Only files known to the IDE can be synchronized after deletion" (`:240`) or "Absolute path is outside the selected project/content roots" (`:202`), where `main` succeeded because `resolveFile` ignores `project_path`. Suggested fix: match absolute paths against all allowed roots (containment is unambiguous, so confinement buys nothing there), and fall back across roots when the selected root is the base path. If the confinement is intended, say so in CLAUDE.md's workspace section and the CHANGELOG.

**3. Small.** `allowedRoots.first()` (`:137`) throws `NoSuchElementException` when every root fails real-path resolution (base path is a dangling symlink, or a Windows drive is unmounted), and `AbstractMcpTool.execute` rethrows it as a raw exception; if only the base path is dropped, a full sync silently refreshes the first surviving content root while reporting "Synchronized entire project." One `isEmpty()` guard with an explicit error covers both.

**4. Small.** The comment at `SyncFilesToolBehaviorTest.kt:68-76` still says explicit paths go through `resolveFile` and cannot be exercised off the EDT; `resolveFile` is no longer referenced in the tool, and `SyncFilesNewTargetScopeBehaviorTest` does exactly that.

## Nits

- N new files under one new directory re-walk and re-refresh the same ancestors N times (`:88-97` vs `:104-116`).
- `catch (_: Exception)` at `:233`, `:329`, `:347` erase the failure kind.
- An absolute path containing `..` that normalizes inside the project is rejected at parse (`:260-262`); `main` and every other tool accept it via canonicalization.
- The "Not found" message implies the caller deleted the file, which is misleading for a plain typo.
- Fixture names such as `missing-review/`, `FreshReview.java` and `HiddenReview.java` read like review-cycle residue.

---

RomanPozdeev:
Thanks — addressed the re-review in `fe4dcc7`, rebased onto current `main` (`d01ebf5`) after #375.

- A base `project_path` now keeps cross-content-root fallback for relative paths; a specifically selected content root still confines relative resolution. Absolute paths are checked against every allowed root regardless of the selected root.
- Empty safe-root sets now return an explicit error. Full sync without `project_path` requires the safely resolved project base instead of silently choosing the first surviving content root.
- Absolute paths containing normalizable `..` are accepted when the normalized target remains in-project; relative traversal remains rejected.
- Narrowed broad catches, clarified missing-path errors, removed review-cycle fixture names/stale comments, and avoid repeating the same shallow ancestor refresh.
- Updated routing docs and moved the complete behavior summary under CHANGELOG `Changed`, including whole-batch errors and the `refreshedRoots`/`deletedPaths` contract.

Local validation on the final committed HEAD: `gw ./scripts/check-pr.sh` reported `Passed: 16  Failed: 0` and reran the complete suite successfully. Tool-manifest/result-shape contracts passed; `SyncFilesToolBehaviorTest` passed 15/15 and the new-target scope regression passed. Temporary local-SDK wiring was removed before push.

PR #379, opened 2026-09-11
feat: add excludeDirectories parameter to ide_open_project

### LINKED ISSUE #378: ide_open_project: add excludeDirectories parameter to exclude non-code directories from indexing
## Problem

When a project root contains non-code directories (workspace doc symlinks, `.claude` directories with markdown plans/specs), IntelliJ indexes them as project content. This causes `ide_refactor_rename` and `ide_move_file` to abort with "read-only files or unresolved conflicts" when the markdown files contain text matching the class name being renamed.

Common scenario: agentic workflows use a `wksp` symlink inside the project root pointing to a companion workspace with plans, specs, and blog posts. These are pure `.md` files — no code, nothing for IntelliJ to refactor — but IntelliJ indexes them and the rename processor's scope check finds text matches there.

## Suggested fix

Add an `excludeDirectories` parameter to `ide_open_project` that marks specified directories as excluded from IntelliJ's indexing and refactoring scope via `ContentEntry.addExcludeFolder()`.

```json
{
  "path": "/Users/dev/myproject",
  "autoLink": true,
  "excludeDirectories": ["wksp", ".claude"]
}
```

This uses the same IntelliJ API that `ide_create_module` already uses for its `excludeDirectories` parameter.

mdproctor:
## Summary

Closes #378.

When a project root contains non-code directories (workspace doc symlinks with markdown plans/specs, `.claude` directories), IntelliJ indexes them as project content. This causes `ide_refactor_rename` and `ide_move_file` to abort when the markdown files contain text matching the class name being refactored — the rename processor's scope check finds "matches" in `.md` files and treats them as conflicts.

Adds an `excludeDirectories` parameter to `ide_open_project` that marks directories as excluded from IntelliJ's indexing and refactoring scope after opening:

```json
{ "path": "/Users/dev/myproject", "excludeDirectories": ["wksp", ".claude"] }
```

### Implementation

After the project opens and reaches smart mode, iterates all modules' content entries and calls `contentEntry.addExcludeFolder()` for any matching child directories — the same IntelliJ API that `ide_create_module` already uses for its `excludeDirectories` parameter. The exclusion runs inside `WriteCommandAction` on the EDT.

Exclusions survive IDE restarts (IntelliJ persists them in module config). Already-excluded directories are skipped (idempotent).

### Use case

Agentic workflows use a `wksp` symlink inside the project root pointing to a companion workspace with plans, specs, and blog posts. These are pure `.md` files that IntelliJ shouldn't care about for refactoring — but without exclusion, they interfere with rename/move operations.

## Test plan

- [x] Compiles clean
- [x] Full test suite passes (1 pre-existing flaky test in `KtorMcpServerHttpTest` unrelated)
- [x] Golden tool manifest regenerated — diff shows only the new `excludeDirectories` property on `ide_open_project`


hechtcarmel:
> **Note:** this is a Claude Code message, drafted with the repo owner (@hechtcarmel).

Thanks for this, @mdproctor. The idea is a good one and we want to land it: the rename tool's own "blocked by read-only files" error already tells the agent to exclude the directory from the project's content roots, and today no MCP tool can do that on an existing project. The implementation also reuses the same `addExcludeFolder` path that `ide_create_module` uses, which is the right API. A few things need to change before it can merge. Blocking items first, then smaller ones.

### Blocking

1. **It silently does nothing when the project is already open.** `doExecute` returns "already open" at `OpenProjectTool.kt:102-104` before the exclusion code runs, because the exclusion lives inside the open-and-wait block. For the workflow in #378 that is the common case: the agent hits the rename error inside a project that is already open, and calling `ide_open_project` again with `excludeDirectories` won't help. Please apply the exclusions on the already-open path too, and report them (see 3). That also makes the feature testable end to end against the test fixture project (see 4).

2. **Ordering relative to `autoLink`.** The exclusion runs before `tryAutoLink` (`OpenProjectTool.kt:119-125`). A Maven/Gradle link imports the build model and rebuilds the module content entries, so exclusions attached to the pre-import module are at risk of being dropped. That is exactly the combination in the PR's own example (`"autoLink": true, "excludeDirectories": [...]`). Please move the exclusion after the link, and after `commit()` wait for smart mode again: committing a root-model change triggers a rescan, and the tool currently returns "open and ready" without waiting, unlike the `autoLink` branch does. Please also verify by hand that the exclusions survive a Gradle/Maven re-sync, since the description says they persist.

3. **No feedback, and errors are swallowed.** A name that doesn't exist under a content root is skipped without a word (`findChild(dirName) ?: continue`), the `catch (e: Exception)` around the root model discards everything including `ProcessCanceledException`, and the success message never changes. The caller cannot tell whether the exclusion happened. CONTRIBUTING.md's error-handling section asks that failures not be silently swallowed. Please have the result list what was excluded, what was already excluded, and what was not found, the way `ide_create_module` reports "Excluded N directories", and surface a failure as a warning or error.

4. **No tests.** `CreateModuleBehaviorTest.testHappyPath_createsModuleWithContentRootAndExcludes` shows the assertion shape: run the tool, then check `excludeFolderUrls` on the module's content entries. With (1) in place, `ProjectWindowToolsTest` can call the tool with `project.basePath` plus `excludeDirectories` and assert the exclusion (see `testOpenProjectIsIdempotentWhenProjectAlreadyOpen` for the setup). A schema assertion in `ProjectWindowToolsUnitTest` and a negative case for a missing directory would round it out.

5. **Merge conflict in `CHANGELOG.md`.** `main` gained `### Fixed` entries under `[Unreleased]` from #370 at the same spot. Rebase or merge `main` and keep both sections.

### Should fix

- **Docs.** `USAGE.md` (the `ide_open_project` section and its parameters table), `src/main/resources/skill/ide-index-mcp/references/tools-reference.md`, and the tool inventory line in `CLAUDE.md` still describe the old parameter set. `autoLink` is missing from the first two as well; adding it while you are there would be welcome but is not required.
- **Naming and validation vs. the sibling parameter.** The PR text says `ide_create_module` has an `excludeDirectories` parameter; it is actually `excludes`, it accepts relative paths, and it rejects blank entries and `..`. The new parameter is named differently, only matches direct children (`findChild` takes a single name, so `docs/generated` is silently ignored), has no validation, and `it.jsonPrimitive.content` throws on a non-string element. Preference: match the sibling (same name, relative paths, same validation) so agents learn one convention. If you would rather keep `excludeDirectories`, please at least add the validation and parse with `mapNotNull { (it as? JsonPrimitive)?.contentOrNull }` like `CreateModuleTool` does.
- **Schema.** Use `SchemaBuilder.stringArrayProperty(...)` instead of the hand-built `buildJsonObject`. CONTRIBUTING asks for the builder, and `ide_find_references`, `ide_search_text`, and `ide_structural_search_replace` already use that helper. It adds `minItems: 1`, so regenerate the golden manifest afterwards and review the diff:
  ```
  ./gradlew test -Ptier=unit --tests "*ToolManifestContractUnitTest" -Dcontract.update=true
  ```

### Nits

- `projectPath` on the new `excludeDirectories(...)` helper is unused.
- The inline fully qualified names (`com.intellij.openapi.command.WriteCommandAction`, `com.intellij.openapi.module.ModuleManager`, `com.intellij.openapi.roots.ModuleRootManager`) should be imports.
- A project-model change does not need an undoable command; a plain `WriteAction`, as in `CreateModuleTool.createModule`, is enough.
- The changelog entry could link the issue like the neighbouring entries do (`[#378](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/issues/378)`).

### One design note for the record

Excluding a directory is a persistent change to the user's project model: it shows as excluded in the Project view, drops out of the project search scope, and nothing in MCP can undo it. That is acceptable for the stated use case, but it is why the result should say exactly what changed and why the docs should mention that it persists.

### Status

CI has not run on this PR yet, so this review is a static read of the diff. Could you also confirm the symlink case from #378 was verified in a real IDE, i.e. a rename that previously aborted now succeeds after opening with `excludeDirectories`?

Happy to re-review once the above is in.

---

hechtcarmel:
> **Note:** this is a Claude Code message, drafted with the repo owner (@hechtcarmel).

Re-reviewed the updated head (383ba28). I see the PR is a draft now, so if this is mid-work, treat the below as a checkpoint against the first review rather than a verdict. Short version: one of the five blocking items is addressed in form, the rest are still open, and the update adds two unrelated behaviors that make the PR harder to accept than the first version.

### Where the first review stands

| Item | Status |
|---|---|
| Exclusions skipped when the project is already open | Addressed in form: the already-open path now applies them. Still nothing in the result says what happened. |
| Exclusion runs before `autoLink`, no smart-mode wait afterwards | Not addressed. Same order on both paths. |
| Silent failures, no result feedback | Not addressed. `findChild(...) ?: continue` and `catch (e: Exception) { rootModel.dispose() }` are unchanged, and the new `ensureContentRoot` swallows exceptions the same way. |
| No tests | Not addressed. The diff still touches no test file. |
| `CHANGELOG.md` conflict | Still conflicts. `main` has since moved the 5.9.6 entries out of `[Unreleased]`, so a rebase is needed either way. |
| Docs, naming/validation vs. `ide_create_module`'s `excludes`, `SchemaBuilder.stringArrayProperty`, nits | Not addressed. |

### Two new behaviors that should not be in this PR

1. **`ensureContentRoot` runs on every call.** It now executes on both paths, including an already-open project called with nothing but `path`. If no content root matches the raw path string, it adds the project directory as a content root of `modules[0]`, or creates a `WEB_MODULE` and writes an `.iml` into the project root. Three problems:
   - The check is a plain string comparison against `VirtualFile.path`, so a symlinked path, a trailing slash, or a Windows-style path never matches, and the code takes the add branch on a project that is already fully configured.
   - It turns an idempotent "open and wait" tool into one that mutates the module structure as a side effect, and drops a file into the user's checkout.
   - The platform already has a documented switch for exactly this case: `OpenProjectTaskBuilder.runConfigurators`, described as "Whether to run configurators (`DirectoryProjectConfigurator` instances) if `isNewProject` or has no modules." `ProjectUtils.openTask()` leaves it unset. Please try `OpenProjectTask { forceOpenInNewFrame = true; runConfigurators = true }` before hand-rolling module creation.

   Whatever the right fix turns out to be, it is a separate bug with its own changelog entry, its own tests, and ideally its own issue. Please split it out.

2. **`tryImportNestedMavenModules` folded into `autoLink`.** When no root build file exists, `autoLink` now scans immediate subdirectories for `pom.xml` and imports each one as a Maven module. That is what `ide_import_modules` is for. It is undocumented (the tool description still says `autoLink` links a build system), has no changelog entry and no tests, and on the already-open path the tool returns without waiting for the new modules to index. It also runs after the exclusions, so an excluded name inside a nested module is never applied. Please drop it from this PR or open it separately.

The PR description was not updated and mentions neither behavior.

### Status

CI still has not run on this PR. GitHub still reports it unmergeable. This is again a static read of the diff.

### What would get this merged

The original feature, trimmed back to what #378 asks for, with the items from the first review: exclusions applied after linking with a smart-mode wait, a result that lists excluded / already-excluded / not-found entries and surfaces failures, validation and parsing matching `ide_create_module`, the schema helper plus a regenerated manifest, a behavior test against the fixture project, docs, and a rebase onto `main`. Happy to re-review when it is marked ready.

---

hechtcarmel:
> **Note:** this is a Claude Code message, drafted with the repo owner (@hechtcarmel).

Re-reviewed the third revision (fab557c). Still a draft, still conflicting with `main` (now at 5.10.0), CI has still never run, and the PR title and description still describe the first version. The verdict has not changed: not mergeable, and the scope has grown again instead of closing the review items. Details below, then a clear statement of what this PR should contain so we stop going in circles.

### What this revision fixes

- Setup now runs on the already-open path, and the result reports counts of what it did.
- The `CLAUDE.md` inventory line is updated and the changelog entry links #378.
- A test file exists.

### Still open from the first two reviews

| Item | Status |
|---|---|
| Exclusion ordered before `autoLink` | Still before it (`applySetup` runs content root → exclude → link). The already-open path has no smart-mode wait at all. |
| Swallowed errors | Unchanged: `catch (e: Exception) { rootModel.dispose() }` in both helpers. `ensureContentRoot` then returns `true` and the result says "Content root registered." even when the write action failed. |
| Not-found / already-excluded names reported | Still silent. Only a count of new exclusions is reported. |
| Naming and validation vs. `ide_create_module`'s `excludes` | Unchanged. |
| `SchemaBuilder.stringArrayProperty` instead of hand-built JSON | Unchanged. |
| `USAGE.md` and the bundled `tools-reference.md` | Not updated. |
| Rebase | Still needed; `CHANGELOG.md` still conflicts. |
| Split out the content-root and nested-import behaviors | Not done. Nested **Gradle** linking was added on top. The `runConfigurators` suggestion was not answered. |
| Nits (unused `projectPath` param, inline FQNs, `WriteCommandAction` for a project-model change) | Unchanged. |

### New problems in this revision

1. **`autoLink` now defaults to `true`.** Every open, and every call on an already-open project, now attempts Maven/Gradle linking and nested-module import unless the caller opts out, and Gradle linking executes build scripts. That is a silent behavior change for every existing user of this tool, and it is filed in the changelog under "Added". A default flip like this needs its own issue and discussion; it cannot ride along in a feature PR.
2. **The tool description lost three true statements**: that at least one project must already be open (the JSON-RPC context), that the tool does not enroll the project in lifecycle management, and the trust-dialog caveat. Please restore them. The golden manifest diff now carries that loss.
3. **The new tests cannot fail.** `testEnsureContentRootAddsRootWhenMissing` asserts `contentRootsAfter >= contentRootsBefore`, which no implementation can violate. `testAutoLinkDefaultsToTrue` asserts only that the call succeeded. `testResponseIncludesAlreadyOpenMessage` checks text that predates this PR. CONTRIBUTING's rule is that a test must fail when the implementation is deleted; none of these would. The file's comment says module-root changes are untestable in the fixture because of the Workspace Model `NoSuchMethodError`. That is exactly what `suppressWorkspaceModelErrors` in `src/test/kotlin/.../testutil/LoggedErrors.kt` exists for, and `CreateModuleBehaviorTest.testHappyPath_createsModuleWithContentRootAndExcludes` uses it to assert on `excludeFolderUrls` today. Please use it and assert the real thing: after calling the tool with `excludeDirectories`, the module's content entry lists the excluded folder; a name that does not exist is reported as not found.
4. **Cosmetic:** a fresh open with no setup action now returns `"... is open and ready. "` with a trailing space, because `applySetup` always returns a non-null string.

### Scope for this PR

To be direct about it, since each revision has added more: **this PR is for #378 only, the `excludeDirectories` parameter.** Everything else needs its own issue and PR so it can be reviewed on its own merits:

- Content-root registration for plain directories, worktrees and slots. Try `OpenProjectTask { forceOpenInNewFrame = true; runConfigurators = true }` in `ProjectUtils.openTask()` first; that is the platform's documented mechanism for a project with no modules.
- Nested Maven/Gradle module detection under `autoLink`.
- Changing the `autoLink` default.

Please drop those three from this branch, including their changelog entries and the description rewrite, and rebase onto `main`.

### What gets this PR merged

1. Exclusions applied after linking, then a smart-mode wait, on both paths.
2. A result that lists excluded, already-excluded and not-found entries, and surfaces a failure instead of swallowing it.
3. Parsing and validation matching `ide_create_module` (`mapNotNull { (it as? JsonPrimitive)?.contentOrNull }`, reject blank and `..`), and a decision on the name.
4. `SchemaBuilder.stringArrayProperty`, then regenerate the manifest; the manifest diff should show the new property and nothing else.
5. A behavior test using `suppressWorkspaceModelErrors` that asserts the exclusion on the fixture project, plus a not-found case.
6. `USAGE.md` and `tools-reference.md` parameter tables updated; the original description text restored.
7. Rebased onto `main` with the changelog entry under `[Unreleased]`.

Happy to re-review once it is marked ready.

---

src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/project/OpenProjectTool.kt
```diff
@@ -80,14 +94,27 @@ class OpenProjectTool : AbstractMcpTool() {
...
+        ProjectUtils.findOpenProjectByPath(path)?.let { existing ->
+            val actions = mutableListOf<String>()
+            if (autoLink) {
+                val linkMsg = tryAutoLink(existing, path)
+                if (linkMsg != null) actions.add(linkMsg)
+            }
+            if (excludeDirs.isNotEmpty()) {
+                val excludeMsg = applyExclusions(existing, path, excludeDirs)
+                if (excludeMsg != null) actions.add(excludeMsg)
+                ProjectUtils.awaitSmartMode(existing)
```
hechtcarmel, line 114: **Blocking: this is what hangs the CI Test job.**

`ProjectUtils.awaitSmartMode` suspends until an `invokeLater` runnable runs on the EDT. The new tests call the tool through `runBlocking` on the EDT (platform tests run there by default; `UsefulTestCase.runInDispatchThread()` returns `true` and `McpPlatformTestCase` does not override it). `runBlocking` blocks the EDT, the runnable is never dispatched, and the call never returns. Every test that passes a non-empty `excludeDirectories` for an already-open project takes this path, which is three of the five.

That matches the CI log for this head: the `:test` task printed nothing for 56 minutes and was cancelled at the 60-minute job timeout, while the whole workflow takes about 15 minutes on `main`.

Production is not affected (the request thread is not the EDT), so the fix is on the wait itself: return immediately when the project is already in smart mode, or make `awaitSmartMode` EDT-safe by calling `DumbService.runWhenSmart` directly when `isDispatchThread` is true instead of going through `invokeLater`.

---

src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/project/OpenProjectTool.kt
```diff
@@ -140,6 +173,79 @@ class OpenProjectTool : AbstractMcpTool() {
...
+                            } else {
+                                contentEntry.addExcludeFolder(excludeUrl)
+                                notFound.remove(dirName)
+                                alreadyExcluded.remove(dirName)
+                                excluded.add(dirName)
+                                modified = true
+                            }
+                        }
+                    }
+                    if (modified) rootModel.commit() else rootModel.dispose()
```
hechtcarmel, line 229: **Blocking: the commit runs without a write action, so the exclusion is never applied.**

`edtAction` only moves the block to the EDT; nothing here takes the write lock. `ModifiableRootModelBridgeImpl.commit()` goes through `WorkspaceModel.updateProjectModel`, whose first line is `ThreadingAssertions.assertWriteAccess()`, so `commit()` throws "Write access is allowed inside write-action only". The `catch` below turns that into an `Errors:` entry, but `excluded.add(dirName)` already ran on line 224, so the response reports `Excluded: wksp` and an error for the same name, and the project model is unchanged.

The PR description says the exclusion runs inside `WriteCommandAction`; the code does not. `CreateModuleTool.createModule` wraps the same `addExcludeFolder` + `commit()` sequence in `WriteAction.compute`, so please do the same here (`edtAction { WriteAction.run<Exception> { ... } }`), and only record a name as excluded after the commit succeeds.

---

src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/project/OpenProjectTool.kt
```diff
@@ -140,6 +173,79 @@ class OpenProjectTool : AbstractMcpTool() {
...
+
+        edtAction {
+            val moduleManager = ModuleManager.getInstance(project)
+            for (module in moduleManager.modules) {
+                val rootModel = ModuleRootManager.getInstance(module).modifiableModel
+                try {
+                    var modified = false
+                    for (contentEntry in rootModel.contentEntries) {
+                        val contentRoot = contentEntry.file ?: continue
+                        if (!contentRoot.path.startsWith(projectPath)) continue
```
hechtcarmel, line 203: **Should fix: raw string prefix comparison silently skips roots.**

`projectPath` is the caller's argument as typed, while `VirtualFile.path` is always forward-slash normalized. A Windows path with backslashes, a trailing slash, or a symlinked path (`/var` vs `/private/var` on macOS) never matches, every content entry is skipped, all four lists stay empty, and the method returns `null`, so the caller gets "already open" with no mention of exclusions at all. That is the silent no-op the earlier reviews asked to eliminate. `findOpenProjectByPath` already canonicalizes through `ProjectUtils.canonicalNormalizedPath`; compare both sides the same way, and report when no content root matched.

This is also why the new tests cannot pass even without the deadlock: the light fixture's only content root is `temp:///src` (`LightProjectDescriptor.createSourceRoot` builds it under `temp:///`), so this filter skips it and the response contains neither `Excluded:` nor `Not found:`.

---

src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/project/OpenProjectTool.kt
```diff
@@ -140,6 +173,79 @@ class OpenProjectTool : AbstractMcpTool() {
...
+                                contentEntry.addExcludeFolder(excludeUrl)
+                                notFound.remove(dirName)
+                                alreadyExcluded.remove(dirName)
+                                excluded.add(dirName)
+                                modified = true
+                            }
+                        }
+                    }
+                    if (modified) rootModel.commit() else rootModel.dispose()
+                } catch (e: Exception) {
```
hechtcarmel, line 230: Nit: `catch (e: Exception)` still catches `ProcessCanceledException` and `CancellationException` and converts them into an `Errors:` string. Rethrow those first, the way `tryAutoLink` below does.

---

src/test/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/tools/project/OpenProjectExcludeDirectoriesBehaviorTest.kt
```diff
@@ -0,0 +1,134 @@
...
+ * Uses [suppressWorkspaceModelErrors] because `addExcludeFolder` triggers a Workspace Model
+ * event handler that fails with `NoSuchMethodError` due to a Kotlin version mismatch in the
+ * test sandbox — the exclusion still works, but the logged error would otherwise fail the test.
+ */
+class OpenProjectExcludeDirectoriesBehaviorTest : McpPlatformTestCase() {
+
+    fun testExcludesDirectoryAndReportsIt() = suppressWorkspaceModelErrors {
+        writeProjectFile("wksp/plans.md", "plan content")
+
+        val result = runBlocking {
```
hechtcarmel, line 25: **Blocking: these tests cannot have passed as written.**

Two independent reasons, both explained inline on `OpenProjectTool.kt`:

1. `runBlocking` here runs on the EDT, and the tool's already-open path now waits for smart mode through `invokeLater`, so this call deadlocks. CI's Test job sat silent for 56 minutes on this head before the timeout cancelled it.
2. The light fixture's module content root is `temp:///src`, not `project.basePath`, so the tool's content-root filter skips it and the response never contains `Excluded:` / `Already excluded:` / `Not found:`. The `excludeFolderUrls` assertion would fail for the same reason.

Please rebuild the three exclusion tests on a real on-disk content root. `CreateModuleBehaviorTest` shows the shape with `HeavyPlatformTestCase`, where the module's content root is the project directory; alternatively register the base directory as a content root of the light module first. The CONTRIBUTING rule that a test must fail when the implementation is removed applies here in both directions: right now they cannot pass with it present either.

The PR's test plan says the full suite passed. Given the above and the CI log, please re-run `./gradlew test` (platform tier included) before the next push and paste the result.

---

hechtcarmel (changes requested):
> **Note:** this is a Claude Code message, drafted with the repo owner (@hechtcarmel).

Reviewed the fourth revision (9fe2cab, one commit rebased on current `main`, six files). Thanks for trimming it back to #378 and working through the earlier lists: the already-open path, ordering after `autoLink`, per-name feedback, validation, `SchemaBuilder.stringArrayProperty`, the manifest, `USAGE.md` / `tools-reference.md`, the restored description, and the rebase are all in place in form, and the changelog entry is where it should be. The feature is still one we want.

It is not mergeable yet: the exclusion never applies in production, and the new tests deadlock, which is why CI's Test job on this head ran for 60 minutes with no output and was cancelled ([job log](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/actions/runs/35279815975/job/105401279551)). The same workflow finishes in about 15 minutes on `main`. Details are inline; summary below.

### Blocking

1. **Root-model commit without a write action** (`OpenProjectTool.kt:195-229`). `edtAction` only moves the block to the EDT. `ModifiableRootModelBridgeImpl.commit()` calls `WorkspaceModel.updateProjectModel`, which begins with `ThreadingAssertions.assertWriteAccess()`, so the commit throws, the `catch` reports it under `Errors:`, and because `excluded.add(dirName)` ran before the commit the same response also says `Excluded: wksp`. The description says this runs inside `WriteCommandAction`; the code does not. Wrap it in `WriteAction` as `CreateModuleTool.createModule` does, and record a name as excluded only after the commit succeeds.
2. **Smart-mode wait deadlocks under the test harness** (`OpenProjectTool.kt:114`). `awaitSmartMode` suspends on `invokeLater`; the tests call the tool through `runBlocking` on the EDT (`UsefulTestCase.runInDispatchThread()` defaults to `true`), so the runnable never runs. Return early when the project is already smart, or make `awaitSmartMode` call `runWhenSmart` directly when already on the EDT.
3. **The tests cannot pass as written** (`OpenProjectExcludeDirectoriesBehaviorTest.kt`). Beyond the deadlock, the light fixture's only content root is `temp:///src`, so the tool's path filter skips it and the response contains neither `Excluded:` nor `Not found:`. Rebuild the three exclusion tests on a real on-disk content root (`HeavyPlatformTestCase` like `CreateModuleBehaviorTest`, or register the base directory as a content root first). The test plan's "full test suite passes" does not hold for this head; please re-run the platform tier before the next push.

### Should fix

- `contentRoot.path.startsWith(projectPath)` compares the raw argument against a normalized VFS path. Backslashes on Windows, a trailing slash, or a symlinked path make every root skip silently and the tool returns no exclusion text at all. Canonicalize both sides (`ProjectUtils.canonicalNormalizedPath`) and report when no content root matched.
- `catch (e: Exception)` in `applyExclusions` still swallows `ProcessCanceledException` / `CancellationException`; rethrow them as `tryAutoLink` does.
- The `ide_open_project` line in `CLAUDE.md`'s tool inventory still lists only `autoLink`.
- The #378 scenario is a symlink. Please confirm in a real IDE that a rename which previously aborted succeeds after opening with `excludeDirectories`, and say so in the description.

### What gets this merged

The diff is small and focused now, and so are the fixes: the write action, an EDT-safe (or skipped) smart-mode wait, tests on a real content root, the path normalization, the rethrow, and the `CLAUDE.md` line. Then one green CI run on the head; none of the four revisions has had one yet.

### Status

`scripts/check-pr.sh` passes 15 of its 16 checks against this head; the remaining one is the test run. This review is a static read plus the CI log: the platform tier could not be executed in the review environment, so the three platform behaviors above were checked against the IntelliJ Community sources (`WorkspaceModelImpl`, `ModifiableRootModelBridgeImpl`, `LightProjectDescriptor`, `UsefulTestCase`) rather than by running the suite.

Happy to re-review once it is updated.

---

hechtcarmel (approved):
> **Note:** this is a Claude Code message, drafted with the repo owner (@hechtcarmel).

Re-reviewed the fifth revision (a4efffb, one commit rebased on current `main`). Every item from the previous review is addressed, and CI is green on this head for the first time in the PR's history, including the Test job at its normal duration. Approving.

### Verified

- **Write action.** `rootModel.commit()` now runs inside `WriteAction.run`, and a name is reported as excluded only after the committed model shows its exclude URL. This is the change that makes the feature work.
- **Deadlock.** The already-open path no longer waits for smart mode after excluding, which removes the EDT deadlock the tests hit. Reasonable call: an exclusion only changes the project model, and the fresh-open path keeps its waits, which run off the EDT in production.
- **Tests.** `setUp` registers the project base directory as a real content root and `tearDown` removes it, and the assertions are exact strings plus `excludeFolderUrls`. The three exclusion tests fail without the implementation.
- **Path comparison.** Both sides canonicalized with a separator-aware prefix check, and a "No content root matched" message where the old code returned nothing.
- **Cancellation** rethrown, **`CLAUDE.md`** inventory line updated, changelog entry under `[Unreleased]`, manifest diff limited to the new property.

### Non-blocking notes, fine to leave

- The post-commit verification loop walks every content entry of the module, not only the ones under the project path, so a name that was already excluded under an unrelated root of the same module could be reported as newly excluded. Edge case only.
- The separate `ProcessCanceledException` catch is unreachable on this platform (it already extends `CancellationException`). Harmless, and it mirrors `tryAutoLink`.
- The description still does not say whether the symlink scenario from #378 was tried in a real IDE. Worth a line for the record.

Thanks for sticking with it, @mdproctor.

---

PR #383, opened 2026-09-14
Add structured targets to definition and symbol-info requests

RomanPozdeev:
## Summary

`ide_find_definition` and `ide_symbol_info` now accept an additive nested `target` containing exactly one of `symbolId`, `position`, or `qualifiedName` with `language`. Existing top-level selectors remain supported, while mixed or ambiguous requests fail validation before PSI synchronization.

Nested symbol handles route to their owning project. The schema itself is the single switch for both routing and argument normalization, and a string-valued `target` such as the one used by `ide_run_tests` cannot accidentally enable semantic symbol routing.

This is the first slice of a planned rollout. The `symbol-handle-adoption` follow-up extends the same contract to references, implementations, super methods, rename, safe delete, change signature, and member-editing tools; the hierarchy-pagination follow-up covers call and type hierarchies. Until those follow-ups land, the flat selectors remain the primary wording in tool descriptions.

The MCP descriptions, user documentation, bundled skill reference, changelog, and golden tool manifest are updated together.

## Validation

- Added regressions for serializers that emit null-valued sibling fields inside `target`, including null-valued unknown fields inside `position`.
- Added a dispatcher regression for nested `target.symbolId` mixed with a top-level `file`; it routes to the owning project and then returns the normalization error naming `file` before tool execution.
- Regenerated and reviewed `ToolManifestContractUnitTest`; its diff is confined to the two tools, then it passed without update mode as part of the full check.
- `gw --no-log ./scripts/check-pr.sh`: 1,304 tests, 0 failures, 0 errors, 0 skipped on IntelliJ IDEA 2025.3.6.1 / JBR 21.
- XML review found no `NoSuchMethodError`; the only hidden logger entries are the unchanged occupied-port probes in `McpServerServiceTest` and `KtorMcpServerWatchdogTest` (`BindException: Address already in use`).
- Review-only local SDK/classpath wiring was removed before publication; the worktree is clean.

## Stack

Follows merged upstream PR #375. This PR is rebased onto the current `main` tip and contains one commit.

## Context

This focused pull request was split from #372 after review.


hechtcarmel:
Thanks for the second slice out of #372. Review of PR #383 at HEAD `3650e0d`.

**This comment was written by Claude Code and drafted together with the repository owner (@hechtcarmel).** It checked out the PR head, diffed it against `main` (`f27daf3`), read every production file in the diff plus the tests, ran the static checks from `scripts/check-pr.sh` (all pass), and confirmed CI on this head: Build, Test (`./gradlew check`), Inspect code and Verify plugin are all green. Paths below are under `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/`; line numbers are at `3650e0d`.

## Bottom line

Small, additive and correctly confined to argument handling. A request without `target` takes an unchanged path (`normalize` returns its input untouched), the nested form is translated into the flat parameters before PSI sync, no PSI or threading is involved, and the golden manifest diff is exactly the two tools. **Mergeable after finding 1, and preferably 2 and 3; the rest are nits.**

One thing to settle alongside the fixes: `target` adds no capability, since every shape it accepts already works as flat parameters and #375 already rejects mixed selectors. It earns its schema cost only if it reaches the other target-taking tools (`ide_find_references`, the hierarchies, `ide_find_implementations`, `ide_find_super_methods`, rename, safe delete, change signature). On two tools alone it is a second dialect for the same request. Please say in the PR whether that rollout is planned for the follow-ups; the wording in finding 3 follows from the answer.

## Findings

**1. Should-fix: null-valued sibling keys inside `target` are rejected.** `tools/UnifiedTargetArguments.kt:58-60` count a variant as present only when its value is not `JsonNull`, but the unexpected-field checks at `:78`, `:87`, `:91` and `:112` are computed from every key. A client that serializes its optionals, `{"target": {"position": {...}, "symbolId": null, "qualifiedName": null, "language": null}}`, fails with `target variant 'position' contains incompatible field(s): language, qualifiedName, symbolId`. That contradicts the null tolerance the same file documents at `:40` for `target` itself, and the top-level `symbolId: null` handling from #375. Fix: compute the unexpected sets from `target.filterValues { it != JsonNull }.keys`, and the same for `position`.

**2. Should-fix (scope): the cursor carve-out changes six tools outside this PR and is a no-op for what it protects.** `server/mcp/McpToolDispatcher.kt:131-136` strips `target` and every legacy selector from the arguments of any tool that declares `cursor` when a cursor is present, and `:143` skips handle routing. None of the six paginated tools (`ide_find_references`, `ide_find_implementations`, `ide_find_class`, `ide_find_symbol`, `ide_search_text`, `ide_find_file`) declares `symbolId` or `target`, so routing never consulted those keys for them, and all six branch on the cursor before reading any selector, so the stripped keys were already ignored. The only observable effect is on the arguments those tools and the command history receive. The #375 review listed the earlier form of this carve-out as unused scaffolding and it was removed there; it is back, larger, with tests pinning it. Drop it, or land it with the PR that gives a paginated tool a `target`. `DispatcherCursorRoutingBehaviorTest` and the cursor test in `McpToolDispatcherTest` go with it.

**3. Should-fix: descriptions.** `tools/navigation/FindDefinitionTool.kt:40` and `SymbolInfoTool.kt:57` now call the flat selectors "Legacy", while every other target-taking tool accepts only the flat form. An agent that takes the hint and sends `target` to `ide_find_references` gets "Must specify file+line+column, or language+symbol" and a wasted round-trip. Keep the flat form as the primary wording until the rest accept `target`. Also `tools/schema/SchemaBuilder.kt:92` lists `className`, which neither tool has, and `:107` describes `position.file` as project-relative only, while the top-level `file` on both tools also accepts library absolute paths and `jar://` URLs; the normalized value goes through the same resolver, so the nested one does too.

**4. Nits.**
- `__unifiedTargetVariant` (`UnifiedTargetArguments.kt:83`, `:108`, `:120`) is injected into the arguments every `doExecute` receives and nothing in production reads it.
- Support is switched twice: `supportsUnifiedTarget` (`tools/AbstractMcpTool.kt:151`, used at `:252`) and the object-typed `target` schema property the dispatcher keys off (`McpToolDispatcher.kt:139-141`). A tool that sets one without the other either routes by a nested handle and then ignores it, or the reverse. Derive one from the other.
- Schema cost: the `target` property is 790 bytes minified plus about 80 bytes of description per tool, roughly 1.7 KB more on every `tools/list`; the golden manifest grows 3 KB (2.7%). Fine for two tools, but every further tool pays the same, which is worth keeping in mind for the rollout.
- The `CHANGELOG.md:9` bullet's last two sentences are implementation notes, not user-visible behavior.
- `UnifiedTargetArgumentsUnitTest.kt:78-84` asserts "exactly one" for the mixing case; it passes only because the mixing message happens to contain those words, and it does not check that the offending selector is named.

**5. Untested.** The null-sibling case in finding 1. A nested `target.symbolId` combined with a top-level `file` sent through the dispatcher: routing accepts the handle in `symbolIdForRouting`, then `normalize` rejects the mix, and one test should pin which error the client sees. The `Assume` on the Java plugin in `UnifiedTargetToolBehaviorTest` never skips on this classpath, which is fine, but the first test in that class runs without it.

## Tests

Real and well targeted. The unit tests over `normalize` fail on inverted logic, `DispatcherCursorRoutingBehaviorTest` pins the "only when the tool declares `cursor`" condition against a real tool, the string-valued `target` probe protects `ide_run_tests`, and `UnifiedTargetToolBehaviorTest` executes both tools end-to-end through `writeProjectFile` and `registerSourceRoot` with a byte-for-byte check that a lookup mutates nothing. Documentation landed in the per-tool sections this time, as asked on #375.

---

RomanPozdeev:
Thanks for the detailed review. Addressed all findings in `02ffd93`.

1. Null-valued sibling keys are now excluded before both the `target` and nested `position` unexpected-field checks. Added regressions for the serializer shape from the review and for a null-valued unknown field inside `position`.
2. Removed the cursor carve-out from the dispatcher and `UnifiedTargetArguments`, together with the tests that pinned it. Cursor semantics can be designed when a paginated tool actually adopts `target`.
3. Kept flat selectors as the primary wording, described nested `target` as an equivalent form, removed the irrelevant `className` list from the schema description, and documented that `target.position.file` accepts project-relative paths plus returned library absolute / `jar://` paths. Regenerated the golden manifest; its follow-up diff is limited to these two tools.
4. Removed `__unifiedTargetVariant`; nested-target support is now derived from the object-valued schema property in both normalization and dispatcher routing. Tightened the changelog to user-visible behavior, and fixed the member-selector assertions to require the offending selector name. I also shortened the shared target description; the per-tool schema cost remains an intentional consideration for the rollout.
5. Added the dispatcher regression for `target.symbolId` mixed with top-level `file`. It proves routing accepts the valid handle first, then normalization returns the mutually-exclusive error naming `file` before the tool body executes.

The rollout is planned and is now explicit in the PR body: the `symbol-handle-adoption` follow-up covers references, implementations, super methods, rename, safe delete, change signature, and member editing; `hierarchy-pagination` covers call/type hierarchies.

Validation on the exact published head:

- one commit over `main@f27daf3`, `git diff --check` clean;
- `gw --no-log ./scripts/check-pr.sh`: 1,304 tests, 0 failures/errors/skips;
- immutable PR snapshot runner: PASS, GitHub/local file parity and clean detached worktree;
- Build, Test, Inspect code, and Verify plugin are all green.

PR #385, opened 2026-09-14
Add dry-run previews for rename refactoring

RomanPozdeev:
## Summary

Add a non-mutating `dryRun` mode to `ide_refactor_rename`. Preview and apply now share target resolution, automatic rename selection, usage discovery, conflict detection, and affected-file calculation. A preview returns the common refactoring plan shape without changing PSI, saving documents, or creating an undo command. Legacy coordinates, opaque `symbolId` handles, and the structured `target` selector are supported.

The change also keeps Kotlin override renames headless: base-method discovery runs off the EDT through the Kotlin light-method API, resolves back to the source declaration, and fails closed instead of entering an IDE chooser. Coverage includes Java/Kotlin override trees, constructor-to-class renames, implicit file/directory collisions, read-only scope, automatic related renames, JS/TS preparation that would require interaction, and end-to-end previews for nested `target.position` and legacy `language` + `symbol` selectors.

## Stack

Depends on merged upstream PR [#383](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/383). This branch is restacked on `upstream/main@8354a36` and contains one commit relative to that base.

## Testing

- `./gradlew test -PkotlinPluginTests=true --tests '*Kotlin*BehaviorTest' --no-configuration-cache --rerun-tasks` — clean-checkout recipe; 9 tests, 0 failures/errors/skips
- Targeted review-fix gate (`RenamePreviewTargetBehaviorTest`, `RenameSymbolToolBehaviorTest`, `ToolsUnitTest`, and `ResultShapeContractUnitTest`) — 45 tests, 0 failures/errors
- `./scripts/check-pr.sh` — PASS; 140 test-result XML files, 0 unexpected findings
- `git diff --check upstream/main..HEAD` — PASS

The local review ran these commands through `gw` with IntelliJ IDEA 2025.3.6.1 / JBR 21 and complete disposable local SDK/runtime wiring, restored after each gate. This build does not read `-PlocalIdePath`; that property was not used as a substitute for the full wiring. Test XML contains no `NoSuchMethodError`; the only background findings are the documented occupied-port baseline in `McpServerServiceTest` and `KtorMcpServerWatchdogTest`.

## Context

This focused pull request was split from upstream PR [#372](https://github.com/hechtcarmel/jetbrains-index-mcp-plugin/pull/372) after review.

PR #392, opened 2026-09-15
Preview Java and Kotlin signature changes with exact targets

RomanPozdeev:
## Summary

- add non-mutating `dryRun` previews to `ide_change_signature`, with the shared refactoring preview shape, exact `symbolId` handles, nested targets, and qualified Java targets
- support Kotlin JVM functions by source position or handle; selecting an override resolves the base declaration so interfaces, implementations, and callers stay consistent
- preview and apply fail closed on incomplete usage/conflict discovery, discovered conflicts, and read-only scope; both reject missing defaults for new required caller/delegate arguments, which would otherwise produce non-compiling calls
- guard interactive overrider decisions; successful apply returns current `updatedSymbol` metadata
- keep preview and apply aligned for caller discovery, generated delegates, trailing varargs, unchanged return types, package-private visibility, and externally created coordinate targets
- preserve existing Java `throws` clauses on apply

## Review follow-up

- use the public static `ChangeSignatureProcessorBase.findUsages(ChangeInfo)` API via reflection to keep Java optional; no accessibility override or protected processor hook
- share strict Kotlin light-method discovery between change signature and rename, preserving the existing tolerant utility contract
- replace obsolete/deprecated API calls and simplify exception propagation
- apply now collects the same conflicts as preview before entering the mutating processor; a duplicate-signature regression asserts refusal and unchanged declaration/caller files
- retain and document the conservative refusal of any return-type change with overriders, even where narrower overriding return types could safely remain unchanged
- document the existing Java override limitation: Java targets do not automatically redirect to their base; select the base explicitly for hierarchy-wide changes. Java/Kotlin target-selection parity remains a follow-up

## CI runtime correction

- CI on `5382d19` failed in `ChangeSignatureExternalFileBehaviorTest`; its report exposed `NoSuchMethodError: SequencesKt.sequenceOf(Object)` inside platform progress support, with the same hidden linkage error across 91 class reports
- the default test configuration allowed Gradle's Kotlin stdlib to shadow the IDE runtime; its exclusion was previously limited to opt-in Kotlin plugin tests
- `270247e` makes the stdlib/jdk7/jdk8 exclusions unconditional for test runtime. Production refactoring code and test assertions are unchanged
- machine-local SDK wiring had already supplied the exclusion and masked this CI mismatch; the fresh branch-local runs used the committed runtime fix, with temporary wiring limited to SDK/compiler/test-framework artifacts

## Validation

- contract goldens regenerated and verified cleanly
- 25 default-mode targeted tests and 28 opt-in Kotlin behavior tests passed with no failures, errors, skips, or hidden XML findings
- dependency insight confirms no external Kotlin stdlib in the default test runtime
- full local `check-pr.sh` gate passed: 1,361 tests, no failures, errors, or skips; XML contains only the two documented occupied-port baselines and no unexpected findings
- published-head verification passed on `270247e`: Kotlin gate and full-gate retry green, one commit over `main@4743d181`, GitHub/local file parity, clean worktree, no unexpected XML findings. The first immutable full-gate run hit a separate local port collision with Android Studio on port 63342; that failed report was retained and not treated as baseline noise
- GitHub Build, Test, Inspect code, and Verify plugin passed on `270247e` in run `34943854297`; Release draft was skipped as expected for a PR
- successful CI test reports are not uploaded by this workflow (`failure()` only), so the remote confirmation is the passing Test job/log. Hidden XML findings were checked in the branch-local and immutable local gates

## Stack

Follows merged #389.


hechtcarmel:
Hi,
First of all thank you for all those great contributions!
Second, note that the ci fails

hechtcarmel:
oh, you just updated 😅

RomanPozdeev:
Thanks! The failure was caused by `RefactoringImpl` being marked as an internal API in newer IDE builds. Fixed in `96eb5a2` by removing that dependency while keeping usage discovery fail-closed. The rerun is now green, including `Verify plugin`.

hechtcarmel:
Thanks for the third slice, and for the quick turnaround on the verifier. Review of PR #392 at HEAD `96eb5a2`.

**This comment was written by Claude Code and drafted together with the repository owner (@hechtcarmel).** It checked out both heads (`536c027`, then `96eb5a2`), diffed the current one against `main` (`4743d18`), read every production file in the diff plus the tests, checked the guards against the 2025.3 platform sources (`ChangeSignatureProcessor`, `ChangeSignatureProcessorBase`, `JavaChangeSignatureUsageProcessor`, `ParameterInfoImpl`, `JavaThrownExceptionInfo`, `KotlinChangeSignatureUsageProcessor` for K2), ran the static checks from `scripts/check-pr.sh` (all pass), and confirmed CI on this head: Build, Test (`./gradlew check`), Inspect code and Verify plugin are all green. It could not run the Gradle suite in its sandbox, and CI never runs the `-PkotlinPluginTests=true` tier, so the Kotlin claims rest on your local run. Paths below are under `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/`; line numbers are at `96eb5a2`.

## Bottom line

This is the right shape: `ide_change_signature` now has the same preview contract and target selectors as rename (#385) and safe delete (#389), and the new fail-closed guards map onto real modal prompts in the platform. `ChangeSignatureProcessor.preprocessUsages` calls `Messages.showYesNoDialog` for weakened visibility with overriders and again in `isProcessCovariantOverriders`, both skipped only in unit-test mode, so before this PR a headless call on those inputs blocked on a dialog in the IDE. The PR also fixes a real bug on `main` that nobody had noticed (finding 3). The update between the two heads rebased onto `main` and replaced `RefactoringImpl`, which is `@ApiStatus.Internal` on the 2026.x builds the verifier targets and was the `INTERNAL_API_USAGES` failure on `536c027`. **Mergeable as is. Finding 1 is worth doing before merge because it is a one-liner; the rest are nits and follow-up notes.**

## Findings

**1. Should-fix: the verifier fix hides the problem rather than removing it.** `tools/refactoring/RefactoringScopeGuard.kt:123-150` (`findUsagesReflectivelyStrict`) walks the class hierarchy to `ChangeSignatureProcessorBase.findUsages()` and calls `trySetAccessible()` at `:132`. That method overrides `BaseRefactoringProcessor.findUsages()`, which is `@ApiStatus.OverrideOnly` in 253; reflection keeps it out of the verifier report (and `trySetAccessible()` sidesteps the `isAccessible = true` grep at `scripts/check-pr.sh:75`). Not a regression, since `main` did the same in `findUsagesReflectively`. But the protected instance method is literally `return findUsages(myChangeInfo)`, and that static overload, `ChangeSignatureProcessorBase.findUsages(ChangeInfo)`, is public. Both call sites (`ChangeSignatureTool.kt:393` and `:671-672`) already have `changeInfo` in scope. Calling the static method gives identical usages with no reflection, no accessibility override and no class walk. Direct or reflective is your call; it is a platform class either way.

**2. Nit (message accuracy): the default-value warning describes a dialog that cannot open for this tool.** `ChangeSignatureTool.kt:819-820` says the IDE "may require an interactive default-value decision". Every `ParameterInfoImpl` is built with `defaultValue ?: ""` (`:767`), and `JavaChangeSignatureUsageProcessor.setupDefaultValues` opens `DefaultValueChooser` only when `getDefaultValue() == null`. With an empty default the processor's `createDefaultValue` path yields no expression (`ParameterInfoImpl.getValue` returns null), so the real outcome of applying is a call site rewritten without an argument for the new parameter: non-compiling callers, no prompt. Failing closed is still the right call; the message, the CHANGELOG bullet and the PR description ("would require an interactive dialog") should say what actually happens.

**3. Nit (changelog): the Fixed section describes this PR's own iterations and misses the real fix.** `CHANGELOG.md:12-13` list behaviors of previews that did not exist on `main`. The genuine user-visible fix is not named: on `main`, `ChangeSignatureTool.kt:253` looked up `JavaThrownExceptionInfo(int, PsiType)` reflectively, but the constructor takes `PsiClassType` on 2025.3, so every apply took the `:260` fallback and passed an empty throws array; `JavaChangeInfoImpl` treats the length mismatch as `isExceptionSetChanged` and the processor stripped the method's `throws` clause. Your `extractExceptions` call (`:591`) fixes that and `testChangeSignaturePreviewAndApplyPreserveExistingThrowsForVisibilityChange` pins it. Suggest one Fixed bullet, "Change signature no longer strips `throws` clauses on apply", and fold the rest into the Added bullet.

**4. Nit (API hygiene).** `RefactoringScopeGuard.kt:165` uses `Application.assertIsDispatchThread()`, which is `@ApiStatus.Obsolete`; `ThreadingAssertions.assertEventDispatchThread()` is the replacement. `:191` adds another `ReadAction.run(ThrowableRunnable)` call, which the verifier lists as deprecated.

**5. Nit (dead branches).** `extractThrownExceptions` at `ChangeSignatureTool.kt:605-608`: all three `when` arms are `throw cause`, so the `when` is a plain `throw cause`.

**6. Nit (reuse).** `resolveSignatureMethod` (`:340-346`) re-implements `PsiUtils.toLightMethods` (`util/PsiUtils.kt:419`) inline. The reason is clear, the helper swallows failures with `runCatching`, which contradicts fail-closed, but a strict variant on `PsiUtils` would keep the Kotlin reflection in one place; `RenameSymbolTool.kt:1724-1737` carries a third copy for the same deepest-super resolution.

**7. Design notes, not blocking.**
- Preview reports conflicts as `canApply: false` (`:455-456`), but apply never collects them. `ChangeSignatureProcessor(Project, JavaChangeInfo)` reaches `BaseRefactoringProcessor(project, null)`, so `preprocessUsages` skips the conflicts dialog and proceeds. Either say in the description that apply does not stop on conflicts, or collect the same conflicts on apply and refuse, which would match safe delete.
- `:837-842` refuses any return-type change while overriders exist. The platform (`askToRemoveCovariantOverriders`) only prompts when an overrider's return type is a strict covariant of the new one. Over-conservative but safe; the message could say "overriders with a narrower return type".
- Kotlin overrides redirect to the deepest super (`:347`); a Java override selected by position or handle still changes only itself (`:328`), which leaves the base untouched and the `@Override` broken. Pre-existing for Java, but the asymmetry is now documented behavior. Applying the same `findDeepestSuperMethods` to Java in a follow-up would make both languages consistent.

## Tests

Real and well targeted. `RefactoringDryRunBehaviorTest` keeps the byte-for-byte snapshots and hooks `run()` so a dry-run that reaches the processor fails; the missing-default test proves apply refuses before the processor is entered; the registered `ChangeSignatureUsageProcessor` extension proves preview metadata matches what apply rewrites when an extension replaces the usage ref; `ChangeSignatureExternalFileBehaviorTest` covers the VFS-miss case that motivated moving file discovery ahead of the read action; the delegate and override-only default tests assert both directions. `KotlinChangeSignatureBehaviorTest` covers position and handle lookup, override-to-base redirection with the handle staying bound to the override, read-only base detection before the processor, and the aborted-processor path, but it lives in the opt-in tier that CI does not run. The gaps your CONTRIBUTING note lists (extension and suspend parameter mapping, Kotlin-only type syntax) are the ones to watch when this reaches real projects.

---

PR #398, opened 2026-09-16
Bound IDE calls and format replaced Kotlin bodies

RomanPozdeev:
### Summary
- Bound ordinary MCP tool execution to a cancellable 55-second deadline while preserving long-poll budgets.
- Propagated cancellation through blocking diagnostics and queued EDT actions.
- Format complete replaced Java/Kotlin members and report final body positions after formatting/import optimization.

### Validation
- Branch is restacked as one commit on the current upstream main.
- Added regression coverage for timeout/cancellation and Java/Kotlin member formatting.
- Kotlin behavior tests and the full check-pr gate run locally before merge.


hechtcarmel:
Thanks for this one. Review of PR #398 at HEAD `2f5a07b`.

**This comment was written by Claude Code at the request of the repository owner (@hechtcarmel).** It fetched both heads (`1d49304`, then `2f5a07b`; the delta is one removed unused import in `DiagnosticsAnalysisService.kt`), diffed against `main` (`86e7980`), read every production file in the diff plus the new and changed tests, ran the static checks from `scripts/check-pr.sh` (15 of 15 pass), and confirmed CI on this head: Build, Test (`./gradlew check`), Inspect code and Verify plugin are all green in run `35058811100`. It could not run the Gradle suite in its sandbox (the JetBrains download hosts are blocked there), and CI never runs the `-PkotlinPluginTests=true` tier, so `KotlinReplaceMemberFormattingBehaviorTest` rests on your local run. Paths below are under `src/main/kotlin/com/github/hechtcarmel/jetbrainsindexmcpplugin/`; line numbers are at `2f5a07b`.

## Bottom line

The right change at the right layer. CLAUDE.md already says no tool call may block past ~55s (issue #277); this enforces it once in `McpToolDispatcher` instead of per tool, and the two helpers fix real problems: a request that timed out while its edit was still queued behind a modal dialog used to keep waiting for EDT (the old `withContext(Dispatchers.EDT)` only returns once EDT dispatches the cancelled block), and a blocked `findCodeSmells` could not be released by `withTimeoutOrNull` at all. The coroutine work checks out: `cancellableEdtAction` keeps thread-context elements (the server epoch) while dropping the parent `Job`, so the cancelled launch never runs its body; `cancellableBlockingAction` restores the coroutine's own cancellation via `ensureActive()` while letting an independent `ProcessCanceledException` through, and `runInterruptible` joins the worker before the diagnostics lock is released. **Mergeable after findings 1 and 2, both small.**

## Findings

**1. Should-fix: two tools lose their documented contract.** `ide_open_project` waits cooperatively for indexing up to `timeoutSeconds` (`tools/project/OpenProjectTool.kt:101`, default 600 at `:162`), and `ide_open_workspace` does the same (`tools/project/OpenWorkspaceTool.kt:141`, `:299`). Neither is in `LONG_POLL_TOOLS` (`server/mcp/McpToolDispatcher.kt:77-79`), so both now end at 55s with the generic dispatcher message at `:174`, while their own timeout branches ("still indexing after Ns", "retry with a larger timeoutSeconds") become unreachable and the docs still promise 600s (`USAGE.md:1253`, `USAGE.md:1937`, `src/main/resources/skill/ide-index-mcp/references/tools-reference.md:724` and `:768`, `src/main/resources/skill/ide-index-mcp/SKILL.md:187`). Either add them to the exempt set, or clamp `timeoutSeconds` to the budget and update those five doc lines. The long-poll shape would be the policy-correct end state, but that is a follow-up, not this PR.

**2. Should-fix: the CHANGELOG bullets are in the released section.** The two new entries sit at `CHANGELOG.md:25-26`, inside `## [5.15.0]` (`:7`), while `## [Unreleased]` at `:5` is empty. Looks like a rebase artifact from the 5.15.0 cut; `check-pr.sh` only checks that an Unreleased header exists, so it cannot catch this. Please move them.

**3. Nit (maintainability).** The exempt list is a hardcoded name set in the dispatcher, but `CLAUDE.md:290` still says a new long-running tool "plugs in with three pieces". Either add a fourth bullet there, or expose it as a property on `McpTool` so a new long-poll tool cannot forget it.

**4. Nit (description accuracy).** `tools/refactoring/ReplaceMemberTool.kt:274-275` now reformats the whole declaration, including its signature, annotations and KDoc; the tool description still says "Auto-reformats the changed range". Fine as a behavior, and the idempotence assertions cover it, but the description should say so.

**5. Nit (whitespace).** `server/mcp/McpToolDispatcher.kt:74` un-indents the KDoc closer above `HISTORY_RESULT_LIMIT` by one space; unrelated to the change.

**6. Behavior note, not blocking.** `withTimeoutOrNull` discards a value returned after the deadline, so a write that could not be cancelled and finished at 60s is reported as a timeout even though it applied. The error text at `:174-177` already says exactly that, so this is documented, but it is worth knowing when reading agent transcripts.

## Tests

Well targeted, and each one fails on `main` for the right reason. `McpToolDispatcherTimeoutTest` covers the timeout error and history status, the queued edit that must not run once EDT frees up (the old `withContext` path hangs its 2s wait), caller cancellation not being converted into a timeout, thread-context propagation into the EDT action, and the long-poll exemption. `CancellableBlockingActionTest` pins the two release mechanisms separately (indicator cancel for platform loops, interrupt for blocking waits) plus join-before-return and preserved independent PCE. `GetDiagnosticsToolBehaviorTest.testBlockingDetectorTimesOutAndReleasesTheAnalysisLock` models the real blocking wait with a latch and proves the next analysis gets the lock. `JavaReplaceMemberFormattingBehaviorTest` verifies both the closing-brace formatting and the post-import-removal line numbers, and `KotlinReplaceMemberFormattingBehaviorTest` covers block and expression bodies, initializers, `reformat: false` byte-for-byte, and full-file idempotence, but it lives in the opt-in tier CI does not run.

---

PR #400, opened 2026-09-16
Paginate hierarchy results with exact declaration identities

RomanPozdeev:
Builds on #398. Adds bounded pagination for call and type hierarchies while preserving exact declaration identities and legacy response compatibility.

The tool schema budget is set to 115,000 bytes because hierarchy pagination adds required cursor, page, and exact identity fields to both hierarchy tools. Keeping the full descriptions preserves discoverability and the protocol contract; the increase is deliberate and bounded rather than a temporary test workaround.

Validation: full local test suite passes on IntelliJ IDEA 2025.3.6.1 / JBR 21 using the local SDK wiring.


hechtcarmel (changes requested):
**Summary.** The feature is a good idea and the core design is solid: the tool owns the BFS, handlers answer one edge at a time, dedup is by smart-pointer identity, retries replay the same successor cursor, and the terminal look-ahead budget keeps each call bounded. The new tests are well built. But the branch was validated against your fork rather than `main`: CI's Test job fails 12 of 1516 tests, seven of them this PR's own new tests. The PR also rewrites released CHANGELOG history and deletes five unrelated dispatcher tests. Requesting changes; happy to re-review after a rebase.

### Blocking

**1. CI Test job is red (12 failures), all traceable to fork drift**

| Failing tests | Cause on `main` |
|---|---|
| `HierarchyTargetIdBehaviorTest` ×4, `SymbolIdBehaviorTest.testTypeHierarchyAcceptsIdAndReturnsResolvableIds` | `TypeHierarchyTool.kt:526` and `CallHierarchyTool.kt:494` call `resolveElementFromArguments(...)` without `allowSymbolId = true`. On `main` that flag defaults to `false`, so `symbolId` input is rejected with "Must specify file+line+column or language+symbol". |
| `ToolManifestContractUnitTest`, `ResultShapeContractUnitTest` | The golden `tool-manifest.json` expects a `target` description ("…Excludes top-level symbolId/file/line/column/language/symbol/className.") that `main`'s `SchemaBuilder.target()` does not produce. Both goldens need regenerating against `main`. |
| `ToolSchemaBudgetUnitTest` | tools/list is 112,733 bytes; the budget is 110,000. The new tool descriptions need trimming. |
| `McpToolDispatcherTest.testOneServerEpochIsCapturedBeforePreExecutionChecksForEveryRegistry`, `…testPaginationCursorIgnoresTargetSelectorsForRouting` | Both expect dispatcher behavior (epoch capture before pre-execution checks; stripping `symbolId`/`target` when `cursor` is present) that `main`'s `McpToolDispatcher` does not have and this PR does not add. The cursor test is also misnamed: the dispatcher has no cursor logic, so the probe tool only exercises the pre-existing "undeclared symbolId is ignored" schema gate. |
| `HierarchyIdentityAndHandlesBehaviorTest` ×2 | Legacy responses evict handles they just returned against `main`'s 4,096-entry `SymbolIdRegistry`. |

Please rebase onto `main` and run the full `./gradlew test` (platform tier included) before pushing.

**2. `CHANGELOG.md` deletes released history**

The diff removes the `## [5.11.0]`, `## [5.10.0]` and `## [5.9.6]` sections, drops several of their entries entirely (the `ide_diagnostics` `files` batch, bind-host validation, `ide_sync_files` batch validation, host-header protection, …), and re-homes others under `[Unreleased] / Added` as if they were new in this PR. Only this PR's own entries belong under `[Unreleased]`; the released sections must be restored verbatim. (`scripts/check-pr.sh` only flags *added* version headers, so it did not catch this.)

**3. Five existing `McpToolDispatcherTest` tests are deleted with no dispatcher change**

`testUndeclaredSymbolIdDoesNotRouteIndexStatus`, `testUndeclaredNestedTargetDoesNotRouteIndexStatus`, `testStringValuedTargetSchemaDoesNotEnableNestedSymbolRouting`, `testNullSymbolIdIsTreatedAsAbsent` and `testSymbolIdRoutesToItsOwningProjectWhenMultipleProjectsAreOpen` (from #375 / #383) are gone, and their imports are left dangling, which looks like a stale-merge accident. Please restore them.

**4. Docs are appended, not updated**

`CLAUDE.md`, `README.md` and `SKILL.md` each get a block appended *after* their footer that re-describes features from #375–#398 (symbol handles, structured targets, rename/safe-delete/change-signature previews). `CLAUDE.md` ends up with four duplicated headings (Structured lookup targets, Rename preview, Safe-delete preview, Change-signature preview). The sentence "Java fixtures and real Kotlin class-kind tests were run. Python/Go/PHP/Rust/JavaScript collector tests use neutral PSI…" is PR commentary, not documentation. Please keep only the hierarchy-pagination section, and update the `ide_type_hierarchy` / `ide_call_hierarchy` entries in the CLAUDE.md tool list inline.

### Should be fixed in the same round

- **Java swallows cancellation on the new reflective path.** `kotlinCallableSource` (`JavaHandlers.kt:249`) calls `getMethod(accessorName).invoke(source)` inside `findCalleesRecursive`'s `try`. A `ProcessCanceledException` / `IndexNotReadyException` from `KtProperty.getGetter()` arrives wrapped in `InvocationTargetException`, misses the three explicit rethrow clauses and lands in the catch-all at `JavaHandlers.kt:966`, producing an empty "complete" callees result. Use `rethrowIfControlFlow()` as the other handlers do (same pre-existing holes at `:1003` and `:1055`).
- **Legacy call-hierarchy paths changed behavior and cost**, contrary to "preserve legacy trees and limits":
  - Java / Go / JS / PHP callers lost their per-level reference-processing cap (`MAX_RESULTS_PER_LEVEL * 2` raw references in Java/Go/PHP, `MAX_BARREL_REFERENCES_TO_COLLECT = 120` in JS) and now scan every reference until 20 unique callers, per level, at the default depth of 3. A function with thousands of call sites from a handful of callers pays the full scan, with no `checkCanceled()` in the JS/PHP consumers.
  - PHP callees: two passes of 20 (method refs, then function refs, up to ~40) became one document-order pass capped at 20. Rust callees: up to 40 → 20. Python callers/callees: dedup key changed from `class.function` to `file|offset`, so legacy results can gain entries.
  
  Either restore the caps on the `page == null` path or list these as behavior changes in the CHANGELOG.
- **Late pages of wide hierarchies are quadratic.** `pendingCallNode` / `pendingTypeNode` dedup by scanning every visited smart pointer with `sameHierarchyDeclaration` per candidate, and every expansion re-reads from offset 0 with a growing prefix. For a node with thousands of direct children that is O(N²) PSI comparisons on late pages, which is exactly the case pagination exists for. A per-page map keyed by (virtual file, text offset), rebuilt from the visited pointers, makes it O(1) per candidate. `JavaHandlers.kt:924` and `:1018` have the same O(n²) `callees.none { callElementKey(it) == callElementKey(element) }` pattern; hoist the key and use a set.
- **Wasted work in paged mode.** Rust `getTraitHierarchy` / `getTypeImplHierarchy` compute both directions and `paginateTypeData` (`RustHandlers.kt:743`) throws one away, so `ide_type_hierarchy` on a widely used struct runs a full `ReferencesSearch` twice per page (Python correctly skips the unrequested direction). Python `getSubtypes` (`PythonHandlers.kt:392`) with `directOnly` runs the *deep* `PyClassInheritorsSearch` and walks every transitive descendant through `isVisibleSubtypeOf` per expansion; Java's non-deep search plus re-expansion of excluded children (`JavaHandlers.kt:578`–`595`) is the cheaper pattern already in this PR.
- **Bridging of excluded intermediates is unverified for JS/PHP.** JS strategy 1 uses the single-arg `JSInheritorsSearch.search(jsClass)` and returns as soon as it yields anything, and PHP's paged subtypes come from `DefinitionsScopedSearch` (`PhpHandlers.kt:1422`) while legacy uses `PhpIndex.getAllSubclasses`. Both rely on the native query being deep, and the new TS/PHP tests inject fake executors, so the real inheritor index is never exercised. Consider one provider for both modes.
- **Handler contract gaps.** `page != null` with `direction == null` silently applies `collectionLimit` to both lists and never reports `nextOffset` (Java `:368`, Go `:375`, JS `:1171`, PHP `:1250`, Python `:315`); unreachable from the tools today, but a `require` would document it. The `HierarchyPageRequest` KDoc's "preserve that query's encounter order" is not achievable for callers, because `ReferencesSearch` order is nondeterministic, and JS's barrel-first promotion (`JavaScriptHandlers.kt:1821`) reorders across limits. The tools sidestep this with offset 0 plus pointer identity, so the doc should say so.
- **Nested searches now run inside the outer query's consumer** (Java `:809`, Go `:813`, Rust `:993`). `AbstractQuery.forEach` serializes the consumer, so this is not a data race, but the nested `ReferencesSearch` runs while holding that monitor and stalls the parallel file-scanning workers on the legacy path. The old collect-then-recurse shape avoided that.
- **New runtime dependency.** `kotlinx-collections-immutable` is used only to share `visited` / `frontier` storage between cursor snapshots; with ≤ 8,192 keys per continuation the saving is marginal. If it stays, please confirm it does not collide with a platform-bundled copy (`excludePlatformProvided()` does not cover it).

### Nits

- `HierarchyContinuationRegistry.evictOverflow` (line 327): `val reason = if (…) LRU else LRU` has identical branches.
- Redundant explicit `catch (PCE | CancellationException | IndexNotReadyException) { throw e }` next to `rethrowIfControlFlow()` in Go `:851`, Rust `:1036`, Python `:810`, JS `:1808`.
- `ToolsTest.testCallHierarchyToolJavaScriptFixturePrioritizesRealisticIndexBarrelImportsBeforeVisibleLimit` now passes `maxNodes: 20`; in paged mode the page is sorted by `callDataComparator`, so the barrel-prioritization assertion passes by lexical order and the legacy behavior the test is named for lost its coverage.
- `HierarchyFileIdentityBehaviorTest` accepts any failure text; `HierarchyPaginationBehaviorTest` checks for "deleted"/"invalid" and this should too.
- The three new `src/kotlinPluginTest` tests never run in CI (`-PkotlinPluginTests=true` only), so "real Kotlin tests were run" is unverifiable here.
- `HierarchyContinuationRegistryBehaviorTest.testDefaultsUseIndependentTenMinuteTtlAndBoundedLru` only pins constants.

The new tests are otherwise strong: deterministic, no violations of the repo's assertion rules, and tight on BFS order, look-ahead budget, LRU/TTL/epoch handling and replay. PHP/Rust/Go coverage is limited to cancellation propagation.

---

hechtcarmel:
Re-checked head a126b39 (rebased onto `main` at dcc4869). Thanks for the rebase and the fixes; the update resolves part of the review, but the PR is still not mergeable. Status against the earlier review:

### Fixed ✅

- `allowSymbolId = true` in both tools — the five symbolId tests pass now.
- Both golden files regenerated against `main`.
- Java `kotlinCallableSource` rethrows control-flow exceptions wrapped in `InvocationTargetException`.
- Java callee dedup uses a set instead of the O(n²) scan.
- Rust type hierarchy skips the direction it was not asked for.
- Legacy handle budget tied to the registry capacity — the two handle tests pass. (Side effect to be aware of: one legacy response can now bind up to 4,095 handles and evict every other tool's handles from the shared registry.)

### Still blocking ❌

1. **CI Test is still red: 2 of 1516.** `McpToolDispatcherTest.testOneServerEpochIsCapturedBeforePreExecutionChecksForEveryRegistry` and `testPaginationCursorIgnoresTargetSelectorsForRouting` fail for the same reason as before: they assert dispatcher behavior (epoch capture before pre-execution checks; stripping `symbolId`/`target` when `cursor` is present) that `main`'s `McpToolDispatcher` does not have and this PR does not add. Either bring the dispatcher change into this PR or drop the two tests. The cursor test is also misnamed, since the dispatcher has no cursor logic.
2. **`CHANGELOG.md` still deletes released history.** The `## [5.11.0]`, `## [5.10.0]` and `## [5.9.6]` sections are still removed, and after the rebase the new `## [5.15.1] - 2026-09-16` header is removed too, so 5.15.1's fixes now sit under `[Unreleased]`. Please restore the released sections verbatim and keep only this PR's own entry under `[Unreleased]`.
3. **The five `McpToolDispatcherTest` tests are still deleted** (`testUndeclaredSymbolIdDoesNotRouteIndexStatus`, `testUndeclaredNestedTargetDoesNotRouteIndexStatus`, `testStringValuedTargetSchemaDoesNotEnableNestedSymbolRouting`, `testNullSymbolIdIsTreatedAsAbsent`, `testSymbolIdRoutesToItsOwningProjectWhenMultipleProjectsAreOpen`), with no dispatcher change to justify it. Please restore them.
4. **Docs are still appended, not updated.** `CLAUDE.md` still has four duplicated headings (Structured lookup targets, Rename preview, Safe-delete preview, Change-signature preview); `CLAUDE.md`, `README.md` and `SKILL.md` still carry the appended blocks re-describing #375–#398 and the "…have not been manually verified" sentence. Keep only the hierarchy-pagination section, and update the two tool entries in the CLAUDE.md tool list inline.

### Sidestepped ⚠️

- The schema budget test was raised from 110,000 to 115,000 bytes instead of trimming the new descriptions. That budget was set deliberately in #395; if a larger tools/list is the intended trade-off, please say so in the PR description so it is an explicit decision rather than a bump in a `test:` commit.

### Not yet addressed (from "should be fixed in the same round")

- Legacy call-hierarchy cap/ordering changes (Java/Go/JS/PHP reference caps removed; PHP and Rust callees 40 → 20; PHP order; Python dedup key) are neither restored nor listed in the CHANGELOG.
- Tool-level dedup in `pendingCallNode` / `pendingTypeNode` is still a linear scan over all visited smart pointers per candidate, so late pages of wide hierarchies stay quadratic.
- Python `getSubtypes` with `directOnly` still runs the deep inheritor search and walks every transitive descendant per expansion.
- JS strategy 1 / PHP paged subtypes still rely on native queries being deep, with only injected executors in tests.
- Handler contract: `page != null && direction == null` still silently truncates; the `HierarchyPageRequest` KDoc still promises encounter order that callers cannot provide.
- Nested `ReferencesSearch` still runs inside the outer query's consumer in Java/Go/Rust callers.
- `kotlinx-collections-immutable` dependency still unjustified / unchecked against the platform's bundled libraries.
- Nits from the review (identical `if/else` branches in `evictOverflow`, redundant explicit catch clauses, the `ToolsTest` barrel-prioritization coverage loss, `HierarchyFileIdentityBehaviorTest` accepting any failure text) are unchanged.

I'll re-review once the draft comes out of draft with the blocking items addressed.

---

hechtcarmel:
Re-checked head d8fe563. CI is fully green now (Build, Test, Inspect code, Verify plugin), and this push resolves every item I had marked blocking. Thanks. Unfortunately the same commit introduces three new problems, two of which block on their own, so the changes-requested review still stands.

### Fixed ✅

- `CHANGELOG.md` matches `main` again, plus one line under `[Unreleased]`.
- The five deleted `McpToolDispatcherTest` tests are restored, and the dispatcher now implements what the two new tests assert (epoch re-check after pre-execution checks; `symbolId`/`target` stripped when `cursor` is present).
- `hierarchyDeclarationKey` replaces the linear scan over visited pointers.
- Java/Go/Rust callers collect inside the search consumer and recurse after it returns.
- Python tries the shallow inheritor query first; JS merges both inheritor strategies.
- `require(page == null || direction != null)` in every type handler; `HierarchyPageRequest` KDoc reworded.
- `ToolsTest` barrel test back in legacy mode with no-cursor assertions; `HierarchyFileIdentityBehaviorTest` checks the error text; registry defaults test exercises LRU/TTL; `evictOverflow` branches collapsed; PR description explains the 115,000-byte budget.

### New regressions ❌

**1. The docs cleanup truncated `README.md` and `CLAUDE.md`.** Relative to `main`, `README.md` drops from 603 to 339 lines and `CLAUDE.md` from 778 to 656. Everything after the first duplicated heading is gone, not just the duplicated blocks:

- `README.md` lost: Project Lifecycle Management Tools, Tool Availability by IDE, Multi-Project Support, Lifecycle Management, Tool Window, Error Codes, Settings, Requirements, Architecture, Contributing, License. The file now ends at the new "Bounded hierarchy pages" section.
- `CLAUDE.md` lost: Symbol handles across navigation and member editing, Structured Lookup Targets, Rename/Safe-delete/Change-signature Preview, Search Collection Pattern (Processor), Useful IntelliJ Platform Classes, Troubleshooting, Contributing / PR Checklist.

Please restore both files from `main` and re-apply only the additions (the updated `ide_type_hierarchy` / `ide_call_hierarchy` list entries and the one new section).

**2. A second Kotlin stdlib is now on the plugin runtime classpath.** `build.gradle.kts:77` dropped `{ excludePlatformProvided() }` from `kotlinx-collections-immutable`. That helper does not exclude the library itself; it excludes the library's *transitive* platform-provided dependencies. Resolving `runtimeClasspath` on this head shows:

```
org.jetbrains.kotlin:kotlin-stdlib:2.1.20
\--- org.jetbrains.kotlinx:kotlinx-collections-immutable-jvm:0.4.0
     +--- runtimeClasspath
```

so the plugin will bundle stdlib 2.1.20 next to the IDE's 2.2.x, which the comment block above `excludePlatformProvided()` explicitly forbids. The same tree also shows `kotlinx-collections-immutable:0.4.0` already arrives transitively via `kotlin-sdk-core`, so the explicit declaration was never needed. Either restore the exclusion or drop the declaration.

**3. The restored raw-reference caps also apply in paged mode.** `GoHandlers.kt:817`, `PhpHandlers.kt:1724`, `RustHandlers.kt:1012` (callers) and `RustHandlers.kt:636` (impl search) stop after `maxResults * 2` inspected references unconditionally. With `page != null`, `maxResults` is the `collectionLimit`, so an expansion can return fewer than `collectionLimit` results and `nextOffset = null` after inspecting only `2 × collectionLimit` references, which the tool then treats as a complete level. That silently truncates paged hierarchies, the one thing the paged contract promises not to do. Java (`JavaHandlers.kt:820`) and JS (`JavaScriptHandlers.kt:1752`) guard the cap with `maxResults == MAX_RESULTS_PER_LEVEL`, but a paged `collectionLimit` can equal 20 as well (e.g. `maxNodes = 18`, or any expansion where `maxNodes - discoveredThisPage + 1 == 19`), so that guard is fragile. Please gate the cap on `page == null` explicitly in all five places.

### Nit

- `McpToolDispatcher.kt:173`: the cursor stripping runs after `symbolId` routing, so a cursor request that still carries a stale `symbolId` and no `project_path` fails at routing with `SYMBOL_ID_EXPIRED` before the stripping helps. Strip (or skip routing) before `symbolIdForRouting` when the tool declares `cursor`.

I'll re-review once these are in.

---

hechtcarmel (approved):
Approving head c34721c (a pure squash of the reviewed commits; identical tree). All code findings from the review are resolved and CI is green on Build, Test, Inspect code and Verify plugin.

Two documentation-only issues remain and are being fixed in the follow-up version bump PR rather than here: `CHANGELOG.md` drops the 5.15.1 release section and its compare link, and `CLAUDE.md` reverts three #398 paragraphs (the dispatcher's 55-second deadline, the `LONG_POLL_TOOLS` registration step, and the Kotlin body-formatting test in the kotlinPluginTest list). Both come from restoring those files from a pre-#398 snapshot instead of `main`.

---

PR #403, opened 2026-09-17
Add structured file outlines with exact handles

RomanPozdeev:
## Summary

- add structured `ide_file_structure` nodes with nested declarations, source ranges, signatures, modifiers, and exact symbol handles
- preserve the legacy formatted structure string and Markdown heading handle resolution
- report handle budget exhaustion explicitly and cover Java, Kotlin, TypeScript, and Markdown declaration kinds
- restack onto upstream main after PR #400 and keep the branch to one commit

## Response and handle policy

- `structure` remains the default response. Structured `nodes` are opt-in with `includeNodes=true`.
- Exact declaration handles are opt-in with `includeSymbolIds=true`, which also enables `nodes`.
- Handle allocation is capped at 100 per response by default and can be lowered with `maxSymbolIds` (1–100), so one large outline cannot evict the registry's live handles from other tools.
- Empty or unparseable files keep the legacy message in `structure` and return an empty structured node list when structured output is requested.

## Kotlin coverage

Kotlin outline-kind coverage is opt-in: `gw ./gradlew test -PkotlinPluginTests=true --tests '*Kotlin*BehaviorTest' --no-configuration-cache --rerun-tasks`.


hechtcarmel (changes requested):
**Summary.** The feature is a good idea and the implementation is the right shape: handles are bound from the exact PSI element while the handler still holds it (so same-line overloads and nested declarations cannot be confused, which re-resolving by line could never guarantee), the preorder budget keeps one outline from evicting its own handles, the legacy `structure` string is preserved, and the transient `pointerTarget` follows the `ToolModels` pattern from #400. CI is fully green on head e6e657f (Build, Test, Inspect code, Verify plugin), the branch merges cleanly onto `main` (only #402 has landed since), and `scripts/check-pr.sh`'s static checks pass. One item blocks, and it is the same one as item 4 of the first #400 review: the docs are appended after the footers and mostly duplicate what is already there. Requesting changes; happy to re-review once that is fixed.

### Blocking

**1. Docs are appended after the footer again, and six of the seven sections are duplicates**

- `CLAUDE.md:780`–`840`: seven `###` sections sit after the `**Template Source**` footer (line 779). Six re-describe #375–#398 features the file already documents at lines 650–695 (Symbol handles across navigation and member editing, Structured Lookup Targets, Rename / Safe-delete / Change-signature Preview), and the `-PkotlinPluginTests=true` paragraph at line 806 repeats the Testing section (line 442). Only "Structured file outlines" (line 835) is new to this PR.
- `README.md:613`–`668`: the same seven sections after the "Plugin based on…" footer (line 612); the handle material is already at line 341.
- `SKILL.md:207`–`262`: the same seven sections, glued directly onto the hierarchy-pagination section with no blank line; guideline items 5 and 6 (lines 105–106) already cover handles, nested targets and previews.

Please keep only the file-outline paragraph, place it next to the existing handle docs rather than below the footer, and update the three existing tool entries inline: `CLAUDE.md:558`, `README.md:317` and `USAGE.md:57` still describe the tool as returning only the text tree with line numbers, while the sibling rows for other tools mention `symbolId`. (`USAGE.md`'s per-tool section and `tools-reference.md` are updated correctly.)

### Design questions (please decide in this round)

- **Every response now carries both representations.** For the two-node example in `USAGE.md`, the text tree is 78 bytes and the `nodes` JSON is 355 bytes, so a response is roughly 5–6× the legacy size, on a tool agents call for almost every file they inspect. Options: an `includeNodes` (or `includeSymbolIds`) parameter defaulting to `false`, or a `format` parameter that lets the caller drop `structure`. If the always-on dual payload is the intended trade-off, please say so in the PR description so it is an explicit decision rather than a side effect.
- **Handle allocation.** `FileStructureTool.kt:77` binds up to `responseHandleBudget` (500) handles per call with no opt-out, into the shared 4,096-entry `SymbolIdRegistry`. The paginated tools bind only per served page (default 25), so after this PR `ide_file_structure` is the registry's heaviest allocator: eight or nine outlines of large files evict every other live handle. The opt-in flag above would address this too; a smaller default budget is the other option.

### Should be fixed in the same round

- **Empty-file behavior change is not in the CHANGELOG.** `FileStructureTool.kt:70`–`80`: a file with no parseable structure used to return a plain-text success; it now returns the JSON envelope with the message in `structure` and `nodes: []`. That is the better shape, but it is a wire change for any client that special-cases that response, so list it under `[Unreleased]`.
- **The annotation-kind fix is not pinned.** `JavaHandlers.kt:1286` reorders the checks so `@interface` reports `ANNOTATION` instead of `INTERFACE`, and `OutlineHandleRoundTripBehaviorTest.kt:42` includes `@interface Marker`, but nothing asserts a `kind`. One `assertEquals` on the kinds of the four Java declarations in that fixture would guard both the annotation fix and the record/enum ordering.
- **`KotlinOutlineKindsBehaviorTest` never runs in CI** (`-PkotlinPluginTests=true` only), and the verification line in the description is the default `./gradlew test`. Please state whether it was run with the flag; I could not run it here.

### Nits

- `PythonHandlers.kt:1098`: stray blank line at the top of the catch block.
- `SKILL.md:206`→`207`: no blank line before the first appended heading (moot once the block goes).

The code itself is in good shape: every PSI touch stays inside the tool's read action, all helpers used (`classDisplayName`, `kotlinClassKind`, `resolveNavigationTarget`, `rethrowIfControlFlow`) already exist on `main`, `git diff --check` is clean, and the two new behavior tests are tight: same-line overloads get distinct handles that resolve to the `int` and `long` declarations, and a 4,096-field file returns every node with exactly 500 handles that all still resolve.

---

hechtcarmel:
Re-checked head 99e399f (`fix: make structured file outlines opt-in`, stacked on the reviewed e6e657f). CI is green on all four jobs (Build, Test, Inspect code, Verify plugin), the branch still merges cleanly onto `main`, `git diff --check` and the static `scripts/check-pr.sh` checks pass, and this push resolves every item from the review. Thanks. What is left is small enough for one more push; nothing blocks.

### Fixed ✅

- **Docs.** The appended blocks are gone from `CLAUDE.md`, `README.md` and `SKILL.md`; each file ends at its footer again and nothing else was lost (relative to the PR base they are +6, +6 and +7 lines). The new paragraph sits next to the existing handle docs, and the three tool entries (`CLAUDE.md:558`, `README.md:317`, `USAGE.md:57`) are updated inline.
- **Payload and handle churn.** `nodes` and handles are opt-in (`includeNodes`; `includeSymbolIds` implies nodes), the budget is 100 by default with `maxSymbolIds` 1–100 validated at `FileStructureTool.kt:92`–`96`, and the policy is stated in the PR description. Four new behavior tests pin it: a legacy call returns no nodes and leaves the registry empty, `includeNodes` alone binds nothing, `maxSymbolIds: 1` reports the omissions, and 101 is rejected with an actionable message.
- Empty-file response change listed in `CHANGELOG.md`.
- `OutlineHandleRoundTripBehaviorTest.kt:48` now asserts `ANNOTATION, ENUM, RECORD, CLASS` for the Java fixture, so the `@interface` fix is pinned.
- Python blank line removed; `tool-manifest.json` regenerated with only the three new `ide_file_structure` parameters and the description change.

### Remaining (small)

1. **Dead property.** `SymbolIdRegistry.responseHandleBudget` (`SymbolIdRegistry.kt:69`), added in the first commit, is no longer referenced by production or test code now that the tool uses its own `DEFAULT_HANDLE_BUDGET` / `MAX_HANDLE_BUDGET` (`FileStructureTool.kt:147`–`148`). Please remove it.
2. **CHANGELOG wording and section.** `CHANGELOG.md:11` reads "Document the empty-file response change: …", which describes the act of documenting rather than the change, and it sits under `### Added`. Something like "`ide_file_structure` now returns the JSON envelope for empty or unparseable files, with the legacy message in `structure`" under `### Changed`. Line 10 could also say that nodes and handles are opt-in, since that is now the headline of the feature.
3. **`tools-reference.md:340`** shows `nodes?, symbolIdsTruncated?, symbolIdsOmitted?`, but the tool serializes with `encodeDefaults = true`, so every response carries `nodes: []`, `false` and `0`; the keys are never absent. Drop the `?`s, or say "empty / false / 0 unless requested".
4. **`SKILL.md:106`**: a blank line was inserted between guideline items 5 and 6, which turns the tight list loose from that point on.
5. The description now gives the `-PkotlinPluginTests=true` command but not its result; a one-liner confirming `KotlinOutlineKindsBehaviorTest` passed with the flag would close that item (I still cannot run it here).

None of these block on their own; I'll approve on the next push that carries them.

---

hechtcarmel (approved):
Approving head 94a6ab3. CI is green on Build, Test, Inspect code and Verify plugin, the branch merges cleanly onto `main`, `git diff --check` and the static `scripts/check-pr.sh` checks pass, and the last two pushes resolve every item from the review and the re-check: `nodes` and handles are opt-in with a validated 1–100 budget and four tests pinning the legacy path; the appended doc blocks are gone and the file-outline paragraph and tool entries live inline; the empty-file change sits under `### Changed`; the Java fixture asserts `ANNOTATION, ENUM, RECORD, CLASS`; the unused `responseHandleBudget` is removed; and `tools-reference.md` no longer marks always-present keys as optional. Thanks for the quick turnarounds.

One non-blocking leftover: the description gives the `-PkotlinPluginTests=true` command but not its outcome, so `KotlinOutlineKindsBehaviorTest` remains unverified in CI and I could not run it here. A one-line confirmation, or a run before the next release, would close it.

---

RomanPozdeev:
I’m very grateful for the excellent product you’ve built.

For now, this will be my last PR. I don’t have any plans to add new features in the near future, but I may come up with some new ideas down the road.

PR #407, opened 2026-09-17
docs: fix ide_restart description that causes agents to give up after restart

mdproctor:
## Summary

The `ide_restart` tool description says:

> "Call this as a final step — do not expect to execute further tool calls in the same session."

This is factually wrong for Streamable HTTP, where each tool call is an independent HTTP POST with no persistent session. Agents read this instruction and treat `ide_restart` as terminal — giving up instead of polling for reconnection. This breaks the common `ide_install_plugin` → `ide_restart` → verify flow.

Updated to explain that the server shuts down temporarily during restart, and agents should poll `ide_index_status` after ~30 seconds to confirm the IDE is back, then continue with follow-up commands.

No code changes — just the tool description text and CLAUDE.md inventory line.

## Test plan

- [x] Golden manifest regenerated — diff shows only the description text change


hechtcarmel:
Thanks, this is correct and it is going in as is.

Verified against the code: the Streamable HTTP endpoint is stateless (a fresh SDK server per POST, no session id), and `StatelessServerLifecycleTest` pins that a bare `tools/call` succeeds without a prior `initialize`, so nothing needs re-establishing once the IDE is listening again. The server restarts on its own with the reopened projects (`McpServerStartupActivity` is a `ProjectActivity`), and `ide_index_status` is enabled by default, so "poll and continue" is the right guidance. The smoke-test protocol in `smoke-tests/mcp-protocol.md` already works exactly that way.

What we are going to do:

1. Merge this PR now.
2. Follow up with a docs-only PR that brings the remaining places in line, since they still say "terminates the MCP connection, reconnect your client": the bundled companion skill (`SKILL.md` and `references/tools-reference.md`, which is what agents actually read), `USAGE.md`, and `README.md`. That PR also adds the `[Unreleased]` changelog entry CONTRIBUTING asks for, and spells out two details the new description leaves out: legacy SSE clients do have to reopen their stream, and symbol handles / search cursors issued before the restart are invalid afterwards.

---

